PR #5, opened 2016-05-22
fix assertions taking std::ostream

lichray:
Without this `const`, `REQUIRE(a_function_returns_ostream_ref)` won't compile.


onqtam:
Hi! I merged it in the dev branch - from the command line - master branch is only for releases.
So if you need this change - the dev branch has it.

1.1 or 1.0.1 is coming in a few days - then this will get into master.

Thanks! I will update the Contributing section on the main readme soon.

PR #15, opened 2016-05-30
Include "program code" in example

martinmoene:
In the example, shows both test and "program" code as well as generates output for both.

Further clarifies what `context.run()` does in the comment (as it puzzled me a bit at first).
I think it's better to remove `(and --no-run)` from the next comment as `shouldExit()` is not true for `--no-run` [1] .

Perhaps also update the (same) example [The main() entry point](https://github.com/onqtam/doctest/blob/dev/doc/markdown/main.md).

[1] rename `shouldExit()` to `mustExit()` ?

### DIFF
--- a/examples/user_supplied_main/main.cpp
+++ b/examples/user_supplied_main/main.cpp
@@ -1,6 +1,8 @@
 #define DOCTEST_CONFIG_IMPLEMENT
 #include "doctest.h"
 
+int program();
+
 int main(int argc, char** argv) {
     doctest::Context context(argc, argv); // initialize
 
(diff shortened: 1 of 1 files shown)


onqtam:
oh! the ```-no-run``` thing in the comment should have been ```-exit``` - thanks for the catch!
the CI builds fail though... because the output doesn't match the reference output (because of the printf) - I will merge it and update the reference output.

I feel it's a bit too late to change the name of ```shouldExit()```.

martinmoene:
First the build should fail, shouldn't it (in truth I forgot about the ref. output ;)

`shouldExit()`, or expectsExit() is perhaps just right; programmers may have their reasons to not follow the advice to exit there. If you've watched one of [Kevlin Henney's talks about testing](https://www.infoq.com/presentations/unit-testing-tips-tricks), you may recognise my hesitation to use the word _should_  when used in relation to testing (ca. 56 min in ;).

PR #26, opened 2016-06-20
dont show green when tests fail

ferkulat:
this is a quick fix.
I would even go further to make the first yellow line red/green.
But this would need more changes to make "numFailed" available there


onqtam:
I might change colors a bit later on - it's a very personal/opinionated topic - I should add color scheme support to make everyone happy

PR #46, opened 2016-12-20
Add exception handling to the framework

eliaskosunen:
This PR the functionality of handling exceptions occurring in tests better. I think this is much needed.

In case of an `std::exception` or alike, it prints out the exception message using `.what()`.


onqtam:
perfect pr - thanks - merged!

Note that I am planning for the assertion macros to capture ```std::exception``` as well - and will add it at some point in the future.

PR #50, opened 2017-02-18
Add option --force-colors/-fc

ligfx:
Thanks for all of the work put into doctest, it's definitely the fastest-compiling test library I've seen.

In this patch I added a feature I've been missing: forcing colored output. The build system Ninja doesn't give subprocesses a pseudotty, but will pass escape codes through. Being able to tell doctest to output in color when using Ninja would be great.


onqtam:
Thanks for the PR! Unfortunately I won't be able to merge it as it is because it's against master and the CI builds don't pass.

I will implement it quickly on my own - with the same flag. It will be in master in version 1.1.4 - which should be released today...

onqtam:
1.1.4 is out - let me know if everything is ok

ligfx:
Works like a charm, thanks!

PR #60, opened 2017-03-29
Add CMake package support + options to skip examples or coverage

piribes:
Add support to install doctest using CMake, as well as package config files (version, targets) useful to find_package [1] in some systems, like [hunter](https://github.com/ruslo/hunter/pull/727).

Add CMake options:
- DOCTEST_SKIP_COVERAGE: When ON, skip coverage build (default OFF)
- DOCTEST_SKIP_EXAMPLES: When ON, skip examples build (default OFF)

[1] https://cmake.org/cmake/help/git-master/manual/cmake-packages.7.html#creating-packages

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.0)
 
 set(PROJECT_NAME doctest)
-project(${PROJECT_NAME} VERSION 1.1.4)
+project(${PROJECT_NAME})
 
 option(DOCTEST_SKIP_COVERAGE "Skip coverage" OFF)
 option(DOCTEST_SKIP_EXAMPLES "Skip examples" OFF)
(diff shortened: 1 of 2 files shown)


onqtam:
Thanks for the contribution!
I will probably try to merge it in the dev branch right before I release version 1.2 which should be in the near month.
I might also tweak it a bit after that - I'm planning an overhaul of the examples folder and the test coverage...

piribes:
Good news :) It is a pleasure.

piribes:
Just realized that hardcoding the version into package() is not so good.
I'll provide an update that uses git to read the version from tags if you agree.

onqtam:
@piribes sure! :)

onqtam:
Actually the version can be read from ```https://github.com/onqtam/doctest/blob/master/scripts/version.txt```

piribes:
Now it uses that file to read version into PROJECT_VERSION.

onqtam:
a few notes:
- manually merged in dev and pushed - see it [here](https://github.com/onqtam/doctest/tree/3c7110d2688ac3086740168401d9483a45528183)
- moved the cmake folder into scripts
- now there is only one option - ```DOCTEST_WITH_TESTS``` instead of the 2 SKIP ones
- properly setting the VERSION (including PROJECT_VERSION_MAJOR, etc.) of the project doctest in cmake
- reworked the cmake code to be suitable for direct usage
    - no longer overriding the add_executable/add_library cmake functions
    - the doctest interface library can now be linked to (with ```target_link_libraries```) with the include directories properly set up (using ```target_include_directories(doctest INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/doctest/>)```)
and a question:
- do you think I should move the header from "<repo>/doctest/doctst.h" into "<repo>/include/doctest/doctest.h" ???

piribes:
> All ok?

Seems ok to me.

> do you think I should move the header from gitrepo/doctest/doctst.h into gitrepo/include/doctest/doctest.h ???

Other projects use include/ for the interface and src/ for the implementation files, but it is a convention; here, I think there's no problem in leaving everything as it is.

It would be different if we were installing doctest.h in a different directory than include/doctest/; in this case, you have to remember that you'd be sharing that folder with other projects, so it would be a better idea to respect the convention.

onqtam:
version 1.2 is out!

PR #64, opened 2017-04-20
Update doctest to work with ARM DS5-compiler

tomasnilefrost:
While trying to use doctest with the ARM DS5-compiler I ran into some issues with the standard library methods (free, malloc, sprintf, printf etc) not being present.

ARM DS5 puts all of the standard library functions in the std-namespace and do not put anything in the global namespace.

See for instance the blog-post: https://developers.redhat.com/blog/2016/02/29/why-cstdlib-is-more-complicated-than-you-might-think/
"You must not assume that <cxxx> adds any names to the global namespace, and you must not assume that <xxx.h> adds any names to namespace std."

The solution is to explicitly point out for instance std::printf and not rely on the standard library for the specific compiler to export these symbols in the global namespace.

This might actually also be related to a previous pull-request ( https://github.com/onqtam/doctest/pull/37 ) but I haven't investigated it further.

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -2955,7 +2955,7 @@ namespace detail
 #ifdef _MSC_VER
 #define DOCTEST_SNPRINTF _snprintf
 #else
-#define DOCTEST_SNPRINTF snprintf
+#define DOCTEST_SNPRINTF std::snprintf
 #endif
 
 #undef DOCTEST_GCS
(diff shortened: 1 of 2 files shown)


onqtam:
Thanks for the PR (made against the dev branch!!!)

The previous issue really seems related to this. I finally understand the difference between the headers - thanks for the redhat article.

I'm merging this as it is because it's the best option - even though a solution could be to just do ```using std::strcpy;``` and etc - or just include the headers in a C style - ```#include <xxx.h>```. I'll also move away from most of these C functions as I move to streams and etc.

I will give a hard deadline for version 1.2 of 1 month from now (I keep postponing it...) - so only then will these changes appear in the master branch.

tomasnilefrost:
Thanks for accepting the pull request. Great job with doctest!

Regarding solution of `using std::strcpy`, it is probably best avoided since that would mean that all code including doctest.h would also get a polluted global namespace of these using-declarations.

PR #87, opened 2017-08-23
Add support for templated scenarios

Lectem:
## Description
There is no macro for templated scenarios, this commit adds `SCENARIO_TEMPLATE`  which in turn calls `TEST_CASE_TEMPLATE` as done with `SCENARIO`.

As I created this PR, I was also wondering if the BDD macros actually worked with `DOCTEST_CONFIG_NO_SHORT_MACRO_NAMES` since it uses the short macro names and not the ones with the DOCTEST_ prefix.
If needed I will update and fix the BDD macros to use the DOCTEST_* variants.


onqtam:
Thanks - I'll review this in a couple of days.

Lectem:
No problem ! Just added a commit so that it is shown in the documentation.

PR #94, opened 2017-10-17
Fix nullptr access

Lectem:
## Description

If no test suite has been set and we fail in a test, then `tc.m_test_suite` is `nullptr` and will crash the program.

### DIFF
--- a/doctest/parts/doctest_impl.h
+++ b/doctest/parts/doctest_impl.h
@@ -1442,7 +1442,7 @@ namespace detail
             forDebugConsole += d1;
             forDebugConsole += d2;
         }
-        if(tc.m_test_suite[0] != '\0') {
+        if(tc.m_test_suite && tc.m_test_suite[0] != '\0') {
             DOCTEST_PRINTF_COLORED(ts1, Color::Yellow);
             DOCTEST_PRINTF_COLORED(ts2, Color::None);
             forDebugConsole += ts1;


onqtam:
Thats strange - can you provide more information on how to reproduce the problem?

There are many test cases that fail in the examples of the project (like [here](https://github.com/onqtam/doctest/blob/master/examples/all_features/assertion_macros.cpp)) and many of them are in no test suite. Also all examples are ran through the sanitizers, and the doctest header itself resets the test suite for each translation unit it is included with [this line](https://github.com/onqtam/doctest/blob/1.2.5/doctest/doctest.h#L3085).

merging this check is not a problem but I'm curious how it happened

Lectem:
Sorry I forgot to check how it happened, but my suspect might be precompiled headers or the fact no parameters at all were given to the context.
I'll check this at the beginning of the week.

Lectem:
OK so I found the issue, sorry for the delay.
It actually happens because doctest was included in the precompiled headers.

onqtam:
strange... I include doctest in my precompiled headers as well (clang/gcc/msvc latest versions)... Anyway - I'll merge this commit soon (not right now - started a contracting work a few days ago and theres a lot of work).

thanks for the PR!

Lectem:
No problem, I also find this strange. If I find more information about it I'll comment here

onqtam:
decided to just merge this :)

PR #96, opened 2017-10-23
Added an option to not install Doctest

NeverMine17:
## Description
I was having a problem with installing so I made an option to skip the installation part.


onqtam:
thanks!

PR #99, opened 2017-11-04
Correct DOCTEST_NO_INSTALL logic; do install unless it is set

OdyX:
## Description

Commit b63bf85ca872c5fd7f53c2e6d97f61813298936d (fixing #96) introduces the `DOCTEST_NO_INSTALL` variable, but it's use is reversed: by default it is set to `OFF` (meaning you _do_ want to install), but the test says "If set to ON, do install".

This made me upload a `doctest.h`-less `doctest-dev` package to Debian.

## Debian bug
This was detected through a failed build of `argagg` and reported as https://bugs.debian.org/880721 .

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,7 +57,7 @@ write_basic_package_version_file(
 
 configure_file("scripts/cmake/Config.cmake.in" "${project_config}" @ONLY)
 
-if (${DOCTEST_NO_INSTALL})
+if (NOT ${DOCTEST_NO_INSTALL})
     install(
         TARGETS ${PROJECT_NAME}
         EXPORT "${targets_export_name}"


NeverMine17:
Such a stupid bug I made >.<

onqtam:
Also my bad for insta-merge of that :)

this will get merged now in master as well without releasing version 1.2.7 - is that ok?

Also many thanks for maintaining a debian package!

OdyX:
No problem for not releasing 1.2.7; it's patched in Debian already. Thanks for the quick merge!

a4z:
just also got hit with this, when will 1.2.7 be available? 

btw, I do not understand the original issue, adding a no install option. Just do not call `make install`
I think the whole thing should be removed, useless, just makes the cmake file more complex, and error prone, as proven with this issue

onqtam:
The fix is currently in master - 1 commit ahead of the ```1.2.6``` tag

a4z:
thanks for your explanation @onqtam 
for this case I would suggest thinking about something like
```
if( ${CMAKE_PROJECT_NAME} STREQUAL ${PROJECT_NAME})
   # have the default install to true, other wise false
endif( ${CMAKE_PROJECT_NAME} STREQUAL ${PROJECT_NAME})
```
so you can detect if  you are included or not and set the proper default per default

PR #112, opened 2017-12-22
Update doctest_impl.h

vladimirgamalyan:
Fix macros `WIN32_LEAN_AND_MEAN` typo

## Description

This pull request changes a define of macros WIN32_**M**EAN_AND_**L**EAN to WIN32_**L**EAN_AND_**M**EAN. This should increase compilation speed.
Also, it removes warnings when a header 'Winsock2.h' be included.

### DIFF
--- a/doctest/parts/doctest_impl.h
+++ b/doctest/parts/doctest_impl.h
@@ -554,9 +554,9 @@ extern "C" __declspec(dllimport) int __stdcall IsDebuggerPresent();
 #ifdef DOCTEST_PLATFORM_WINDOWS
 
 // defines for a leaner windows.h
-#ifndef WIN32_MEAN_AND_LEAN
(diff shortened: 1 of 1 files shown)


onqtam:
look at what [I found](https://github.com/nothings/stb/blob/master/tests/stb_cpp.cpp#L1) :D

There are a few other mentions of ```WIN32_MEAN_AND_LEAN``` across the web but I guess they are the wrong thing.

Thanks for this - merging!

vladimirgamalyan:
Yeah, usually it's not clear to see when the macro works properly. Actually, I found it when I added a header "Winsock2.h", which conflicts with "Winsock.h". Last one should not be included in "Windows.h" by WIN32_LEAN_AND_MEAN macro.

PR #131, opened 2018-04-30
Fix deprecated-declarations warning with GCC-8.0

### LINKED ISSUE #130: GCC 8.0 std::uncaught_exception() is deprecated
When compiling with GCC-8.0 the deprecation warning for std::uncaught_exception is not silenced and will cause problems when compiling with '-Werror'

[Example travis output](https://travis-ci.org/AMS21/Andres-Standard-Library/jobs/373145144#L1230)

* doctest version: **v1.2.8**
* Operating System: **Trusty Ubuntu (Travis CI)**
* Compiler+version: **GCC/G++-8 (Ubuntu 8-20180424-0ubuntu1~14.04) 8.0.1 20180424 (experimental) [trunk revision 259590]**

AMS21:
This should fix #130


onqtam:
Weird that old compilers don't complain for the unknown pragma (i have ignored the unknown pragmas warnings but I also remember having to explicitly check if warnings are supported by version X or above...)

thanks - merging!

PR #158, opened 2018-09-27
Make it possible to change the command line options prefix

tbleher:
## Description

When including doctest in a production executable, the "dt-" prefix is too
unclear for most users. Add a DOCTEST_CONFIG_OPTIONS_PREFIX define so
the prefix can be changed.

Also make sure that when DOCTEST_CONFIG_NO_UNPREFIXED_OPTIONS is set,
the help text displays the correct help options.

Note: this patch arose out of two different needs:
* The first was that, when DOCTEST_CONFIG_NO_UNPREFIXED_OPTIONS is set, the help text was quite unclear (since all options were shown in their unprefixed form, but that form couldn't be used). So the help text had to be changed to show the prefixed form at least when that define was set.
* The second was, that ```dt-``` is clear as a prefix for users who know about doctest, but is not so clear when integrating the code into a production application, which already has a number of options. So a way was needed to choose a different prefix (we chose ```selftest-```).

I know that this patch makes the printing a bit harder to read - I'm open to suggestions on how to do that more cleanly.
Also, this patch solves two issues at once, but since they touch the same code area and I developed them at the same time, this is just one patch for now. If need be, I can split them into two patches.


onqtam:
Thanks for the big patch - I will review it soon!

Out of curiosity - is doctest being used at BMW? :)

onqtam:
Thanks again and sorry for the huge delay!

merged into dev - will probably release version 2.1 tonight

tbleher:
Thank you for merging, and no problem about the delay! I really appreciate your work and the improvements you did on my changes.
To answer a previous question I forgot to answer: yes, doctest is used inside BMW - I found doctest very useful and recently introduced it into a simulation software I maintain.

Issue #115, opened 2018-01-16
clang-tidy warning about cert-err58-cpp


alpha-X-omega:
## Description
run-clang-tidy-5.0.py is detecting

   warning: initialization of '_DOCTEST_ANON_VAR_3' with static storage duration may throw an 
   exception that cannot be caught [cert-err58-cpp] TEST_CASE("example") {

The macro expansion ended at

   note: possibly throwing constructor declared here
        TestCase(funcType test, const char* file, unsigned line, const TestSuite& test_suite,

### Extra information
* doctest version: **v42.42.42**
* Operating System: **Ubuntu 16.04.3**
* Compiler+version: **Clang-5.0**

The good news is that doctest satisfies all the other checks :)

onqtam:
Any idea how to suppress this? I cannot use ```// NOLINT``` where the macros are defined because that has to be used where they are used. I also don't want to add a ```try {} catch(...) {}``` block in the test registration function just because of this...

alpha-X-omega:
I passed `--line-filter=[{"name":` "<absolute_path>/doctest/doctest.h", "lines": [[1, 3048], [3050, 5701]]}]` to clang-tidy :-)

onqtam:
In that case I'll be happy to close this issue :D

BrannonKing:
This is clearly not fixed. CLion complains about it on every test.

onqtam:
@BrannonKing Can you use the workaround that is provided in the issue (with using ```--line-filter=...```)?

I might think about adding ```noexcept``` but this is not a priority for now - let me know if you want this issue reopened and something done about it..........

onqtam:
It seems that this is actually solvable precisely by using ```// NOLINT``` - as can be seen in this PR:

https://github.com/onqtam/doctest/pull/159

Trass3r:
> I might think about adding `noexcept` but this is not a priority for now - let me know if you want this issue reopened and something done about it.

I'd also expect these kind of diagnostics to be resolvable by `noexcept`/`constexpr`.

onqtam:
Back in 2018 `noexcept` wasn't being used in doctest (I think). Now it is.

So yes, we should prefer using `DOCTEST_NOEXCEPT` to suppressing linter warnings. PRs are welcome!

serguei-k:
Hi Viktor,
making the following change suppresses this warning:
[...]

PR #159, opened 2018-11-03
Remove clang-tidy warnings for static fields created by doctest

rantasub:
## Description
When writing unit tests with doctest, clang-tidy raises a warning at each invocation of the `TEST_CASE()` or `SCENARIO()` macro, explicitly `fuchsia-statically-constructed-objects` and `cert-err58-cpp` for my current version of clang-tidy. The root cause is the intended creation of static variables by the framework. As these warnings are not a consequence of the user written code and there are plenty of them for a big test suite, they should be disabled.

This pull request adds an explicit suppression for these two tests at the responsible line in the implementation. 

## GitHub Issues
The warning `cert-err58-cpp` was already discussed in issue #115. The solution provided there however requires some action by the user.


onqtam:
A comment for suppression at the end of a macro works when the macro is expanded elsewhere through a few macro indirections? Nice!

Thanks for the PR!

PR #162, opened 2018-11-16
Fixed build error under Android NDK

tals:
## Description
This allow doctest to build cleanly with the Android NDK, using the clang compiler targeting arm64

Without it, clang will complain about
```
cannot initialize a member subobject of type 'int' with an rvalue of type 'nullptr_t'
            struct sigaction sa = {nullptr};
                                   ^~~~~~~
```

### DIFF
--- a/doctest/parts/doctest_impl.h
+++ b/doctest/parts/doctest_impl.h
@@ -1430,7 +1430,7 @@ namespace {
             sigStack.ss_size  = sizeof(altStackMem);
             sigStack.ss_flags = 0;
             sigaltstack(&sigStack, &oldSigStack);
-            struct sigaction sa = {nullptr};
+            struct sigaction sa = {0};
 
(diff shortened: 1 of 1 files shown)


onqtam:
Thanks!

I'll also commit warning suppressions around that piece of code because of not using nullptr with other compilers

doctest/parts/doctest_impl.h
```diff
@@ -1430,7 +1430,7 @@ namespace {
             sigStack.ss_size  = sizeof(altStackMem);
             sigStack.ss_flags = 0;
             sigaltstack(&sigStack, &oldSigStack);
-            struct sigaction sa = {nullptr};
+            struct sigaction sa = {0};
```
theodelrieu: Shouldn't it be `struct sigaction sa{}`?

I'm not sure if this initializes all the struct members, whereas value-initialization will.

PR #179, opened 2019-01-27
Fix CMake include path #175

### LINKED ISSUE #175: Change the include path of the `doctest` CMake interface target so users need to specify the folder as well
I am just curious about the rationale behind using "doctest.h" instead of "doctest/doctest.h".

ncihnegn:
Close #175


onqtam:
thanks!

PR #185, opened 2019-02-18
Rename doctest_impl.h to doctest.cpp for less confusion.

dimztimz:
## Description

doctest_impl.h was not really a header file, it contains ODR definitions.
It was meant to be included by one and exactly one implementation file.
It was confusing for contributors. Now it is more clear.

Catch2 names its implementation files .cpp.

As a consequence, we simplified the creation of doctest with main.
Extra file doctest_main.cpp is not needed.


onqtam:
Good idea - thanks!

PR #193, opened 2019-03-13
Fix propagating include directories of target doctest_with_main

dimztimz:
I broke this cmake target in my last patch
0dbbd4f0ff6cfc69517e64ffd2fa0ca1e6610b5b

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,16 +46,12 @@ add_custom_target(assemble_single_header ALL DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}
 ## TESTS/EXAMPLES/HELPERS
 ################################################################################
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -46,16 +46,12 @@ add_custom_target(assemble_single_header ALL DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}
 ## TESTS/EXAMPLES/HELPERS
 ################################################################################
 
-if(${DOCTEST_WITH_MAIN_IN_STATIC_LIB} AND NOT CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+if(${DOCTEST_WITH_MAIN_IN_STATIC_LIB})
```
dimztimz: concerning this change, please leave it as, it makes development inside IDE easier because they detect the target and thus they detect the .cpp file too.

onqtam: fair enough :)

PR #197, opened 2019-03-15
Fix UB of defining stuff in namespace std. Just include the headers.

### LINKED ISSUE #194: Forward-declaring identifiers in std:: is UB - consider including some of the cheaper C/C++ stdlib headers
I noticed that you redefine some stuff that is already part of C and C++ standard library.

For example:

1. You wrap `memcpy` inside `my_memcpy`
2. You redefine `nullptr_t` (which is UB)
3. You forward declare some iostream stuff (again UB).

`cstddef` and `cstring` are two headers that are EXTREMELY common. I can't think of a single non-trivial program that does not already includes these two headers. Just including C++ string or vector brings in these two. So, if some client code has already included them, and Doctest includes them, the second inclusion is practically a no-op because of header include guards.

Therefore, when redefining this stuff, you are optimizing the compile time for the best case, and that is using doctest alone, but you are worsening the common case, which is using doctest to test some other software.

dimztimz:
Having this UB may randomly break just by some compiler update.
Fixes #194

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -363,39 +363,9 @@ extern "C" __declspec(dllimport) void __stdcall DebugBreak();
 #define DOCTEST_BREAK_INTO_DEBUGGER() ((void)0)
 #endif // linux
 
-#if DOCTEST_CLANG
-// to detect if libc++ is being used with clang (the _LIBCPP_VERSION identifier)
-#include <ciso646>
-#endif // clang
-
-// Forward declaring 'X' in namespace std is not permitted by the C++ Standard.
-DOCTEST_MSVC_SUPPRESS_WARNING_WITH_PUSH(4643)
-
-#if defined(_LIBCPP_VERSION) || defined(DOCTEST_CONFIG_USE_IOSFWD)
-// not forward declaring ostream for libc++ because I had some problems (inline namespaces vs c++98)
-// so the <iosfwd> header is used - also it is very light and doesn't drag a ton of stuff
-#include <iosfwd>
-#else  // _LIBCPP_VERSION
-namespace std {
-template <class charT>
-struct char_traits;
-template <>
-struct char_traits<char>;
-template <class charT, class traits>
-class basic_ostream;
-typedef basic_ostream<char, char_traits<char>> ostream;
-} // namespace std
-#endif // _LIBCPP_VERSION || DOCTEST_CONFIG_USE_IOSFWD
-
-#ifdef _LIBCPP_VERSION
 #include <cstddef>
-#else  // _LIBCPP_VERSION
-namespace std {
-typedef decltype(nullptr) nullptr_t;
-}
-#endif // _LIBCPP_VERSION
-
-DOCTEST_MSVC_SUPPRESS_WARNING_POP
+#include <cstring>
+#include <iosfwd>
 
 #ifdef DOCTEST_CONFIG_INCLUDE_TYPE_TRAITS
 #include <type_traits>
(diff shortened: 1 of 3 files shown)


onqtam:
I can confirm that the code is correct, even though clang gives the ```Wdynamic-class-memaccess``` warning. I'll suppress it.

dimztimz:
I tried with clang and indeed got that warning. Fixing the warning requires header `<new>` which is another cheap header and using placement new.

Suppressing the warning is just hiding more bugs.

onqtam:
When we go down that line... why not get rid of the String class altogether and use std::string? I shouldn't have said that :| Now I'm thinking about it...

dimztimz:
Well, in C++ constructing object with memcpy is UB. The String class is not UB.

onqtam:
regarding the memcpy and <new> - I'm convinced, that's the right way to go indeed.

dimztimz:
No, to be completely correct, memcpy can be used only on trivially copyable types. Otherwise you have UB.

dimztimz:
I will close this as obviously more stuff needs fixing.

onqtam:
It's just that object construction - I'll fixup the code once I get to it.

Here's the deal: I'll work a few more days on the framework and before the release I'll experiment with stdlib headers. I'll most probably merge this.

dimztimz:
OMG this is just frustrating. MSVC 2015 fails for warnings in its own headers.

I can't be bothered to fix this anymore, please just reduce the warning level. MSVS level 4 is known to produce warnings in its own headers, nobody uses that.

onqtam:
@dimztimz :D It's actually not /W4 but /Wall that is used... and I have absolutely no clue as to why I decided to go down that road and suffer.

Anyway don't bother - I'll handle this PR going forward.

dimztimz:
I would suggest you to do the following

1. Reduce the reported warnings to something sane like Wall Wextra on gcc and clang
2. Delete almost all suppresions of warnings. I hace never seen code that wasted 100s of lines just to supress warnibgs.
3. Fix the actual warnings

onqtam:
Someone made a PR where even ```std::tuple``` is forward-declared... https://github.com/onqtam/doctest/pull/203

And in there he uses ```_LIBCPP_BEGIN_NAMESPACE_STD``` so apparently there is an easy way to deal with the inline namespaces of some stdlib implementations. I also found facebook/folly forward-declaring types in the std namespace:
https://github.com/facebook/folly/blob/master/folly/Traits.h#L628

So if Facebook are OK with this kind of UB - then I'm fine with it as well. I'm guessing that if other stdlib implementations (not only libc++) move to inline namespaces they will provide a similar macro.

I will make a compromise however: I'll change [```DOCTEST_CONFIG_USE_IOSFWD```](https://github.com/onqtam/doctest/blob/master/doc/markdown/configuration.md#doctest_config_use_iosfwd) to something like ```DOCTEST_CONFIG_USE_STD_HEADERS``` and when it is defined - the appropriate std headers will be included (not only for the streams but also for ```nullptr_t```) instead of the forward-declarations. So the only piece of UB left when using that identifier will be the construction of the object with memcpy instead of using placement new.

This has the following benefit for me: I don't have to think about this right now and I'm skipping having to re-do the benchmarks & update the graphs (because the compilation times will have significantly changed) which takes around 1 day...

So I'm closing this PR.

dimztimz:
I'd advise you not to make the code any messier with more config macros.

But I don't really care anymore, I decided not to use doctest at all. The forward declaring is not the only UB. The memcpy you use instead of placement new to construct object is another much dangerous UB. Abd probable there is more ub hidden somewhere.

horenmar:
@dimztimz I think you would be interested in Catch2

PR #204, opened 2019-03-20
Fix unmatched bracket in DOCTEST_TEST_CASE_CLASS

patstew:


onqtam:
Weird how it actually worked like it is currently - thanks!

PR #209, opened 2019-03-23
Add two very simple examples of using doctest with CMake

pr0g:
## Description
I was hoping it might be possible to add a couple more minimal examples of how to set up and use doctest.

I found the existing examples helpful but there's also quite a lot going on in each. They can be a little confusing/intimidating to a newcomer to the library. These two new examples provide the bare minimum for using doctest with a plain exe or a dll.

There may well be some stuff I've missed and I'm happy to provide more info in a README file if you think that'd help. If you can think of better names or locations for the examples please let me know too.

The library is brilliant by the way and I'm hoping to use it a lot more in the future! 😄 

Thanks!

Tom

### DIFF
--- a/examples/dll_installed_doctest/CMakeLists.txt
+++ b/examples/dll_installed_doctest/CMakeLists.txt
@@ -1,0 +1,14 @@
+cmake_minimum_required(VERSION 3.0)
+
+project(example_dll VERSION 0.0.1 LANGUAGES CXX)
+
+find_package(doctest REQUIRED)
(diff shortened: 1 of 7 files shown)


onqtam:
I moved them in a different directory:

https://github.com/onqtam/doctest/tree/master/examples/installed_doctest_cmake

pr0g:
Awesome! Thank you for the merge :)

Sounds perfect, much appreciated!

PR #213, opened 2019-03-26
[regression] Workaround MSVC preprocessor issue triggered by REQUIRE_THROWS

zhihaoy:
In 2.3.0 REQUIRE_THROWS goes through one extra level
when passing its expr argument, and this seems to
cause MSVC 19.16 preprocessor to refuse to do
stringification with C2026 (string too big, trailing
characters truncated).  The string is indeed big (6.8KB),
but not that big (16KB).

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -2107,8 +2107,6 @@ constexpr T to_lvalue = x;
 #define DOCTEST_REQUIRE_FALSE_MESSAGE(cond, msg) do { DOCTEST_INFO(msg); DOCTEST_ASSERT_IMPLEMENT_2(DT_REQUIRE_FALSE, cond); } while((void)0, 0)
 // clang-format on
 
-#define DOCTEST_ASSERT_THROWS(expr, assert_type) DOCTEST_ASSERT_THROWS_WITH(expr, assert_type, "")
-
 #define DOCTEST_ASSERT_THROWS_AS(expr, assert_type, ...)                                           \
     do {                                                                                           \
         if(!doctest::getContextOptions()->no_throw) {                                              \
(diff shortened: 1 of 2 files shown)


onqtam:
Can you paste an example of what the code looks like (with the string shortened from 6.8kb to a few chars)? This might be a problem for other macros as well.

I'll go ahead and merge this since it is a simplification anyway but you'll have to use the dev branch for now

zhihaoy:
> Can you paste an example of what the code looks like (with the string shortened from 6.8kb to a few chars)? This might be a problem for other macros as well.

I did more research and found the cause.  The code used to work because the traditionally MSVC preprocessor only expand one layer of macros in stringification each time you pass it down.

My code looks like this

```
REQUIRE_THROWS(somefunc(
         KWARG(.labels, "x1", "y1"),
         KWARG(.ppg, 0.3),
         ...);
```

Only 5 lines, but each `KWARG` expands to a complicated PEGTL_STRING (strongly typed compile-time string) inside tuple along with some macros -- at this stage, it's 6.8K.  But if all macros are expanded, it's 799K.  The standard seems to require this to be fully expanded.  However, it works in gcc because gcc has no string literal size limit.  It worked in msvc because passing it down once doesn't cause the whole thing to be fully expanded.  It does not work with msvc next-gen preprocessor because it has both the "correct" behavior and the limit (https://devblogs.microsoft.com/cppblog/msvc-preprocessor-progress-towards-conformance/).  So what I should probably is to change the call to a named lambda instead.

> I'll go ahead and merge this since it is a simplification anyway but you'll have to use the dev branch for now

I'm wandering whether there is a way to not to expand the nested macros at all.  Because after all users only want to see the code they wrote, not the code compiler wrote, when encounters a test failure.

zhihaoy:
(Anyhow the change you just merged has its value [although I sent the PR without digging into the issue too deep]: without the change `REQUIRE_THROWS`'s test failure message is different from that of `REQUIRE_THROWS_AS` under msvc preprocessor's "friendly" behavior.)

PR #225, opened 2019-04-15
Remove architecture check from CMake package

mmha:
## Description
CMake adds an architecture check to each package by default. This causes the package to break if e. g. the package was built on a 32 bit system and used on a 64 bit system, even though this is a header-only library. This can happen if a caching package manager like conan is being used.

In CMake 3.14 this check can be disabled by passing `ARCH_INDEPENDENT` to `write_basic_package_version_file`. In versions prior to that `CMAKE_SIZEOF_VOID_P` needs to be unset for the duration of this function.

## GitHub Issues
catchorg/Catch2#1368

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,9 +87,19 @@ set(targets_export_name "${PROJECT_NAME}Targets")
 set(namespace "${PROJECT_NAME}::")
 
 include(CMakePackageConfigHelpers)
+
+# CMake automatically adds an architecture compatibility check to make sure
+# 32 and 64 bit code is not accidentally mixed. For a header-only library this
+# is not required. The check can be disabled by temporarily unsetting
+# CMAKE_SIZEOF_VOID_P. In CMake 3.14 and later this can be achieved more cleanly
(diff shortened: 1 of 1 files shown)


onqtam:
Thanks!

Morwenn:
Is this something that happens despite specifying `header_only` in `package_info` in the conanfile?

mmha:
This happens *because* specifying `header_only` in `package_info`.

When stating to conan that it's a header-only lib, you're telling conan that you can reuse the built package for all architectures and OSes.

The docs say [this](https://docs.conan.io/en/latest/reference/conanfile/methods.html#configure-config-options):
> That means, if someone consumes MyLib with the header_only=True option, the package downloaded and used will be the same, irrespective of the OS, compiler or architecture the consumer is building with.

 But the CMake `write_basic_package_version_file` does add a platform check, which makes CMake `install()`ed header-only libs platform-dependent unless worked around (which is what this PR did).

Morwenn:
Nice & concise explanation, thank you very much!

PR #231, opened 2019-05-01
Explicitly specify the doctest_with_main C++ standard in CMake.

DaanDeMeyer:
This fixes compilation issues on older compilers that don't default to
C++11 yet.


onqtam:
This could even be applied to the interface doctest target as well since people might "link" to it and complain that their code doesn't compile - I've had that already as a complain

DaanDeMeyer:
I don't think it will work on the interface target since the properties will only be defined on the target itself and not propagate to its dependents. To specify it on the interface target, I think we'll need `target_compile_features(doctest INTERFACE cxx_std_11)` but this requires CMake 3.8.

onqtam:
Thanks! and sorry for the late merge :)

PR #232, opened 2019-05-04
Added meson file, to declare a dependency.

jormundgand:
Added a meson build file, so that doctest can be used as a subproject from git.

The build file doesn't actually build the project, but just defines a dependency.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,3 @@
+project('doctest', ['cpp'], version: '2.3.1', meson_version:'>=0.50')
+doctest_dep = declare_dependency(include_directories: include_directories('doctest'))
+


onqtam:
Thanks for the contribution!

Do you think it's possible to get the version programmatically from https://github.com/onqtam/doctest/blob/master/scripts/version.txt ?

jormundgand:
Meson is very big on not being able to change anything programmatically and the project definition has to be the first thing to be defined -- that makes it a tad difficult to do it. Personally, I'm not that worried about the version lagging a bit. Since you're already storing the result of the build in the repo, I guess you could construct the meson build file in the same go.

onqtam:
I'll perhaps generate it

PR #235, opened 2019-05-15
Ignore redundant-decls warning on MinGW

AMS21:
### What
Added warning suppression for -Wredundant-decls on MinGW.

### Why
When compiling with MinGW and in your test file you include windows.h first and then doctest.h, you'll get this -Wredundant-decls warning. Another solution would be to include windows.h to silence the warning. But I think that wouldn't be preferable. [Link to CI build displaying the Problem](https://ci.appveyor.com/project/CppAndre/cpplib/builds/24548210/job/opilwxec2kdj9968#L32)

## GitHub Issues

N/A


onqtam:
Thanks for the fix! Strange that this doesn't happen in the doctest CI builds... Merging it.
Note that since this commit doesn't regenerate the final ```doctest.h``` this fix won't be immediately available in the dev branch.

PR #236, opened 2019-05-15
Add doctest_discover_tests()

RedDwarf69:
Heavily inspired on
https://github.com/catchorg/Catch2/blob/7be35af167f969433d9775f6ddc5d294f69dd4b5/docs/cmake-integration.md#catchcmake-and-catchaddtestscmake

Probably it shouldn't be merged as it is, I'm far from a cmake expert. But I didn't know how to discuss it in https://github.com/onqtam/doctest/issues/171 without actually creating the branch.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -124,6 +124,11 @@ if(NOT ${DOCTEST_NO_INSTALL})
     )
 
     install(
+        FILES "scripts/cmake/doctest.cmake" "scripts/cmake/doctestAddTests.cmake"
+        DESTINATION "${config_install_dir}"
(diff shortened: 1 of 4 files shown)


RedDwarf69:
I had tried the script directly from my project, as was when I was developing it, but hadn't tried it after integrating it with the doctest installation.
I have updated the PR so CMAKE_MODULE_PATH now points to the scripts installation directory and users can just do `include(doctest)` to be able to use `doctest_discover_tests()`.

onqtam:
decided to merge it now and review it later. Thanks for the contribution!

PR #237, opened 2019-05-15
 Reverted 8b29739 

jormundgand:
I mistakenly changed the directory that should be included -- the change in  8b29739 means you have to use #include <doctest/doctest.h>, while it should be #include <doctest.h>. This reverts to the correct behaviour.


onqtam:
I'm a bit confused by this PR - it's against master instead of dev and "Files Changed" shows 0..?!

jormundgand:
My bad -- it looks like I targeted master. It should *naturally* have been dev, where I previously made a wrong change. That change isn't in master yet (luckily), and this just reverts it.

onqtam:
I just committed and pushed that change in dev so I'm closing this PR.

PR #240, opened 2019-05-23
fix #239 - use long macro name

### LINKED ISSUE #239: doctest given defined with short macro name
It should be defined with the long name.
It is not defined otherwise (if no short macro names).

https://github.com/onqtam/doctest/blob/b6dbff7c5576a5f377caa7c5f53f01df3c479827/doctest/doctest.h#L2514

m-bd:
## Description

Use long macro name to define DOCTEST_GIVEN.
 
I should add a test this weekend to check that short names are not used in definitions.

## GitHub Issues

fix onqtam/doctest#239

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -362,7 +362,9 @@ DOCTEST_MSVC_SUPPRESS_WARNING(26444) // Avoid unnamed objects with custom constr
 #elif DOCTEST_MSVC
 #define DOCTEST_BREAK_INTO_DEBUGGER() __debugbreak()
 #elif defined(__MINGW32__)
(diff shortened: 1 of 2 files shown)


onqtam:
I'm not sure how to add tests to check if something doesn't compile - there isn't such infrastructure in the repository yet. I would merge it as it is for now.

onqtam:
seems I can just merge it directly (make the PR from a draft to a ready one). Thanks :)

m-bd:
Welcome! I didn’t have enough time to make it this weekend.
There is a way to do it in CMake (pattern match a failure message). I’ll make a PR later

PR #247, opened 2019-06-23
add alias target for doctest for use in build tree

trondhe:
## Description
Create alias target such that you can link against doctest::doctest either from install or from the same build tree.


onqtam:
Thanks!

PR #254, opened 2019-07-18
Fix #253

### LINKED ISSUE #253: Windows Clang GNU command line warnings
I'm back with more warnings from an even more obscure setup.

CMake 3.15 just released with support for using clang/clang++ on Windows instead of clang-cl. Compiling doctest with this results in a few issues:

1. clang++ on Windows needs to be in C++14 mode to work with MSVC stdlib headers.

We can solve this by setting the `doctest_with_main` standard to C++14 if this specific setup is     used.

2. Some unused macro warnings:

```
../doctest/parts/doctest.cpp:3:9: warning: macro is not used [-Wunused-macros]
#define DOCTEST_LIBRARY_IMPLEMENTATION
        ^
../doctest/parts/doctest.cpp:130:9: warning: macro is not used [-Wunused-macros]
#define VC_EXTRA_LEAN
        ^
```

I'm not sure if these are false positives or actually valid.

3. A non-portable system include warning:

```
../doctest/parts\doctest.cpp:140:10: error: non-portable path to file '<Windows.h>'; specified path differs in case from file name on disk [-Werror,-Wnonportable-system-include-path]
#include <windows.h>
         ^~~~~~~~~~~
         <Windows.h>
```

DaanDeMeyer:
This fixes most of the issues reported in #253. One issue that remains is that `DOCTEST_LIBRARY_IMPLEMENTATION` in doctest.cpp is reported as unused. I'm suspecting it probably isn't so the solution might be to just ignore the warning.

(Note that I had to remove `EXCLUDE_FROM_ALL` from `doctest_with_main` to trigger most of the warnings fixed in this pr).


onqtam:
Thanks for this PR!

However I'll be able to evaluate it in August - these are some pretty busy weeks for me.

Also this issue seems related:
https://github.com/onqtam/doctest/issues/243
(I just wanted to point it out)

DaanDeMeyer:
The cleanest way to solve the C++ version issue in CMake is requiring CMake 3.6 and using `target_compile_features(target PUBLIC cxx_std_11)`. But of course this requires a higher CMake minimum version so its your call.

DaanDeMeyer:
I tried removing the `DOCTEST_LIBRARY_IMPLEMENTATION` and got compilation errors as expected so i just ignore the Wunused-macros waring for `DOCTEST_LIBRARY_IMPLEMENTATION`.

With the last commit, all the issues mentioned in #253 are fixed.

DaanDeMeyer:
Dropped the specific handling of Clang GNU frontend on Windows since CMake is fixing this itself, see https://gitlab.kitware.com/cmake/cmake/issues/19496.

onqtam:
I changed the base from ```master``` to ```dev``` since all PRs should go into ```dev``` - I'll look into releasing a new official version soon which will go into ```master``` :)

PR #256, opened 2019-07-26
only look for C++ compiler with CMake

zhihaoy:
Problem: https://github.com/ContinuumIO/anaconda-issues/issues/9096

For unclear reason, only the C compiler fails CMake's compiler tests.
But since doctest is a header-only C++ library, we can just allow it
to be built with miniconda3's toolchain on MacOSX under SDK>=10.14.

This change can also reduce compiler testing time.


onqtam:
I changed this PR to be against the ```dev``` branch instead of ```master``` and merged it there. A new version will be released at some point in the future.

Thanks for the contribution!

PR #258, opened 2019-07-27
fix : includeing windows.h header caause error

rinechran:
## Description
includeing windows.h header caause error 
because includeing std:: max will call max in windows.h.
## GitHub Issues
Noting

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -3432,7 +3432,7 @@ Approx& Approx::scale(double newScale) {
 bool operator==(double lhs, const Approx& rhs) {
     // Thanks to Richard Harris for his help refining this formula
     return std::fabs(lhs - rhs.m_value) <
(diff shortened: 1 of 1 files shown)


claremacrae:
Thanks very much for this fix, as we had the exact same problem adding doctest support to ApprovalTest.cpp.

The fix almost works - it just has a stray `(`

We suggest just changing `std::max` to `std::max<double>` - this worked for us in both VS2017 and cygwin's gcc.

rinechran:
Thank you for fixing it!!!

claremacrae:
@rinechran You're welcome, but we only fixed it in our copy of the doctest single header...

It remains to be fixed in the master repo, so please could you re-open this issue?

rinechran:
@claremacrae sorry  I first tried to pull request and I accidentally closed the issue.

onqtam:
thanks!

PR #263, opened 2019-08-05
Revert "fix : includeing windows.h header caause error"

onqtam:
Reverts onqtam/doctest#258
https://github.com/onqtam/doctest/commit/92e8615f2bbe64138d3e54c928a4aba34bc71579#commitcomment-34571251

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -3432,7 +3432,7 @@ Approx& Approx::scale(double newScale) {
 bool operator==(double lhs, const Approx& rhs) {
     // Thanks to Richard Harris for his help refining this formula
     return std::fabs(lhs - rhs.m_value) <
-           rhs.m_epsilon * (rhs.m_scale + (std::max<double>(std::fabs(lhs), std::fabs(rhs.m_value)));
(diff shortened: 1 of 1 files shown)


suoniq:
Can you please state the rationale behind this revert? Our clang builds on Windows fail with `-Wnonportable-system-include-path` because of this. Thank you.

onqtam:
@suoniq this revert is not for the windows.h include but for the addition of ```<double>``` in the call to ```std::max``` - look at the changes in the PR. The original PR was broken because it was missing a paren (as can be seen in the changes). As far as I remember we fixed the ```windows.h``` include - does version 2.3.5 not work for you?

suoniq:
> this revert is not for the windows.h include

Then I got distracted by the issue description. Sorry.

> Currently windows.h is included like this: #include <Windows.h> - does version 2.3.5 not work for you? If not - that could easily be changed.

Exactly. It has to be `#include <windows.h>` (small caps) to silence Clang using `[-Werror,-Wnonportable-system-include-path]`. Then it works on Windows/Linux X gcc/clang - each with very picky warning flags. I located three ill-formed includes. Will open a MR fixing that.

PR #270, opened 2019-08-11
Rework `INFO` lazy evaluation to use lambdas.

### LINKED ISSUE #269: Remove INFO() limitation for using only lvalues and no rvalues
reproc has a function `reproc_strerror` that takes a `REPROC_ERROR` functions and converts it into a string representation. Ideally, I could do something like `INFO(reproc_strerror(error));` in order to print a string representation of the error that occurred when a test case fails.

I did not find a way to accomplish this since `INFO` doesn't work with rvalue references. An initial approach to implementing this could be to capture the passed expression in a lambda and calling it when a test case fails and printing the output.

It might be possible to store a lambda like follows in `ContextScope` (which would become a templated class):

```c++
#define DOCTEST_INFO(expression) \
auto DOCTEST_ANONYMOUS(_DOCTEST_CAPTURE_) = \
  make_context_scope([](std::ostream *os) { return os << expression; })
```

and call it when an error occurs. From what I can see from the documentation of `INFO`, this should be backwards compatible if it works.

I'm willing to try implementing this but I wanted to touch the idea off to see if I'm not missing any obvious constraints that prevent implementation of this solution.

DaanDeMeyer:
Instead of capturing everything by pointer, we capture the given
expression into a lazily evaluated lambda which we store as a reference.
This allows passing any expression to `INFO` including rvalues.

Fixes #269.

There's probably more code that can be removed but I wanted to get the initial version out there.

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -1976,13 +1976,13 @@ int registerReporter(const char* name, int priority) {
 
 // for logging
 #define DOCTEST_INFO(expression)                                                                   \
-    DOCTEST_INFO_IMPL(DOCTEST_ANONYMOUS(_DOCTEST_CAPTURE_), expression)
+    DOCTEST_INFO_IMPL(DOCTEST_ANONYMOUS(_DOCTEST_CAPTURE_), DOCTEST_ANONYMOUS(_DOCTEST_CAPTURE_), expression)
 
-#define DOCTEST_INFO_IMPL(lambda_name, expression)                                                 \
+#define DOCTEST_INFO_IMPL(lambda_name, mb_name, expression)                                        \
     auto lambda_name = [&](std::ostream* s) {                                                      \
-        doctest::detail::MessageBuilder mb(__FILE__, __LINE__, doctest::assertType::is_warn);      \
-        mb.m_stream = s;                                                                           \
-        mb << expression;                                                                          \
+        doctest::detail::MessageBuilder mb_name(__FILE__, __LINE__, doctest::assertType::is_warn); \
+        mb_name.m_stream = s;                                                                      \
+        mb_name << expression;                                                                     \
     };                                                                                             \
     doctest::detail::ContextScope<decltype(lambda_name)> DOCTEST_ANONYMOUS(_DOCTEST_CAPTURE_)(lambda_name)
 
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -1544,7 +1544,7 @@ namespace detail {
     public:
         explicit ContextScope(const L &lambda) : lambda_(lambda) {};
 
-        DOCTEST_DELETE_COPIES(ContextScope);
+        ContextScope(ContextScope &&other) : lambda_(other.lambda_) {};
 
         void stringify(std::ostream* s) const override { lambda_(s); }
 
@@ -1570,6 +1570,11 @@ namespace detail {
         bool log();
         void react();
     };
+    
+    template <typename L>
+    ContextScope<L> MakeContextScope(const L &lambda) {
+        return ContextScope<L>(lambda);
+    }
 } // namespace detail
 
 #define DOCTEST_DEFINE_DECORATOR(name, type, def)                                                  \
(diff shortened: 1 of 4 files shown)


doctest/doctest.h
```diff
@@ -2057,9 +1975,17 @@ int registerReporter(const char* name, int priority) {
...
 // for logging
-#define DOCTEST_INFO(x)                                                                            \
-    doctest::detail::ContextScope DOCTEST_ANONYMOUS(_DOCTEST_CAPTURE_)(                            \
-            doctest::detail::ContextBuilder() << x)
+#define DOCTEST_INFO(expression)                                                                   \
+    DOCTEST_INFO_IMPL(DOCTEST_ANONYMOUS(_DOCTEST_CAPTURE_), expression)
+
+#define DOCTEST_INFO_IMPL(lambda_name, expression)                                                 \
+    auto lambda_name = [&](std::ostream* s) {                                                      \
+        doctest::detail::MessageBuilder mb(__FILE__, __LINE__, doctest::assertType::is_warn);      \
```
onqtam: ```mb``` can conflict with a local variable captured with ```[&]``` - that's why I proposed to have the name of the message builder as an argument to the macro as well - like the lambda

DaanDeMeyer: Ahh, I was wondering why that was the case. I'll add it back then.

doctest/parts/doctest_fwd.h
```diff
@@ -1525,107 +1525,25 @@ namespace detail {
...
-                } else {
-                    head = tail = curr;
-                }
-
-                my_memcpy(tail->chunk.buf, &temp, sizeof(Chunk));
-            }
-            ++numCaptures;
-            return *this;
-        }
+    class ContextScopeBase : public IContextScope {
```
onqtam: actually these 2 classes could be merged

doctest/doctest.h
```diff
@@ -1528,107 +1528,25 @@ namespace detail {
...
-        explicit ContextScope(ContextBuilder& temp);
+        explicit ContextScope(const L &lambda) : lambda_(lambda) {};
 
         DOCTEST_DELETE_COPIES(ContextScope);
 
-        ~ContextScope() override;
+        void stringify(std::ostream* s) const override { lambda_(s); }
 
-        void stringify(std::ostream* s) const override;
+        ~ContextScope() override { destroy(); };
```
onqtam: ```destroy()```  could simply be called from the dtor of the base class - and the method could actually be moved entirely into the dtor

doctest/parts/doctest_fwd.h
DaanDeMeyer: I tried doing that but got errors because I was using stuff out of detail in IContextScope, it had something with DOCTEST_CONFIG_DISABLE I think. It also generally required moving stuff around a bit more compared to this solution which is why I went with this.

onqtam: ok, not a problem.

navigating the header around DOCTEST_CONFIG_DISABLE scopes is a bit of a pain...

doctest/doctest.h
DaanDeMeyer: Doing that resulted in a warning that we were calling a pure virtual method (stringify) from the destructor which is dangerous because at that point the derived class destructor has already completed so I went with this instead.

DaanDeMeyer:
I can't reproduce the Appveyor failure locally and the Appveyor debug logs aren't exactly helpful.

(Btw, you might want to apply for the Github actions beta (https://github.com/features/actions). If it turns out to be any good I'll likely move reproc's CI entirely to that and might do the same for doctest if it supports all necessary features).

doctest/doctest.h
onqtam: true... hadn't thought of that. sorry

DaanDeMeyer: I'll add a comment explaining why we do not inline destroy into the destructor for future reference.

onqtam:
I just noticed in the appveyor log that it is instantiating ```ContextScope``` with ```int``` o_O
https://ci.appveyor.com/project/onqtam/doctest/builds/26618380/job/edi528mpqshj2fj6#L132

I could test it locally at home with MSVC.

I signed up for Actions, thanks.

DaanDeMeyer:
Seems like MSVC just didn't like `decltype` on lambdas. I added a `MakeContextScope` instead to deduce the lambda type via template function argument deduction and that seems to work.

onqtam:
I guess clang will again complain about the extra semicolon in some places: ```{};```

apart from that - this will probably work!

DaanDeMeyer:
I made `DOCTEST_TO_LVALUE` a noop but didn't remote it to keep everything backwards compatible.

DaanDeMeyer:
That last commit only updates documentation so you don't necessarily have to wait for the entire build to complete again. The previous build completed successfully.

onqtam:
good work - thanks for this! :)

PR #283, opened 2019-09-02
No minimum version limitation of Meson

ydm:
In order to support building a project using doctest with older versions of Meson, the minimum version limitation should be either further lowered or dropped at all.


onqtam:
Thanks! Is this for your company? : )

PR #285, opened 2019-09-08
WIP: Add Github Actions CI

DaanDeMeyer:
Opening the pull request so I can start testing the configuration. I've started by taking reproc's Github Actions configuration and removing everything reproc related. Once that works I'll start adding all doctest related configuration.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,23 +13,23 @@ jobs:
         # Github Actions requires a single row to be added to the build matrix.
         # See https://help.github.com/en/articles/workflow-syntax-for-github-actions.
         name: [
-          ubuntu-18.04-gcc-4.8,
-          ubuntu-18.04-gcc-4.9,
-          ubuntu-18.04-gcc-5,
-          ubuntu-18.04-gcc-6,
-          ubuntu-18.04-gcc-7,
-          ubuntu-18.04-gcc-8,
-          ubuntu-18.04-gcc-9,
-          ubuntu-18.04-clang-3.5,
-          ubuntu-18.04-clang-3.6,
-          ubuntu-18.04-clang-3.7,
-          ubuntu-18.04-clang-3.8,
-          ubuntu-18.04-clang-3.9,
-          ubuntu-18.04-clang-4.0,
-          ubuntu-18.04-clang-5.0,
-          ubuntu-18.04-clang-6.0,
-          ubuntu-18.04-clang-7,
-          ubuntu-18.04-clang-8,
+          ubuntu-16.04-gcc-4.8,
+          ubuntu-16.04-gcc-4.9,
+          ubuntu-16.04-gcc-5,
+          ubuntu-16.04-gcc-6,
+          ubuntu-16.04-gcc-7,
+          ubuntu-16.04-gcc-8,
+          ubuntu-16.04-gcc-9,
+          ubuntu-16.04-clang-3.5,
+          ubuntu-16.04-clang-3.6,
+          ubuntu-16.04-clang-3.7,
+          ubuntu-16.04-clang-3.8,
+          ubuntu-16.04-clang-3.9,
+          ubuntu-16.04-clang-4.0,
+          ubuntu-16.04-clang-5.0,
+          ubuntu-16.04-clang-6.0,
+          ubuntu-16.04-clang-7,
+          ubuntu-16.04-clang-8,
           windows-2016-cl,
           windows-2016-clang-cl,
           windows-2016-clang,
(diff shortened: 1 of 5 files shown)


DaanDeMeyer:
Current status can be followed here: https://github.com/DaanDeMeyer/doctest/actions

DaanDeMeyer:
Do you want each platform to use its native build system? I'm using Ninja everywhere at the moment which is extremely convenient but it does reduce the testing surface as we're only building with a single CMake generator.

On the other hand, doctest is a header-only library so the build system used to compile tests and examples shouldn't matter to the end user.

(I'm strongly in favor of just using Ninja everywhere as it makes my life a whole lot easier)

DaanDeMeyer:
Are clang 3.5, 3.6 and 3.7 required for CI? LLVM does not provide Xenial apt repositories for those versions. If we want those I'll have to start juggling with ubuntu versions in the LLVM apt repository url's which is not impossible but makes stuff a little bit more complex.

DaanDeMeyer:
Do I have to keep all the sanitizer builds separate? There's a comment in the Travis config on knowing exactly what caused the test run to fail but can't you see that from the sanitizer output?

onqtam:
Hmmm I missed your comments here.
- Using only ninja - absolutely!
- if some compiler versions are a problem they could be skipped.
- about the sanitizers - I don't understand the problem. If a sanitizer build fails for some reason the entire build should fail. If a sanitizer fails because of some bug in the compiler/sanitizer itself then that sanitizer is disabled for that build job.

DaanDeMeyer:
Yes, but currently you seem to do a separate build for each specific sanitizer in travis.yml (https://github.com/onqtam/doctest/blob/f13a00cc27ed3c1ec4f755572ab7556c4cb01716/.travis.yml#L372). I was wondering if I couldn't just do a single build with all sanitizers (asan, tsan and ubsan) combined (per compiler of course).

onqtam:
I think ASAN and UBSAN can be combined, but am really not sure about TSAN... But I'm also not sure we need debug vs release builds with the sanitizers... perhaps debug is enough?

DaanDeMeyer:
I'm wondering if we can't just go one step further and do a single build with everything enabled (including static analysis later). For reproc I just do a debug build in CI instead of both debug and release. Is it often that there are differences in debug/release builds in terms of test or build results?

onqtam:
the most common differences between debug/release have been warnings, so both build types are important.

DaanDeMeyer:
I'm kind of in an annoying situation where I don't want to make separate jobs for debug/release builds as it causes the job count to balloon but doing both in the same job requires an annoying amount of duplication because I can't use the same steps for Windows and POSIX due to Github Actions limitations and as a result have to add 3 steps each time I want to do something extra in a single job instead of a single one.

I've reported the issue (https://github.community/t5/GitHub-Actions/Support-global-environment-variables/m-p/30894#M601) but I'm not sure if or when it will be fixed.

onqtam:
Let me know whenever this is ready for a review :)

DaanDeMeyer:
Still need to add a lot of stuff. I've been on vacation this week so I didn't really have time to finish it yet.

There's still a bit of duplication involved since we have to configure, build and test quite a few configurations so I might look into moving stuff into an actual Action so we can build a cross product of combinations with Javascript and reuse code for all those combinations.

DaanDeMeyer:
Also, while we do have 20 concurrent jobs, it does take quite a while for the second batch of remaining jobs to start after the first 20 complete so if we can remove some combinations in order to get the total amount of jobs down to 20 that would reduce waiting times for CI to complete somewhat.

Of course, it depends on whether there are configurations that can be dropped. If they are all important, there's always the chance that the speed of Github Actions improves later on as its still in beta.

onqtam:
I just checked and there are 31 different compiler jobs, so if all other actions/steps/builds are within a single per-compiler job then we would be left with 31 jobs in total, so 2 batches of GitHub Actions. I'm not sure it's worth optimizing around the '20' parallel jobs of Actions since that number could change in any direction in the future, and dropping compilers isn't ideal since we want as much testing as possible to minimize client issues/reports.

onqtam:
I just force-pushed into dev (after rebasing it on top of master and rebranching dev from master) and it seems that didn't play well with this PR and there are some conflicts

DaanDeMeyer:
Most of the configurations (except macOS) are doing x86 now. GCC on Windows is giving some access violations when running the tests that I still need to look into fixing. 

There's also a test failure on macOS when compiling with gcc 6 (https://github.com/DaanDeMeyer/doctest/runs/231858542) which might indicate a bug (either in gcc or in doctest).

Next after this is getting sanitizers working.

DaanDeMeyer:
I'm having a hard time finding the causes of the failure on macos with gcc-6 and the mingw x86 failures. 

I'm pretty sure the gcc-6 failure on macos is just due to a GCC bug as gcc 7, 8 and 9 have no issues. I can't find anything about it online though. Is it fine to just remove that one? It's a bit of a pain to debug as I don't have a macos machine on hand.

I think the MinGW x86 failures are either due to the version of MinGW targeting x86 using a different exception handling model or it's linking against x64 stuff somewhere in the build. 

I'm not sure if it's worth it to fix this stuff until we know that someone's actually using it.

DaanDeMeyer:
Got a bit further with the MinGW errors on Windows. gdb reports:

```
Program received signal SIGSEGV, Segmentation fault.
0x6fee65e9 in libstdc++-6!_ZNSt7__cxx1119basic_ostringstreamIcSt11char_traitsIcESaIcEED1Ev () from C:\Users\daan\scoop\apps\gcc\current\bin\libstdc++-6.dll
```

Which happens after the main function returns. The debugger callstack shows this happens in the `RtlFreeUserStack` function.

onqtam:
Pretty sure those are toolchain issues (compiler bugs)... This might be related to the thread local globals - one of them is an ostringstream. You could try defining ```DOCTEST_THREAD_LOCAL``` to nothing (like it's being done for the XCode 6.4 & 7.3 builds on travis) and see if it continues to fail (but then the ```concurrency.cpp``` tests will probably not work correctly).

Right now on appveyor I build only with a single version of mingw (GCC 8.1) and only in 64 bit mode - I think it's fair to drop x86 mingw builds on Windows and not even try ```DOCTEST_THREAD_LOCAL```.

About the macos issues... idk :)
I've had really weird exception issues on macos - especially when using gcc from homebrew... I've actually dropped gcc builds on MacOS in my travis config due to the enormous amount of wasted effort.

DaanDeMeyer:
Any idea on [this](https://github.com/DaanDeMeyer/doctest/runs/240726643) one?

onqtam:
seems like the exception translator from plugin.cpp for char doesn't get picked up... What did you change? This must have worked previously, and this seems like the normal Debug build - before the sanitizer tasks run...

DaanDeMeyer:
I added the sanitizers to the normal builds. I didn't do anything but add `-DCMAKE_CXX_FLAGS="-fsanitize=address,undefined -fno-omit-frame-pointer -g"` to the CMake invocations.

onqtam:
I guess all the steps are executed in a sequential order and what has been configured in "Configure x64 Sanitizers" is later used in "Build & Test Debug x64".

One thing I would try is to build only with asan or ubsan, and not both at the same time (even though it should be possible).

I also see that in the travis config I set ```SANITIZER_CXX_FLAGS``` to be ```-fuse-ld=gold``` for all GCC versions under linux - I don't remember anymore why I've done that - I should have left comments... (perhaps git log/blame could help).

DaanDeMeyer:
I know I initially had to add fuse-ld=gold to make GCC sanitizer builds work for reproc as well but I was able to remove that recently without issues.

It's also extremely weird that the GCC 4.9 build does succeed but all the others fail.

I tried with separate undefined and address which revealed that it's the address sanitizer that's causing the issue.

Unfortunately, I'm not able to reproduce this locally on Arch Linux with GCC 9.1 which means I can't run it with a debugger to see what's going on.

onqtam:
The most important thing is to compile and run the tests on as much compilers/platforms as possible. Sanitizers don't need to be enabled everywhere - there are plenty of toolchain issues which are just a waste of time and effort... Most projects settle with 1-2 configurations/compilers being tested with sanitizers and even if we disable some sanitizers for 1/2 of the compilers it would still be more than enough - so feel free to cut out the configs which are problematic, and for those which still require the gold linker - use that if it's easy.

DaanDeMeyer:
I'm going to try a little bit more with this one as tests failing on LTS Ubuntu with all major versions of GCC does not seem exactly confidence inspiring even if its an edge case feature.

I setup a Ubuntu bionic container using systemd-nspawn and managed to reproduce the issue in there. It turns out the tests aren't being registered simply because the call to dlopen is failing in main.cpp in executable_plugin_and_dll. I still need to find out exactly why this is the case when sanitizers are enabled in GCC (and why it only happens on Ubuntu).

onqtam:
FYI someone has already had a problem with GCC/sanitizers/plugins: https://github.com/onqtam/doctest/issues/201

DaanDeMeyer:
Adding `-static-libasan` to the compiler flags fixes the issue. I'm guessing the `dlopen` search paths somehow differ from what's done normally.

Anyway, I'll just add `-static-libasan` to the compiler flags.

DaanDeMeyer:
address/undefined behaviour sanitizers are now enabled where possible. Should I add TSAN as well? I've always wondered why that one was needed for doctest.

onqtam:
TSAN is necessary for the [```concurrency.cpp```](https://github.com/onqtam/doctest/blob/master/examples/all_features/concurrency.cpp) tests where there are asserts and logging from multiple threads.

onqtam:
I'll take a look after at least 1 week since I have a really tight deadline for a presentation. I'll let you know as soon as possible. On the point about static analysis - I completely agree. Thanks for the effort - great work!

.github/workflows/main.yml
```diff
@@ -0,0 +1,500 @@
...
+             [ $(version ${{ matrix.version }}) -le $(version "3.9") ]; then
+            exit 0
+          fi
+
+          if [ "${{ runner.os }}" = "macOS" ] && \
+             [ "${{ matrix.compiler }}" = "gcc" ]; then
+            exit 0
+          fi
+
+          TSAN_FLAGS="-fsanitize=thread -pie -fPIE"
```
onqtam: I wonder if -fno-omit-frame-pointer is helpful for the TSAN flags as well

DaanDeMeyer: The docs don't seem to mention it so I don't think it's relevant. Though if we get garbled stacktraces we'll immediately know whether we have to enable it.

onqtam: fair enough :)
I still haven't looked at this in-depth - this was just a random comment.

.github/workflows/main.yml
```diff
@@ -0,0 +1,500 @@
...
+
+          if [ "${{ runner.os }}" = "Linux" ] && [ "${{ matrix.compiler }}" = "gcc" ]; then
+            TSAN_FLAGS="$TSAN_FLAGS -static-libtsan"
```
onqtam: a comment explaining why this was necessary would be helpful in the future

onqtam:
This is great - I'm merging it as it is.

onqtam:
it would be good to have ```--version``` called on the current compiler (whatever it is - I guess for cl.exe it's a different switch or it could just be called directly without a switch) at the start of each build after installing the compiler - just to be absolutely pedantic and aware which compiler version was used.

.github/workflows/main.yml
```diff
@@ -0,0 +1,500 @@
...
+          if [ "${{ runner.os }}" = "macOS" ] && \
+             [ "${{ matrix.compiler }}" = "gcc" ]; then
+            exit 0
+          fi
+
+          ASAN_UBSAN_FLAGS="-fsanitize=address,undefined -fno-omit-frame-pointer"
+          
+          # Link statically against ASAN libraries because dynamically linking 
+          # against ASAN libraries causes problems when using dlopen on Ubuntu.
+          # See: https://github.com/DaanDeMeyer/doctest/runs/249002713
```
onqtam: I guess this is how the following issue could be resolved: https://github.com/onqtam/doctest/issues/201

onqtam:
> it would be good to have `--version` called on the current compiler (whatever it is - I guess for cl.exe it's a different switch or it could just be called directly without a switch) at the start of each build after installing the compiler - just to be absolutely pedantic and aware which compiler version was used.

nevermind - it's all in the detailed logs of the build because of the cmake config step :)

- ```The CXX compiler identification is MSVC``` ... for VS
- ```The CXX compiler identification is GNU``` for gcc and something similar for clang

.github/workflows/main.yml
DaanDeMeyer: It's the same explanation as for why we add -static-libasan. I didn't want to repeat the comment twice but could add a link to it if needed.

onqtam: no need :)

DaanDeMeyer:
I did notice that out of habit I did everything with 2 spaces indentation while the rest of doctest uses 4 space indentation. Do you want me to switch everything to 4 spaces? I'd also recommend adopting a .editorconfig file in the repository which allows most editors to pick up this stuff automatically.

onqtam:
4 spaces sounds good (the 2 spaces thing I tried to stick to in travis.yml annoyed me quite a bit and I never even tried with 4 spaces...)

an editorconfig file would be fine as well :)

PR #286, opened 2019-09-08
Add support for compiling with clang-cl

DaanDeMeyer:
The previous logic was adding GNU style command line flags when clang-cl was used which wants MSVC style command line flags. I've changed the logic to the same logic used in reproc to support clang-cl which fixes the issue.

### DIFF
--- a/scripts/cmake/common.cmake
+++ b/scripts/cmake/common.cmake
@@ -72,10 +72,14 @@ endmacro()
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
     add_compiler_flags(-Werror)
-    add_compiler_flags(-pedantic)
-    add_compiler_flags(-pedantic-errors)
-    add_compiler_flags(-fvisibility=hidden)
     add_compiler_flags(-fstrict-aliasing)
+
+    # The following options are not valid when clang-cl is used.
+    if(NOT MSVC)
+        add_compiler_flags(-pedantic)
+        add_compiler_flags(-pedantic-errors)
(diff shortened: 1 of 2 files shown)


DaanDeMeyer:
I also had to disable a few warnings to get everything (the examples) to compile without warnings.

DaanDeMeyer:
Turns out the warnings were already disabled somewhere else. I changed the logic to reuse that instead of disabling the same warnings twice.

scripts/cmake/common.cmake
```diff
@@ -72,10 +72,14 @@ endmacro()
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
     add_compiler_flags(-Werror)
-    add_compiler_flags(-pedantic)
-    add_compiler_flags(-pedantic-errors)
-    add_compiler_flags(-fvisibility=hidden)
     add_compiler_flags(-fstrict-aliasing)
+
+    # The following options are not valid when clang-cl is used.
+    if(UNIX)
```
onqtam: I guess this check could be ```UNIX OR CMAKE_CXX_COMPILER_ID MATCHES "GNU"``` since we use MinGW on windows

DaanDeMeyer:
`NOT MSVC` should handle every case.

scripts/cmake/common.cmake
DaanDeMeyer: I did `NOT MSVC` which should handle all cases. It might be too extensive but we can solve that once we actually encounter issues with this approach.

PR #288, opened 2019-09-10
Add -order-by=name to filter_2 test

DaanDeMeyer:
The default -order-by=file is compiler-dependent which currently causes the filter_2 test to fail depending on the compiler used. By using -order-by=suite, the test always succeeds regardless of the compiler used.

This PR also updates the filter_2_xml test output to the new output produced by adding the -order-by=suite option.

See https://github.com/onqtam/doctest/issues/287


DaanDeMeyer:
@onqtam I'll keep making pr's for now and letting CI run before merging so you can add comments and what not if needed.

onqtam:
will wait for the builds before pushing - just in case

PR #289, opened 2019-09-10
Document that -order-by=file is compiler-dependent

DaanDeMeyer:


onqtam:
Btw the .html files are generated only when doing a release of doctest (with some of the python scripts) so I don't bother updating them when iterating over the .md files. I also wonder if the .html docs are even worth it...

DaanDeMeyer:
Anecdotally, I've never used them.

PR #308, opened 2019-11-11
fix warnings from PVS Studio

cyyever:


onqtam:
Hi!

well the PR cannot be accepted in its current form - mainly because an include has been moved from the implementation to the interface header (```#include <fstream>```) and also because the builds don't pass. Is there a way to suppress PVS studio warnings in the code - with some comment?

onqtam:
I guess removing and recreating a remove ```dev``` branch triggered this to get closed - my bad.

TheQwertiest:
> Is there a way to suppress PVS studio warnings in the code - with some comment?

PVS provides a few ways to suppress warnings. The simplest one is to add the comment with the warning number after the problematic line, e.g. 
```cpp
obj.specialFunc(obj); //-V678
```

Source: https://www.viva64.com/en/m/0017/

PR #325, opened 2020-01-14
Switch to catching unhandled exceptions on Windows Closes #324

### LINKED ISSUE #324: Doctest SEH handlers are called before __except handlers
## Description
Doctest will report exceptions even if code exists that uses Windows `__try`/`__except` Structured-Exception Handlers.

### Steps to reproduce
This test case will report a SIGSEGV even though the crash is handled by SEH:
```
#ifdef _MSC_VER
static void NoCrash()
{
    __try
    {
        int* p = 0;
        *p = 1;
    }
    __except (1)
    {
        // Caught!
    }
}

TEST_CASE("SEH Avoided Crash")
{
    NoCrash();
    CHECK(1);
}
#endif
```

### Extra information

* doctest version: **v2.3.6**
* Operating System: **Windows 10 x64**
* Compiler+version: **MS Visual Studio 2017**

jkriegshauser:
## GitHub Issues
Closes #324

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,24 +24,6 @@
 *.ii
 *.s
 
-# Visual Studio garbage
-.vs/
-*.sln
-*.vcxproj
-*.vcxproj.filters
-*.vcxproj.user
-*.log
-*.tlog*
-*.ilk
-*.pdb
-*.exp
-
-# CMake garbage
-*.rule
-*.cmake
-generate.stamp*
-CMakeFiles/
-
 # My garbage :)
 Win32/
 x64/
@@ -55,6 +37,3 @@ build-mingw/
 *vscode*
 .idea/
 cmake-build-*/
-CMakeCache.txt
-examples/all_features/temp
-Testing/Temporary/
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,24 @@
 *.ii
 *.s
 
+# Visual Studio garbage
+.vs/
+*.sln
+*.vcxproj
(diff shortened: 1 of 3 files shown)


onqtam:
Hi - thanks for this fix! Can you confirm that doctest continues to capture correctly the crash from code such as ```int* p = 0; *p = 1;``` if there isn't a ```__try``` section in the test case?

Btw I suppose this same issue is present with Catch2, since the exception handling was [taken from there](https://github.com/catchorg/Catch2/blob/master/include/internal/catch_fatal_condition.cpp).

I don't think the changes made to ```.gitignore``` should be part of this PR (especially adding the ```*.cmake``` wildcard) - making an out-of-source build should be preferred.

jkriegshauser:
Yes, in the case you mention, a SIGSEGV continues to be caught and reported by doctest as it was before the fix.

I can revert the changes to .gitignore. For out-of-source builds, what parameters are you running CMake with to build out-of-source. I'd recommend adding a building guide to CONTRIBUTING.md, and include info about how to build, run tests, add tests, etc. :)

onqtam:
About the out-of-source builds - I usually make a ```build``` directory within the doctest directory and call cmake from there like this: ```cmake ..``` - standard CMake practice - that ```build``` directory already has an entry in the ```.gitignore```, but it can be outside of the doctest folder as well - all you need to do is pass the path the root of the doctest repository do CMake. Perhaps I should indeed add some info in there :)

I'm merging this. Thanks!

PR #328, opened 2020-01-29
Support std::string for SUBCASE

jkriegshauser:
## Description
Allow SUBCASE to use std::string operators

## GitHub Issues
We're trying to convert from Catch2 to doctest, and have a lot of cases like this:
`SUBCASE("Test " + name) {}`
Where name is a std::string. This works fine in Catch2 and I'd like to do the same in doctest.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,24 +24,6 @@
 *.ii
 *.s
 
-# Visual Studio garbage
-.vs/
-*.sln
-*.vcxproj
-*.vcxproj.filters
-*.vcxproj.user
-*.log
-*.tlog*
-*.ilk
-*.pdb
-*.exp
-
-# CMake garbage
-*.rule
-*.cmake
-generate.stamp*
-CMakeFiles/
-
 # My garbage :)
 Win32/
 x64/
@@ -55,6 +37,3 @@ build-mingw/
 *vscode*
 .idea/
 cmake-build-*/
-CMakeCache.txt
-examples/all_features/temp
-Testing/Temporary/
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,24 @@
 *.ii
 *.s
 
+# Visual Studio garbage
+.vs/
+*.sln
+*.vcxproj
+*.vcxproj.filters
+*.vcxproj.user
+*.log
+*.tlog*
+*.ilk
+*.pdb
+*.exp
+
+# CMake garbage
+*.rule
+*.cmake
+generate.stamp*
+CMakeFiles/
+
 # My garbage :)
 Win32/
 x64/
@@ -37,3 +55,6 @@ build-mingw/
 *vscode*
 .idea/
 cmake-build-*/
+CMakeCache.txt
+examples/all_features/temp
+Testing/Temporary/
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -675,7 +675,7 @@ struct DOCTEST_INTERFACE MessageData
 
 struct DOCTEST_INTERFACE SubcaseSignature
 {
-    std::string m_name;
+    String      m_name;
     const char* m_file;
     int         m_line;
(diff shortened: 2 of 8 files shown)


onqtam:
Hi @jkriegshauser ,

I probably won't accept this since it's including ```<string>``` in the interface part of the library - that's against [the design](https://github.com/onqtam/doctest/blob/master/doc/markdown/features.md) and one of the reasons doctest compiles so fast.

However, if we indeed change ```const char* m_name;``` to ```String m_name;``` you could do what you want like this without doctest having to include ```<string>```:

```c++
#undef SUBCASE
#define SUBCASE(...) DOCTEST_SUBCASE(std::string(__VA_ARGS__).c_str())
```

jkriegshauser:
Hey there, thanks for the feedback. I've pushed changes (had to `pragma`
around a warning in GCC) that do what you suggest and will watch the build
to see how they fare.

jkriegshauser:
@onqtam I'm not sure why it's failing on GCC 5 _only_ for the warning that there is a pragma for.
https://travis-ci.org/onqtam/doctest/jobs/644240138

All other versions of GCC 5 seem okay, and a [simple test case on godbolt](https://godbolt.org/z/ne6aY_) seems to compile just fine for GCC 5. Is it possible something is misconfigured for the build?

onqtam:
Hey, I still haven't gotten to reviewing this, but I guess I should explain what I meant in my previous comment: ```<string>``` cannot be included in ```doctest_fwd.h``` but I see that it's still being included there in this PR... The changes I proposed (using ```doctest::String```) would allow you to define the ```SUBCASE``` macro as I showed in the last comment - using ```std::string```, but that's not something this library can do in the _fwd.h header.

jkriegshauser:
@onqtam Thanks for the response. I though I had already removed `#include <string>` but I hadn't pushed it yet. It's pushed now. This of course requires that `<string>` be included wherever `SUBCASE` is used. I'll watch to see how GCC 5 behaves, but I suspect it will fail again.

jkriegshauser:
@onqtam as suspected, GCC 5 failed again. Can you look at the configuration and see why it might be failing but GCC 4 and GCC 6 are fine? The warning-as-error that it's failing on should be disabled via pragma.

onqtam:
Thanks for the PR. I decided to implement this and close the PR since there were so many changes in it - seems you did follow the instructions for contributing about formatting the changes but turns out I myself haven't been doing that lately - so there were a lot of unrelated changes (I really ought to follow my own guidelines...).

One issue with the PR in its current form is that ```std::string``` was still being used in the ```SUBCASE``` macro - even though the ```<string>``` header was no longer being included - this means that doctest users would need the ```<string>``` header if they used subcases - even if they didn't need ```std::string```.

I just released version 2.3.7 and there you can do the following:
https://github.com/onqtam/doctest/blob/2.3.7/examples/all_features/subcases.cpp#L144

```c++
#undef SUBCASE
#define SUBCASE(...) DOCTEST_SUBCASE(std::string(__VA_ARGS__).c_str())

TEST_CASE("subcases with changing names") {
    for(int i = 0; i < 2; ++i) {
        SUBCASE("outer " + std::to_string(i)) {
            for(int k = 0; k < 2; ++k) {
                SUBCASE("inner " + std::to_string(k)) {
                    MESSAGE("msg!");
                }
            }
        }
    }
    SUBCASE("separate") {
        MESSAGE("separate msg!");
    }
}
```

this way the doctest header has nothing ```std::string```-related but can now support runtime values for the name of subcases. All you have to do is change the definition of the ```SUBCASE``` macro locally - or call it like this: ```SUBCASE(std::string("inner " + std::to_string(i)).c_str())```.

Sorry for the delay - hope this helps!

jkriegshauser:
Awesome, thanks! I think this will work!

PR #340, opened 2020-02-18
fix std::uncaught_exceptions

### LINKED ISSUE #333: C++20 removed std::uncaught_exception
`std::uncaught_exception` was deprecated in C++17 and removed in C++20. [Section [diff.cpp17.depr]/1](https://eel.is/c++draft/diff.cpp17.depr#1) of the C++20 Working Draft explains:

> **Change:** Remove `uncaught_­exception`.
> **Rationale:** The function did not have a clear specification when multiple exceptions were active, and has been superseded by `uncaught_­exceptions`.
> **Effect on original feature:** A valid C++ 2017 program that calls `std​::​uncaught_­exception` may fail to compile. It might be revised to use `std​::​uncaught_­exceptions` instead, for clear and portable semantics.

This function is being used here:

https://github.com/onqtam/doctest/blob/503de03ee2cbbfc58c2063dce36dc96f917f2da3/doctest/parts/doctest.cpp#L929-L939

https://github.com/onqtam/doctest/blob/503de03ee2cbbfc58c2063dce36dc96f917f2da3/doctest/parts/doctest.cpp#L1273-L1280

And similarly in [doctest/doctest.h](https://github.com/onqtam/doctest/blob/503de03ee2cbbfc58c2063dce36dc96f917f2da3/doctest/doctest.h) (which appears to be auto-generated).

Now that MSVC's STL has merged https://github.com/microsoft/STL/pull/380, doctest will fail to compile in `/std:c++latest` mode with VS 2019 16.6 Preview 2 (not yet released).

We (the MSVC team) are seeing this because we test a number of open-source projects with development builds of the MSVC toolset. (We encountered this case via https://github.com/lewissbaker/cppcoro/blob/master/test/doctest/doctest.h .) This allows us to find and fix bugs in our toolset before they can affect you, and it also allows us to provide advance notice of source-breaking incompatibilities like this one.

MSVC provides "escape hatch" macros that can be defined project-wide to suppress the deprecation warning and restore the removed type trait. (Specifically, compiling with `/D_SILENCE_CXX17_UNCAUGHT_EXCEPTION_DEPRECATION_WARNING` and `/D_HAS_DEPRECATED_UNCAUGHT_EXCEPTION=1`.) 

However, libraries like doctest don't have control over their programmer-users' project-wide settings, so global escape hatches aren't feasible. As you've discovered, the deprecation warning can be locally silenced by push-disable-popping warning C4996, but the removal can't be locally restored. Therefore, the deprecated/removed machinery should be entirely avoided if possible.

Fortunately, a Standard feature-test macro is available to detect when the replacement `std::uncaught_exceptions` (plural) can be called. You can activate that based on `#ifdef __cpp_lib_uncaught_exceptions` and otherwise fall back to the older function. This will allow doctest to work in all language modes.

cyyever:
Closes #333

### DIFF
--- a/doctest/parts/doctest.cpp
+++ b/doctest/parts/doctest.cpp
@@ -926,9 +926,6 @@ namespace detail {
         DOCTEST_ITERATE_THROUGH_REPORTERS(subcase_start, m_signature);
     }
 
-    DOCTEST_MSVC_SUPPRESS_WARNING_WITH_PUSH(4996) // std::uncaught_exception is deprecated in C++17
-    DOCTEST_GCC_SUPPRESS_WARNING_WITH_PUSH("-Wdeprecated-declarations")
-    DOCTEST_CLANG_SUPPRESS_WARNING_WITH_PUSH("-Wdeprecated-declarations")
     Subcase::~Subcase() {
         if(m_entered) {
             // only mark the subcase stack as passed if no subcases have been skipped
@@ -936,7 +933,12 @@ namespace detail {
                 g_cs->subcasesPassed.insert(g_cs->subcasesStack);
             g_cs->subcasesStack.pop_back();
 
-            if(std::uncaught_exception() && g_cs->shouldLogCurrentException) {
+#if __cplusplus >= 201703L && defined(__cpp_lib_uncaught_exceptions) && __cpp_lib_uncaught_exceptions >= 201411
+            if(std::uncaught_exceptions() > 0
+#else
+            if(std::uncaught_exception()
+#endif
+            && g_cs->shouldLogCurrentException) {
                 DOCTEST_ITERATE_THROUGH_REPORTERS(
                         test_case_exception, {"exception thrown in subcase - will translate later "
                                               "when the whole test case has been exited (cannot "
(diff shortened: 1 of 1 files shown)


cyyever:
May be possible to remove surrounding SUPPRESS_WARNING_WITH_PUSH

onqtam:
I think the ```#if __cplusplus >= 201703L``` conditional should be reversed to ```#if __cplusplus < 201703L``` - https://en.cppreference.com/w/cpp/error/uncaught_exception

cyyever:
> 
> 
> I think the `#if __cplusplus >= 201703L` conditional should be reversed to `#if __cplusplus < 201703L` - https://en.cppreference.com/w/cpp/error/uncaught_exception

I try to use __cplusplus and Feature-Test Macro together to avoid error-prone compiler version checks.

cyyever:
> 
> 
> I think the `#if __cplusplus >= 201703L` conditional should be reversed to `#if __cplusplus < 201703L` - https://en.cppreference.com/w/cpp/error/uncaught_exception

Sorry, I made a stupid error. I should exchange the if statements.

cyyever:
One test case failed, but it seems unrelated to the PR.

onqtam:
I merged it into the ```dev``` branch - will soon release an official version in ```master```. Thanks for this and sorry for the delay - been busy lately!

PR #341, opened 2020-02-18
using std namespace where necessary

oktonion:
## Description
Fix of a mess with using different std functions from global namespace and `stricmp` ambiguity with standard library extensions.

### DIFF
--- a/doctest/parts/doctest.cpp
+++ b/doctest/parts/doctest.cpp
@@ -388,6 +388,7 @@ void String::setOnHeap() { *reinterpret_cast<unsigned char*>(&buf[last]) = 128; 
 void String::setLast(unsigned in) { buf[last] = char(in); }
 
 void String::copy(const String& other) {
+    using namespace std;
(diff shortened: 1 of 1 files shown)


oktonion:
Also fixed win ticks and timer ticks in general with correct types.

doctest/parts/doctest.cpp
```diff
@@ -290,10 +301,10 @@ namespace detail {
         //unsigned int getElapsedMilliseconds() const {
         //    return static_cast<unsigned int>(getElapsedMicroseconds() / 1000);
         //}
-        double getElapsedSeconds() const { return getElapsedMicroseconds() / 1000000.0; }
+        double getElapsedSeconds() const { return (getCurrentTicks() - m_ticks) / 1000000.0; }
```
onqtam: shouldn't we keep the division by 1 million? I don't see ```getCurrentTicks()``` being changed in terms of behavior...

onqtam:
Thanks for this!

PR #352, opened 2020-04-03
drop __cplusplus check because MSVC handles it incorrectly

cyyever:
Refer to this blog https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/ for the reason.


onqtam:
I'll apply the important changes (the ones regarding `__cplusplus`) from this PR manually since there are some other changes in there which seem unrelated (about `Wdeprecated-declarations`)

PR #363, opened 2020-04-29
Use CMake's CMP0077 policy if available

thelink2012:
## Description

Using the project as an `add_subdirectory` (and setting options with `set` before doing so) produces an warning regarding policy [CMP0077](https://cmake.org/cmake/help/git-stage/policy/CMP0077.html).

This happens because the behavior of using `set`  on variables which will be cached (by `option`) has changed in CMake 3.12. Before, the value set would be unset, and now the value set is honored.

The project should have `CMP0077` set to `NEW` to ensure it understands this change (the old behaviour is deprecated and may be removed in the future). Doing this outside doctest tree (i.e. the tree that performs `add_subdirectory` on it) has no effect.

## GitHub Issues

Didn't find it necessary. The change is minimal and any necessary discussion can take place in the PR.


onqtam:
Thanks!

Issue #215, opened 2019-03-27
Data-driven testing -- print out the deepest DOCTEST_SUBCASE


jktjkt:
## Description

I'm trying to use what the Qt landscape calls a ["data driven testing"](https://doc.qt.io/qt-5/qttestlib-tutorial2-example.html). Here's my code:

```c++
TEST_CASE("converting a byte buffer to a float")
{
    float f;
    std::string buffer;

    DOCTEST_SUBCASE("0x41480000 (= 12.5)") { buffer = "\x00\x00\x48\x41"s; f = 12.5f; }
    DOCTEST_SUBCASE("0x3dcccccd (= 0.1)") { buffer = "____"s; f = 0.1f; }
    DOCTEST_SUBCASE("0x3a83126f (= 0.001)") { buffer = "\x6f\x12\x83\x3a"s; f = 0.001f; }
    DOCTEST_SUBCASE("0x43160000 (= 150.0)") { buffer = "\x00\x00\x16\x43"s; f = 150.0f; }

    auto it = buffer.cbegin();
    float result = cla::utils::parseLittleEndianBuffer<float>(it);
    REQUIRE(it == buffer.cend());
    REQUIRE(result == f);
}
```

As I'm migrating from Catch where it [required a patch](https://github.com/catchorg/Catch2/issues/734) which tweaked the test nesting that is printed out upon a failure. And indeed, it seems that a similar thing is needed with doctest. With its current `dev` branch, the test failure report only includes `"converting a byte buffer to a float"`, but it does not show that the offending data came from the `0x3dcccccd (= 0.1)` subcase:

```
[doctest] doctest version is "2.3.1"
[doctest] run with "--help" for options
===============================================================================
tests/numbers.cpp:213:
TEST CASE:  converting a byte buffer to a float

tests/numbers.cpp:225: FATAL ERROR: REQUIRE( it == buffer.cend() ) is NOT correct!
  values: REQUIRE( {?} == {?} )

===============================================================================
[doctest] test cases:      4 |      3 passed |      1 failed |      0 skipped
[doctest] assertions:    109 |    108 passed |      1 failed |
[doctest] Status: FAILURE!
```

[...]

PR #364, opened 2020-04-30
Print subcase stack upon a subcase entry

jktjkt:
This is largely a hack to give us at least *something* for data driven testing (#215). In that pattern, it's common to structure code in a way that first "generates the data", and only then executes the actual test. It can look like this:

```c++
  TEST_CASE("parsing stuff") {
    std::string input;
    Something expected;

    DOCTEST_SUBCASE("single-line") {
      DOCTEST_SUBCASE("empty") {}
      DOCTEST_SUBCASE("trivial") {
        input = "create foo";
	expected = {OP_CREATE, {"foo",}};
      }
      DOCTEST_SUBCASE("two args") {
        input = "create blah bar";
	expected = {OP_CREATE, {"blah", "bar",}};
      }
    }
    DOCTEST_SUBCASE("multi-line") {
      DOCTEST_SUBCASE("trailing whitespace") {
        input = "create foo\n\n";
	expected = {OP_CREATE, {"foo,}};
      }
      DOCTEST_SUBCASE("one word per line, two words") {
        input = "create\nfoo";
	expected = {OP_CREATE, {"foo",}};
      }
    }

    REQUIRE(parse(input) == expected);
  }
```

The important part is that once we're executing actual tests, the subsections will have been exited already, so the usual ways of showing the subcase stack can be no longer applied.

In Catch2, there was a patch which tried to track the deepest subcase stack so far. Let's try to implement something very simple, just printing the current state of the stack once it is entered. It does not tie into any error handling, so the output will be always there, and it also probably does not support any fancy filtering when skipping subcases. However, it's something which at least shows what is going on when an error gets reported.

Cc: https://github.com/onqtam/doctest/issues/215
Cc: https://github.com/onqtam/doctest/issues/125
Cc: @syyyr

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -5268,6 +5268,12 @@ namespace {
             std::lock_guard<std::mutex> lock(mutex);
             subcasesStack.push_back(subc);
             hasLoggedCurrentTestStart = false;
+
+            s << Color::Cyan << "[doctest] SUBCASE" << Color::None << "\n";
+            for(auto& curr : subcasesStack)
+                if(curr.m_name[0] != '\0')
+                    s << " - " << Color::LightGrey << curr.m_name << Color::None << " (" << curr.m_file << ":" << curr.m_line << ")\n";
+            s << "\n";
         }
 
         void subcase_end() override {
--- a/doctest/parts/doctest.cpp
+++ b/doctest/parts/doctest.cpp
@@ -2656,6 +2656,12 @@ namespace {
             std::lock_guard<std::mutex> lock(mutex);
             subcasesStack.push_back(subc);
             hasLoggedCurrentTestStart = false;
+
+            s << Color::Cyan << "[doctest] SUBCASE" << Color::None << "\n";
+            for(auto& curr : subcasesStack)
+                if(curr.m_name[0] != '\0')
+                    s << " - " << Color::LightGrey << curr.m_name << Color::None << " (" << curr.m_file << ":" << curr.m_line << ")\n";
+            s << "\n";
         }
 
         void subcase_end() override {


onqtam:
Thanks for the PR. Obviously I'll implement this in a different way, but this has been a sign that this feature is indeed necessary - I'll try to do it today (and if not today - in the upcoming week).

Currently doctest prints only the current subcase callstack when an assert fails but I'm thinking of keeping a list of the deepest subcase stack reached so far, and printing that only when necessary (and not always by default such as in the PR)

I'll close the PR now and will close the 2 other issues when I'm done with the feature.

onqtam:
done - for a program like this:
```c++
TEST_CASE("data-driven") {
    int a;
    SUBCASE("subcase 1")
        a = 5;
    SUBCASE("subcase 2")
        a = 42;
    CHECK(a == 6);
}
```
the output will be something like this:
```
E:\doctest\scripts\playground\test.cpp:61:
TEST CASE:  data-driven

DEEPEST SUBCASE STACK REACHED (DIFFERENT FROM THE CURRENT ONE):
  subcase 1

E:\doctest\scripts\playground\test.cpp:67: ERROR: CHECK( a == 6 ) is NOT correct
!
  values: CHECK( 5 == 6 )

E:\doctest\scripts\playground\test.cpp:61:
TEST CASE:  data-driven

DEEPEST SUBCASE STACK REACHED (DIFFERENT FROM THE CURRENT ONE):
  subcase 2

E:\doctest\scripts\playground\test.cpp:67: ERROR: CHECK( a == 6 ) is NOT correct
!
  values: CHECK( 42 == 6 )
```

PR #367, opened 2020-05-07
fix: avoid UB for increment of nullptr pointer

oktonion:

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -3554,6 +3554,10 @@ namespace detail {
         DOCTEST_ITERATE_THROUGH_REPORTERS(subcase_start, m_signature);
     }
 
+    DOCTEST_MSVC_SUPPRESS_WARNING_WITH_PUSH(4996) // std::uncaught_exception is deprecated in C++17	
+    DOCTEST_GCC_SUPPRESS_WARNING_WITH_PUSH("-Wdeprecated-declarations")	
+    DOCTEST_CLANG_SUPPRESS_WARNING_WITH_PUSH("-Wdeprecated-declarations")
+
     Subcase::~Subcase() {
         if(m_entered) {
             // only mark the subcase stack as passed if no subcases have been skipped
@@ -3577,6 +3581,10 @@ namespace detail {
             DOCTEST_ITERATE_THROUGH_REPORTERS(subcase_end, DOCTEST_EMPTY);
         }
     }
+
+    DOCTEST_CLANG_SUPPRESS_WARNING_POP	
+    DOCTEST_GCC_SUPPRESS_WARNING_POP	
+    DOCTEST_MSVC_SUPPRESS_WARNING_POP
 
     Subcase::operator bool() const { return m_entered; }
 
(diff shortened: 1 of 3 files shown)


oktonion:
Ooops, sorry @onqtam. I've captured your commits from master branch. Hope this is ok.

onqtam:
I guess the only changes you wanted to make are in `wildcmp`:
```c++
const char* cp = str;
const char* mp = wild;
```
How was the UB identified?

oktonion:
> 
> 
> I guess the only changes you wanted to make are in `wildcmp`:
> 
> ```c++
> const char* cp = str;
> const char* mp = wild;
> ```
> 
> How was the UB identified?

Yep, exactly. Just merged recent master branch to the develop with this PR. Function `wildcmp` is poorly written, for UB check the final **else** block:
```cpp
...
        const char* cp = nullptr;
        const char* mp = nullptr;
...
        while(*str) {
            if(*wild == '*') {
...
            } else {
                wild = mp;   //!OCLINT parameter reassignment
                str  = cp++; //!OCLINT parameter reassignment
            }
        }
...
```

1. Value of `cp` is incremented however it may and _would_ be **nullptr** - this is already the problem, what is **nullptr**++? 
2. Then `wild` pointer gets value of **nullptr** and `ptr` pointer gets value of **nullptr**++ leading to dereference of **nullptr** at `while(*str)` and `if(*wild == '*')` - things are getting nasty.

onqtam:
But did UBSAN detect this? or perhaps some static analysis tool? Did you get a crash?
I've taken the code from here: https://www.codeproject.com/Articles/1088/Wildcard-string-compare-globbing
and I adapted it to be optionally case-insensitive.

If this was reported by a static analysis tool - can you prove that it is indeed possible (and not just a false positive) to get to the case where we are incrementing a null pointer and later even dereferencing `wild` which could have been assigned a `nullptr` because of `mp`?

I'm hesitant to accept these changes without thorough testing of the function (which is currently missing in doctest's test suite...).

oktonion:
On second thought the code is... okayish.

I've missed this part (that goes before problematic code with hidden UB):
```cpp
...
while ((*str) && (*wild != '*')) {
...
```
The thing is that `*wild` always will be '*' by the point of second **while** loop so first **if** of the loop `if(*wild == '*')` will always be **true** and 
```cpp
...
		if (*wild == '*') {
			if (!*++wild) {
				return 1;
			}
			mp = wild;
			cp = str + 1;
		}
...
```
block will always init `mp` and `cp` pointers.

But I highly doubt that this is a good thing and the code is alright. It feels very unsafe when one change to `while ((*str) && (*wild != '*')) {` could pop UB out of nowhere. Just the fact that it takes an hour and two developers to sort this out is indicative.

My alternative with
```cpp
const char* cp = str;
const char* mp = wild;
```
is much safer but it is up to you as maintainer to accept the changes or take the risks of future refactoring.

onqtam:
Actually I just made the effort to understand the code and it seems that your proposed change won't hurt indeed - and if it silences some static analysis tool I'll accept it. On that note... did you use any specific tool?

And can we remove the other 2 commits and leave just this change? Or I can just commit it myself - no problem.

oktonion:
> 
> 
> Actually I just made the effort to understand the code and it seems that your proposed change won't hurt indeed - and if it silences some static analysis tool I'll accept it. On that note... did you use any specific tool?
> 
> And can we remove the other 2 commits and leave just this change? Or I can just commit it myself - no problem.

Thanks for the review. I've created PR with just this change. This PR could be closed.

PR #368, opened 2020-05-08
fix: possible UB with nullptr increment

oktonion:
see #367


onqtam:
Thanks, and sorry for all the hassle :)

PR #377, opened 2020-05-22
CMakeLists uses CXX_STANDARD witch is only available from 3.1

flagarde:
The actual CMakeLists uses CXX_STANDARD witch is only available from 3.1.
Using the cmake_minimum_required(VERSION [...] [FATAL_ERROR]) allows to add the policies automatically. It's running smoothly on 3.17.2 so the max version set to this version would be fine.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,8 +1,4 @@
-cmake_minimum_required(VERSION 3.0)
-
-if(POLICY CMP0077)
-    cmake_policy(SET CMP0077 NEW)
-endif()
+cmake_minimum_required(VERSION 3.1...3.17.2)
 
 ################################################################################
 ## DOCTEST
@@ -11,9 +7,13 @@ endif()
 file(READ ${CMAKE_CURRENT_SOURCE_DIR}/scripts/version.txt ver)
 project(doctest VERSION ${ver} LANGUAGES CXX)
 
+set(CMAKE_CXX_STANDARD "11")
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
+
 option(DOCTEST_WITH_TESTS               "Build tests/examples" ON)
 option(DOCTEST_WITH_MAIN_IN_STATIC_LIB  "Build a static lib (cmake target) with a default main entry point" ON)
-option(DOCTEST_NO_INSTALL  "Skip the installation process" OFF)
+option(DOCTEST_INSTALL  "Perform the installation process" ON)
 
 add_library(${PROJECT_NAME} INTERFACE)
 add_library(${PROJECT_NAME}::${PROJECT_NAME} ALIAS ${PROJECT_NAME})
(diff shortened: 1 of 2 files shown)


onqtam:
Hi!

This PR seems to do more than it's supposed to - changing the option name and inverting it's behavior is a breaking change in terms of integration - it might have been a mistake to make it that way initially (when the option was introduced), but I'm hesitant to change things without good reason...

I'm also not familiar with that syntax for `cmake_minimum_required` - does it actually set a maximum version in addition to a minimum? Won't that bring maintenance burden as time goes by? What about using the latest unofficial CMake or minor version bumps & patches?

Also about the removal of `set(CMAKE_CXX_STANDARD 11)` - it has history - there have been a few commits that shuffle things around regarding how the CXX standard is set and I'd be happy not to touch things without good reason. Why was it removed? Did it actually break a usecase? Can you prove that no other usecase has been broken? The CI seems to pass: https://travis-ci.org/github/onqtam/doctest/builds/690111601 (not sure how/when it got disintegrated with github PRs and no longer shows up in here... anyway), but in the past there were usecases which were not covered by the CI setup of doctest.

CMakeLists.txt
```diff
@@ -1,8 +1,4 @@
-cmake_minimum_required(VERSION 3.0)
-
-if(POLICY CMP0077)
-    cmake_policy(SET CMP0077 NEW)
-endif()
+cmake_minimum_required(VERSION 3.1...3.17.2)
```
Ghabry, line 1: The ``3.1`` change makes sense. Everything starting from ``...`` is not required. I also havn't ever seen a CMake file specifying a "max version".

About the CMP0077 removal: This is wrong.

CMakeLists.txt
```diff
@@ -11,9 +7,13 @@ endif()
 file(READ ${CMAKE_CURRENT_SOURCE_DIR}/scripts/version.txt ver)
 project(doctest VERSION ${ver} LANGUAGES CXX)
 
+set(CMAKE_CXX_STANDARD "11")
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
```
Ghabry, line 12: using ``CMAKE_CXX_STANDARD`` is the correct way to do this in older CMake versions.
These lines should be at the beginning (before any ``add_library/executable``) because they are inherited. So after ``project`` is the best place to put it.

CMakeLists.txt
friendlyanon: No, it's not wrong. See https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#policy-settings  
The minimum version is for erroring out if the CMake version is too old, but the maximum version is for policies. `CMP0077` was introduced in `3.13` so a max of `3.17` will set it to `NEW`.

If you somehow don't trust the CMake documentation, then it's easy to verify these:
```cmake
cmake_minimum_required(VERSION 3.1...3.17)
cmake_policy(GET CMP0077 value)
message(NOTICE "${value}")
```
`cmake -P test.cmake` will print `NEW`.

CMakeLists.txt
friendlyanon: If the minimum version requirement is `3.1`, then these variables can be dropped, because [`CXX_STANDARD`](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html) and the others were introduced in `3.1`.  
Regardless, [`target_compile_features`](https://cmake.org/cmake/help/latest/command/target_compile_features.html) would be the idiomatic way to set a minimum standard requirement on a target, as that automatically propagates and doesn't assume certain directory structures.  
The less traditional CMake-isms there in the root lists file, the easier it will be to trivially pull in doctest as a dependency in modern CMake.

mitchgrout:
Closing as obsolete via #811

PR #388, opened 2020-06-22
Add a note that doctest can be installed through Homebrew

cameronwhite:
A Homebrew package was added recently (https://github.com/Homebrew/homebrew-core/pull/56464).


onqtam:
Thanks! Glad to see it available from more places :)

PR #412, opened 2020-08-10
Port Catch2 break into debugger for Linux. closes #411

### LINKED ISSUE #411: Break into debugger support is missing for Linux
Thanks for this library!

## Description
Currently doctest does not break into the debugger on failed assertions if running on Linux.

### Steps to reproduce
When running tests with a failing assertion inside a debugger on Linux, the tests run to completion and report a failure.

mikezackles:
## Description
This adds support for breaking into the debugger on Linux. The code is ported over from Catch2. (I assume that isn't an issue.)

## GitHub Issues
#411

Note that I did not run clang-format because it picked up a bunch of other things, and I felt it would muddy the diff.

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -355,7 +355,15 @@ DOCTEST_MSVC_SUPPRESS_WARNING(26812) // Prefer 'enum class' over 'enum'
 
 #ifndef DOCTEST_BREAK_INTO_DEBUGGER
 // should probably take a look at https://github.com/scottt/debugbreak
-#ifdef DOCTEST_PLATFORM_MAC
+#ifdef DOCTEST_PLATFORM_LINUX
+#if defined(__GNUC__) && (defined(__i386) || defined(__x86_64))
+// Break at the location of the failing check if possible
+#define DOCTEST_BREAK_INTO_DEBUGGER() __asm__("int $3\n" : :)
+#else
+#include <signal.h>
+#define DOCTEST_BREAK_INTO_DEBUGGER() raise(SIGTRAP)
+#endif
+#elif defined(DOCTEST_PLATFORM_MAC)
 #define DOCTEST_BREAK_INTO_DEBUGGER() __asm__("int $3\n" : :)
 #elif DOCTEST_MSVC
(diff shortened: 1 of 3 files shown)


doctest/doctest.h
```diff
@@ -355,7 +355,15 @@ DOCTEST_MSVC_SUPPRESS_WARNING(26812) // Prefer 'enum class' over 'enum'
 
 #ifndef DOCTEST_BREAK_INTO_DEBUGGER
 // should probably take a look at https://github.com/scottt/debugbreak
-#ifdef DOCTEST_PLATFORM_MAC
+#ifdef DOCTEST_PLATFORM_LINUX
+#if defined(__GNUC__) && (defined(__i386) || defined(__x86_64))
+// Break at the location of the failing check if possible
+#define DOCTEST_BREAK_INTO_DEBUGGER() __asm__("int $3\n" : :)
+#else
+#include <signal.h>
```
onqtam, line 363: I think we could get by with `extern int raise (int __sig) throw();`

onqtam:
Sorry for the delay. This is a welcome addition!

However I had [some problems recently](https://www.reddit.com/r/cpp/comments/hh917p/doctest_the_fastest_featurerich_singleheader/fw9kwdt/) with code being taken as-is from Catch2 and although I do acknowledge at [the top of the doctest header](https://github.com/onqtam/doctest/blob/2.4.0/doctest/doctest.h#L20-L32) that the breaking into the debugger is taken from Catch and even though Catch is BSL-licensed which is MIT-compatible based on my reading, this might need to be changed a bit.

IMHO there shouldn't be a problem taking this directly (how is someone else expected to implement it - differently?...), but I'll fix it up a bit after merging (remove the comments, etc) and point to Catch2 for the explanation.

mikezackles:
> Sorry for the delay. This is a welcome addition!

No problem, thanks for spending time on maintenance!

Thank you for being thorough with the licensing, and sorry if it's causing headaches. (I hope I was clear enough that this came mostly verbatim from Catch2. In retrospect I should have linked the relevant sections explicitly.) Very interesting to see the additional discussion context on licensing.

Anyway, thanks again for your time and the library! I'm excited to have this feature upstream.

PR #416, opened 2020-09-04
Add option to build with std headers.

avostrik:
If doctest forward declaration differs from standard library implementation
in cross-compilation SDK build of doctest for the target fails.

Signed-off-by: Andrey Vostrikov <andrey.vostrikov@cogentembedded.com>


onqtam:
Thanks!

PR #417, opened 2020-09-06
Add JUNIT_OUTPUT_DIR option to doctest_discover_tests

Tradias:
Adds support for generating junit-style xml output when discovering tests with `doctest_discover_tests` and running them through `ctest --parallel`. This feature has been inspired by a similar feature in [CMake's GoogleTest module](https://cmake.org/cmake/help/v3.18/module/GoogleTest.html) called `XML_OUTPUT_DIR`. In short:

```
  ``JUNIT_OUTPUT_DIR dir``
    If specified, the parameter is passed along with ``--reporters=junit``
    and ``--out=`` to the test executable. The actual file name is the same
    as the test target, including prefix and suffix. This should be used
    instead of EXTRA_ARGS to avoid race conditions writing the XML result
    output when using parallel test execution.
```

The implementation is based on [GoogleTest.cmake](https://github.com/Kitware/CMake/blob/v3.18.2/Modules/GoogleTest.cmake)  and [GoogleTestAddTests.cmake](https://github.com/Kitware/CMake/blob/v3.18.2/Modules/GoogleTestAddTests.cmake). It does not take care of creating the specified output directory.

### DIFF
--- a/scripts/cmake/doctest.cmake
+++ b/scripts/cmake/doctest.cmake
@@ -33,6 +33,7 @@ same as the doctest name; see also ``TEST_PREFIX`` and ``TEST_SUFFIX``.
                          [TEST_SUFFIX suffix]
                          [PROPERTIES name1 value1...]
                          [TEST_LIST var]
+                         [JUNIT_OUTPUT_DIR dir]
     )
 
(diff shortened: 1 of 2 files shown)


onqtam:
Thanks!

Issue #413, opened 2020-08-17
Support unit tests with MPI


BerengerBerthoul:
Hello, thanks for this very useful and convenient library!

I work in the field of computational fluid mechanics and particularly on HPC topics. In our field, it means we have distributed memory over multiple processors, communicating with the MPI standard.

Thus we need a way to unit-test mpi-distributed code. For that, we developped (Bruno Maugars and myself) a doctest extension that copes with distributed tests. The extension is working well and we are currently using it for two projects at ONERA, our company.

The extension is pretty small (<500 loc) and non-intrusive to doctest.h (in the sense that it does not change its sources). I forked the doctest repository on [github](https://github.com/BerengerBerthoul/doctest/tree/mpi_tests) (branch `mpi_tests`). The documentation is [here](https://github.com/BerengerBerthoul/doctest/blob/mpi_tests/doc/markdown/mpi_tests.md).

If you are interested, we would like to have this functionnality integrated to doctest. It seems useful to have, and even if MPI uses are marginal, I think that there is still a fair number of people using it. The MPI standard is used by almost everyone doing HPC-related developments, and it will stay that way for still many years. To our knowledge, no unit test framework supports mpi-distributed memory well.

Regarding the packaging, I don't know how it should be done. My guess is that the sources for MPI should be somehow optional. For now I put the sources in `doctest/mpi/` and the examples in `examples/mpi/`. Regarding CMake, I just copy-pasted-edited the all_features one as a proof-of-concept solution (hence `ctest` is not working since it doesn't know it should run the mpi tests with the `mpirun` command). Personnally, I ran the example test by hand with `cd $BUILD_DIR/examples/mpi && mpirun -np 3 ./mpi_features`.

The build process was done with `CMake 3.12` and `gcc/8.3` with C++17 (use of `[[maybe_unused]]` to quickly remove the unused-parameter -Werror).

I am ready to discuss the technical details, but the first question is: are you interested in supporting distributed-memory unit tests ?

onqtam:
Hello, and sorry for the *really* late reply on my side - I'll take a look at what work has accumulated around doctest in the following days/weeks.

[...]

PR #418, opened 2020-09-08
Initial MPI unit tests implementation

BerengerBerthoul:
## Description
Adds an mpi extension for testing mpi-distributed unit tests

## GitHub Issues
Implements the feature described in #413

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -213,6 +213,8 @@ jobs:
           fi
 
           sudo apt-get update
+          # sudo apt-get install libopenmpi-dev
+          # sudo apt-get install openmpi-bin
 
           # libc6-dbg:i386 is required for running valgrind on x86.
           sudo apt-get install -y ninja-build valgrind libc6-dbg:i386
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,7 @@ option(DOCTEST_USE_STD_HEADERS  "Use std headers" OFF)
 add_library(${PROJECT_NAME} INTERFACE)
 add_library(${PROJECT_NAME}::${PROJECT_NAME} ALIAS ${PROJECT_NAME})
 
-if(NOT CMAKE_VERSION VERSION_LESS 3.8) 
+if(NOT CMAKE_VERSION VERSION_LESS 3.8)
     target_compile_features(${PROJECT_NAME} INTERFACE cxx_std_11)
 endif()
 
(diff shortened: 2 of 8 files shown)


BerengerBerthoul:
Please feel free to refactor / criticize the development. Some points to note :

- there is a TODO item in the .md documentation that is here for information but should probably be removed / moved somewhere else
- I changed some lines in the top-level CMakeLists to work with the extension, but it could be cleaned
- There is an example in examples/mpi. I placed it here but it should probably go in another extensions folder (and then the documentation should be updated). I copy/pasted the all_features CMakeLists because I didn't not know how to do this properly, but obviouly  it should be replaced by something else

BerengerBerthoul:
Ah yes : the tests are failing because the new MPI test build requires, logically, the MPI package, which seems to not be present by default on the CI machine. I guess there is three possible correction to this problem:

- Either we don't use CI for extensions ("use at your own perils'), at least for now
- Always include MPI for the doctest CI
- Only try to load MPI for the doctest_mpi CI

onqtam:
I'll try to review this whenever possible

philbucher:
Awesome feature, looking forward to this!

Regarding the Github CI you could install OpenMPI on the fly like I do [here](https://github.com/philbucher/communication-tests/blob/master/.github/workflows/build_and_test.yml#L61-62)

BerengerBerthoul:
The MPI package has been made optionnal for regular tests. The solution of @philbucher allows to have mpi, but then the mpi test fails because the doctest test executable must be launched with "mpirun -np 3", and we don't know how to do it. But I think @onqtam will be able to fix it easily. Anyway, rather than failing the CI because of this missing "mpirun -np 3", the openmpi installation in the ci file are there but commented out

philbucher:
from my experience running (Open)MPI in the CI:
- don't run it as root, OpenMPI doesn't like this and might abort
- newer versions of OpenMPI check the available resources. If using more processes than what is available then you need the `--oversubscribe` option, e.g. like this: `mpiexec --oversubscribe -np 3 python3 script.py`

BerengerBerthoul:
Ok take your time ;) If it can help, we could set up a quick video call to present the why, how and what remains to be done.

onqtam:
Sorry for the late reply. I merged it last week and just pushed a few small fixes - I played with it a bit locally and overall it looks good! You seem to have your own TODO roadmap in the documentation of the extension and I won't be messing with that.

There's 1 other suggestion I have in mind: perhaps we should embed `mpi_reporter.h` into `doctest_mpi.h` since it's not meant to be used from anywhere else anyway...?

Thanks again for the contribution - future responses will be swifter!

BerengerBerthoul:
Great! Thanks for the merge and the `CmakeLists.txt` cleaning. I will try to test the CI  with openmpi on my fork and submit a new PR once it is done (+ minor corrections/update in the doc)
Regarding `mpi_reporter.h`, it is up to you. There is no need to have two files, I just found it cleaner.

PR #425, opened 2020-10-09
Silence remarks on old C-style casts

### LINKED ISSUE #424: Avoid old C-style casts
My IAR C++ compiler (EWARM8.40) issues a remark on constructs such as `(void)0`:
> Remark[Pa137]: use of C-style cast [...] 

Of course, this can be silenced in a doctest-trampoline header, but so far, it is the only remark that needs this.
I'd like to point out this feature of doctest: it is really silent warning/remark-wise! :-)

Pull-request #425 should fix that.

UnePierre:
## Description

Changed all old C-style casts `(void)x` to C++-style `static_cast<void>(x)`.

This silences remarks about using old C-style casts, e.g. from IAR EWARM.

## GitHub Issues

Closes #424.


onqtam:
Thanks for the improvement! merging.

PR #428, opened 2020-10-25
Support Clang 3.4

AlexanderLanin:
While the compiler is rather old it seems trivial enough to support Clang 3.4.
**Although I basically have no idea what the effects of this small change are...**
I personally don't care much about performance on this old compiler.

Note: it's also possible to add clang-3.4 to GitHub Actions. I can add this to the pull request if you like.
It's something like:
```
            if [ "${{ matrix.config.version }}" = "3.4" ]; then
              echo "CC=clang" >> $GITHUB_ENV
              echo "CXX=clang++" >> $GITHUB_ENV
              echo "CXXFLAGS=-stdlib=libc++" >> $GITHUB_ENV

              # clang-3.4 is available from trusty
              sudo add-apt-repository "deb http://dk.archive.ubuntu.com/ubuntu/ trusty main"
              sudo add-apt-repository "deb http://dk.archive.ubuntu.com/ubuntu/ trusty universe"
            fi
            sudo apt install -y clang-${{ matrix.config.version }} g++-multilib
```


onqtam:
Thanks!

PR #429, opened 2020-10-30
Enum support (fix for Issue #121)

### LINKED ISSUE #121: Add default printers for enums
I've added an assertion for a function returning an enum and the enum value. When the test failed I've got pretty informative failure output {?} == {?}...

I don't intend to write a custom printer for my enum type, but still want to see readable values when a test fails. I've workaround the problem by casting to int, but I'd prefer if the library does this for me.

jkriegshauser:
## Description
When a `CHECK` fails and the decomposed expression includes an enum, the values are not printed but instead `{?}` is printed. This resolves that.

## GitHub Issues
Closes #121

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -417,12 +417,6 @@ struct char_traits<char>;
 template <class charT, class traits>
 class basic_ostream;
 typedef basic_ostream<char, char_traits<char>> ostream;
-template <class charT, class traits>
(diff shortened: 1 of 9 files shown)


onqtam:
Good work! I was surprised to see that all compilers in the current build matrix actually have `__underlying_type` and `__is_enum` with the same names.

My only concern for this (as noted in #121 ) is that it adds SFINAE on the main instantiation path for every assert because of the use of `enable_if`... However, this will impact compile times only a bit, and this is useful enough to have enabled by default for all users. Merging!

PR #431, opened 2020-11-05
switch from using operator<< to the comma operator for logging

biojppm:
In the `operator<<(stream, T)` calls from doctest, using `stream.write()` was possible from the CHECK() macros, but not from INFO() and CHECK_MESSAGE(). This PR addresses that.

An example of the use case can be seen in the added test in stringification.cpp. I've updated the test result files.

I do have a doubt about the size passed into `write()`. The correct type would be `std::streamsize` but the standard only specifies that it is a `typedef`, and nothing regarding the particular type, so it is difficult or impossible to provide a safe forward declaration. So I've went with `unsigned`, as `size_t` would require `#include <cstdint>`.

### DIFF
--- a/examples/all_features/stringification.cpp
+++ b/examples/all_features/stringification.cpp
@@ -75,6 +75,33 @@ struct Foo
     friend bool operator==(const Foo&, const Foo&) { return false; }
 };
 
+struct MyStringType
+{
+    std::string s;
+    template <size_t N> MyStringType(const char (&in)[N]) : s(in, in+N) {}
+    friend bool operator==(const MyStringType &lhs, const MyStringType &rhs) { return lhs.s == rhs.s; }
+};
+
+template <typename T, typename K>
+std::ostream& operator<<(std::ostream& stream, const MyType<T, K>& in) {
+    stream << "[" << in.one << ", " << in.two << "]";
+    return stream;
+}
+
+// you can use the stream.write() method to write blocks of characters, and
+// you also can use a template operator<< if your code does not use
+// std::ostream and you'd like to avoid the include
+template <class OStream>
+OStream& operator<<(OStream& stream, const MyStringType &in) {
+    // the terminating \0 character may be present in size()
+    auto size = in.s.size() > 0u ? in.s.size() - 1u : 0u;
+    constexpr const char quote = '\'';
+    stream.write(&quote, 1u);
+    stream.write(in.s.data(), static_cast<unsigned>(size));
+    stream.write(&quote, 1u);
+    return stream;
+}
+
 // as a third option you may provide an overload of toString()
 inline doctest::String toString(const Foo&) { return "Foo{}"; }
 } // namespace Bar
@@ -126,6 +153,18 @@ TEST_CASE("all asserts should fail and show how the objects get stringified") {
     lst_2.push_back(666);
 
     CHECK(lst_1 == lst_2);
+
+    {
+        // use of ostream::write() is possible:
+        Bar::MyStringType s1 = "some";
+        Bar::MyStringType s2 = "contents";
+
+        // also inside the message builder
+        INFO("s1=" << s1 << " s2=" << s2);
+
+        CHECK(s1 == s2);
+        CHECK_MESSAGE(s1 == s2, s1 << " is not really " << s2);
+    }
 
     // lets see if this exception gets translated
     throw_if(true, bla1);
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -1547,6 +1547,8 @@ namespace detail {
             return *this;
         }
 
+        MessageBuilder& write(const char* data, unsigned size);
+
         bool log();
         void react();
     };
@@ -4268,6 +4270,11 @@ namespace detail {
 
     IExceptionTranslator::IExceptionTranslator()  = default;
     IExceptionTranslator::~IExceptionTranslator() = default;
+
+    MessageBuilder& MessageBuilder::write(const char* data, unsigned size) {
+        m_stream->write(data, std::streamsize(size));
+        return *this;
+    }
 
     bool MessageBuilder::log() {
         m_string = getTlsOssResult();
--- a/doctest/parts/doctest.cpp
+++ b/doctest/parts/doctest.cpp
@@ -1627,6 +1627,11 @@ namespace detail {
     IExceptionTranslator::IExceptionTranslator()  = default;
     IExceptionTranslator::~IExceptionTranslator() = default;
 
+    MessageBuilder& MessageBuilder::write(const char* data, unsigned size) {
+        m_stream->write(data, std::streamsize(size));
+        return *this;
+    }
+
     bool MessageBuilder::log() {
         m_string = getTlsOssResult();
         DOCTEST_ITERATE_THROUGH_REPORTERS(log_message, *this);
--- a/doctest/parts/doctest_fwd.h
+++ b/doctest/parts/doctest_fwd.h
@@ -1544,6 +1544,8 @@ namespace detail {
             return *this;
         }
 
+        MessageBuilder& write(const char* data, unsigned size);
+
         bool log();
         void react();
     };
--- a/examples/all_features/test_output/stringification.cpp.txt
+++ b/examples/all_features/test_output/stringification.cpp.txt
@@ -15,6 +15,15 @@ stringification.cpp(0): ERROR: CHECK( vec1 == vec2 ) is NOT correct!
 stringification.cpp(0): ERROR: CHECK( lst_1 == lst_2 ) is NOT correct!
   values: CHECK( [1, 42, 3, ] == [1, 2, 666, ] )
 
+stringification.cpp(0): ERROR: CHECK( s1 == s2 ) is NOT correct!
+  values: CHECK( 'some' == 'contents' )
+  logged: s1='some' s2='contents'
+
+stringification.cpp(0): ERROR: CHECK( s1 == s2 ) is NOT correct!
+  values: CHECK( 'some' == 'contents' )
+  logged: s1='some' s2='contents'
+          'some' is not really 'contents'
+
 stringification.cpp(0): ERROR: test case THREW exception: MyTypeInherited<int>(5, 4)
 
 ===============================================================================
@@ -25,6 +34,6 @@ stringification.cpp(0): ERROR: test case THREW exception: 5
 
 ===============================================================================
 [doctest] test cases: 2 | 0 passed | 2 failed |
-[doctest] assertions: 4 | 0 passed | 4 failed |
+[doctest] assertions: 6 | 0 passed | 6 failed |
 [doctest] Status: FAILURE!
 Program code.
--- a/examples/all_features/test_output/stringification.cpp_junit.txt
+++ b/examples/all_features/test_output/stringification.cpp_junit.txt
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <testsuites>
-  <testsuite name="all_features" errors="2" failures="4" tests="4">
+  <testsuite name="all_features" errors="2" failures="6" tests="6">
     <testcase classname="stringification.cpp" name="all asserts should fail and show how the objects get stringified" status="run">
       <failure message="Foo{} == Foo{}" type="CHECK">
 stringification.cpp(0):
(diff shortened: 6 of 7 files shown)


biojppm:
I just realized that clang dislikes this, while VS and GCC compile and run correctly.

The issue with clang is an ambiguity:
```
/mnt/d/proj/c4core/test/test_charconv.cpp:1015:17: error: use of overloaded operator '<<' is ambiguous (with operand types 'doctest::detail::MessageBuilder' and 'c4::csubstr' (aka 'basic_substring<const char>'))
    INFO("val=" << val);
    ~~~~~~~~~~~~^~~~~~~
/mnt/d/proj/extern/doctest/doctest/doctest.h:2486:14: note: expanded from macro 'INFO'
#define INFO DOCTEST_INFO
             ^
/mnt/d/proj/extern/doctest/doctest/doctest.h:1969:61: note: expanded from macro 'DOCTEST_INFO'
                      DOCTEST_ANONYMOUS(_DOCTEST_CAPTURE_), expression)
                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~
/mnt/d/proj/extern/doctest/doctest/doctest.h:1976:20: note: expanded from macro 'DOCTEST_INFO_IMPL'
        mb_name << expression;                                                                     \
        ~~~~~~~~~~~^~~~~~~~~~
/mnt/d/proj/c4core/src/c4/substr.hpp:1588:17: note: candidate function [with OStream = doctest::detail::MessageBuilder, C = const char]
inline OStream& operator<< (OStream& os, basic_substring<C> s)
                ^
/mnt/d/proj/extern/doctest/doctest/doctest.h:1545:25: note: candidate function [with T = c4::basic_substring<const char>]
        MessageBuilder& operator<<(const T& in) {
                        ^
1 error generated.
```

examples/all_features/stringification.cpp
```diff
@@ -75,6 +75,33 @@ struct Foo
...
+template <class OStream>
+OStream& operator<<(OStream& stream, const MyStringType &in) {
+    // the terminating \0 character may be present in size()
+    auto size = in.s.size() > 0u ? in.s.size() - 1u : 0u;
+    constexpr const char quote = '\'';
+    stream.write(&quote, 1u);
+    stream.write(in.s.data(), static_cast<unsigned>(size));
+    stream.write(&quote, 1u);
+    return stream;
+}
```
biojppm, line 103: **Here**. clang fails to compile when this function is a template. OTOH, if this function is not a template, then it compiles and does the right thing. VS2019 and gcc have no problem.

Also, to make it clear, stringification from the standard `CHECK_EQ()` call works as intended on every compiler - template or not.

This would seem a subtle SFINAE difference in behavior between `CHECK_EQ()` and both `INFO()`/`CHECK_MESSAGE()`.

biojppm:
So let's recap. 
 
My type has a freestanding `template<class OStream> operator<<(OStream& stream, T)` function defined. This is sketched in the current `stringification.cpp` changeset of this MR.
 
doctest correctly stringifies this type in any calls to `CHECK_EQ()`. But calling any of `INFO("myvar" << myvar)` or `CHECK_MESSAGE(..., "myvar=" <<
myvar)` has different behavior across all compilers:

|           	| `CHECK_EQ(myvar1 == myvar2)` 	|                     `INFO("myvar=" << myvar)` or `CHECK_MESSAGE(..., "myvar=" << myvar)`                    	|
|----------:	|:----------------------------:	|:-----------------------------------------------------------------------------------------------------------:	|
|    VS2019 	|              ok              	|                                                      ok                                                     	|
|     g++10 	|              ok              	|                         fails to compile, picks `operator<<(MessageBuilder&, myvar)`                        	|
| clang++10 	|              ok              	| fails to compile, ambiguous between `MessageBuilder::operator<<()` and `operator<<(MessageBuilder&, myvar)` 	|

More details:

### Visual Studio

Visual Studio compiles successfully, and as desired, picks my template function and passes it the `std::ostream`.

### GCC

GCC fails with this error:
``` shellsession
/mnt/d/proj/extern/doctest/examples/all_features/stringification.cpp: In instantiation of ‘OStream& Bar::operator<<(OStream&, const Bar::MyStringType&) [with OStream = doctest::detail::MessageBuilder]’:
/mnt/d/proj/extern/doctest/examples/all_features/stringification.cpp:165:9:   required from here
/mnt/d/proj/extern/doctest/examples/all_features/stringification.cpp:101:12: error: ‘struct doctest::detail::MessageBuilder’ has no member named ‘write’
  101 |     stream.write(&quote, 1u);
      |     ~~~~~~~^~~~~
/mnt/d/proj/extern/doctest/examples/all_features/stringification.cpp:102:12: error: ‘struct doctest::detail::MessageBuilder’ has no member named ‘write’
  102 |     stream.write(in.s.data(), static_cast<unsigned>(size));
      |     ~~~~~~~^~~~~
/mnt/d/proj/extern/doctest/examples/all_features/stringification.cpp:103:12: error: ‘struct doctest::detail::MessageBuilder’ has no member named ‘write’
  103 |     stream.write(&quote, 1u);
      |     ~~~~~~~^~~~~
[
```
Which means that GCC does pick my template function, but passes it the `MessageBuilder` class instead of the `std::ostream` class. `MessageBuilder` does not have a `write()` method, so that's why my initial thought was just adding a `MessageBuilder::write()` method. Doing that did fix the GCC behavior. Problem solved, right? No.

### clang

`write()` or no `write()`, clang fails with this error:
``` shellsession
/mnt/d/proj/extern/doctest/examples/all_features/stringification.cpp:165:20: error: use of overloaded operator '<<' is ambiguous (with operand types 'doctest::detail::MessageBuilder' and 'Bar::MyStringType')
        INFO("s1=" << s1 << " s2=" << s2);
        ~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
/mnt/d/proj/extern/doctest/doctest/parts/doctest_fwd.h:2483:14: note: expanded from macro 'INFO'
#define INFO DOCTEST_INFO
             ^
/mnt/d/proj/extern/doctest/doctest/parts/doctest_fwd.h:1966:61: note: expanded from macro 'DOCTEST_INFO'
                      DOCTEST_ANONYMOUS(_DOCTEST_CAPTURE_), expression)
                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~
/mnt/d/proj/extern/doctest/doctest/parts/doctest_fwd.h:1973:20: note: expanded from macro 'DOCTEST_INFO_IMPL'
        mb_name << expression;                                                                     \
        ~~~~~~~~~~~^~~~~~~~~~
/mnt/d/proj/extern/doctest/doctest/parts/doctest_fwd.h:1542:25: note: candidate function [with T = Bar::MyStringType]
        MessageBuilder& operator<<(const T& in) {
                        ^
/mnt/d/proj/extern/doctest/examples/all_features/stringification.cpp:96:35: note: candidate function [with OStream = doctest::detail::MessageBuilder]
template <class OStream> OStream& operator<<(OStream& stream, const MyStringType &in) {
                                  ^
```
So it can't decide between `MessageBuilder::operator<<` and my template function, which would be passed the `MessageBuilder` (similar to GCC).

### Conclusions?

***The what***. For gcc and clang, the stringification behavior for `INFO()`
and `CHECK_MESSAGE()` is different from `CHECK()` when the user `operator<<`
is a template parameterized on the stream. I think we can agree that this is
not desired.

***The how***. Well, adding `write()` to `MessageBuilder` would fix GCC, but
there's still clang. So `write()` is not enough, and probably not necessary.
And the problem seems to be somewhere in the SFINAE leading to
`operator<<`. Can the behavior for `INFO()` and `CHECK_MESSAGE()` be made
equal to `CHECK()`?

I await your input.

biojppm:
@onqtam just to clarify that I am waiting for your input. Not to press here, but I was wondering whether the WIP tag is preventing you from looking at this PR; what initially seemed an easy PR is apparently harder, and at this point I hesitate to proceed further without agreeing on a direction.

onqtam:
@biojppm Sorry for not answering earlier - I have this bookmarked and have been putting it off for 2+ weeks now - will try to get to it when possible.

biojppm:
No problem and no rush, that's just how it goes. I just wanted to make sure you read this.

examples/all_features/stringification.cpp
```diff
@@ -75,6 +75,33 @@ struct Foo
...
+    std::string s;
+    template <size_t N> MyStringType(const char (&in)[N]) : s(in, in+N) {}
+    friend bool operator==(const MyStringType &lhs, const MyStringType &rhs) { return lhs.s == rhs.s; }
+};
+
+template <typename T, typename K>
+std::ostream& operator<<(std::ostream& stream, const MyType<T, K>& in) {
+    stream << "[" << in.one << ", " << in.two << "]";
+    return stream;
+}
```
onqtam, line 89: this doesn't seem related to `MyStringType`

onqtam:
I'm a bit confused... Here is what I tried and it works as expected:

```c++
#include <ostream>

namespace Bar
{
struct MyStringType
{
    std::string s;
    template <size_t N> MyStringType(const char (&in)[N]) : s(in, in+N) {}
    friend bool operator==(const MyStringType &lhs, const MyStringType &rhs) { return lhs.s == rhs.s; }
};

static std::ostream& operator<<(std::ostream& stream, const MyStringType& in) {
    // the terminating \0 character may be present in size()
    auto size = in.s.size() > 0u ? in.s.size() - 1u : 0u;
    constexpr const char quote = '\'';
    stream.write(&quote, 1u);
    stream.write(in.s.data(), static_cast<unsigned>(size));
    stream.write(&quote, 1u);
    return stream;
}
} // namespace Bar

TEST_CASE("") {
    // use of ostream::write() is possible:
    Bar::MyStringType s1 = "some";
    Bar::MyStringType s2 = "contents";

    // also inside the message builder
    INFO("s1=" << s1 << " s2=" << s2);

    CHECK(s1 == s2);
    CHECK_MESSAGE(s1 == s2, s1 << " is not really " << s2);
}
```

And I get this output:

```
/home/onqtam/dt2/scripts/playground/test.cpp:39: ERROR: CHECK( s1 == s2 ) is NOT correct!
  values: CHECK( 'some' == 'contents' )
  logged: s1='some' s2='contents'

/home/onqtam/dt2/scripts/playground/test.cpp:40: ERROR: CHECK( s1 == s2 ) is NOT correct!
  values: CHECK( 'some' == 'contents' )
  logged: s1='some' s2='contents'
          'some' is not really 'contents'
```

Also, templated `operator<<` operators already work as in the case for `MyType<T, K>` - problems begin when the stream itself becomes part of the template parameters but I'm not sure why you would want to do that - is it to avoid including the `<ostream>` header for the `write()` calls...?

onqtam:
In hindsight... using `operator<<` for the `MessageBuilder` class was a mistake - I should have gone with the comma operator and use varargs in the `INFO` macro instead of the awkward use of `<<` within the `()` braces...

examples/all_features/stringification.cpp
biojppm: indeed, it was moved inadvertently and has nothing to do with `MyStringType`.

onqtam:
The problem is not related to SFINAE - actually, clang is the only compliant compiler according to the standard - here's the same issue with a templated comma operator:
https://stackoverflow.com/questions/25110429/ambiguous-operator-overload-on-clang

In there, the author has an **`Update:`** where he suggests that using an `any` struct with a templated constructor forces the templated member operator to always be preferred but that's not true - with that trick the other operator will be preferred, and only then would the addition of `write` to the `MessageBuilder` class make sense, but I don't really like that - I'd like to use the `operator<<` of the `MessageBuilder` class because it calls `doctest::detail::toStream` and that's the only way to have stringification behave the same way as the operands in a `CHECK` macro would be stringified.

Seems that in our case the only way to force the `operator<<` from the `MessageBuilder` class to be always picked up is to explicitly call it via `.operator<<` but that would mean that we can no longer use the pretty syntax with just chaining a bunch of `<<` but have to rewrite everything in terms of function calls which is obviously a no-go.

I managed to get all logging macros to work with the comma operator - the left shift streaming operator is still supported but since it won't compile with your custom stream-templated `operator<<` you can simply use the new syntax - you can see this commit for details: https://github.com/onqtam/doctest/commit/063b858e56356c30f0e79c15ffa631551f713216 - the relevant parts are in `doctest_fwd.h` in the `MessageBuilder` class and in the `DOCTEST_INFO_IMPL` macro.

I'm thinking of renaming this PR to something along the lines of =`switch from using operator<< to the comma operator for logging` so that it ends up in the changelog that way, and to close this PR - is this solution OK for you?

biojppm:
Hey @onqtam, I've typed a reply previously, but just noticed that it is not here. Maybe I simply forgot to click the send button. Anyway, it was just to let you know that I'm entirely on board with your solution, and in fact even like it better.

Thanks for the fix.

PR #440, opened 2020-11-27
Fix 32bit support on macOS

AlexanderLanin:
## Description
macOS 32bit support is broken since https://github.com/onqtam/doctest/commit/3c5e0fa8d2cc7e6409de5e4e6df8e52fb9627a26.

## GitHub Issues
Reported and fixed in https://github.com/ccache/ccache/issues/731


onqtam:
Thanks for this patch!

PR #445, opened 2020-12-09
Add MAIN_PROJECT check for test option

globberwops:
## Description

When adding `doctest` to a **CMake** project using **FetchContent**, testing has to be explicitly disabled.
This check only enables testing, if `doctest` is the main project being built.

This change enables the cleaner usage of `doctest` with **FetchContent**.

#### Before
```cmake
FetchContent_Declare(
  onqtam_doctest
  GIT_REPOSITORY https://github.com/onqtam/doctest.git
  GIT_TAG dev)
set(DOCTEST_WITH_TESTS OFF)
FetchContent_MakeAvailable(onqtam_doctest)
```

#### After
```cmake
FetchContent_Declare(
  onqtam_doctest
  GIT_REPOSITORY https://github.com/onqtam/doctest.git
  GIT_TAG dev)
FetchContent_MakeAvailable(onqtam_doctest)
```

## GitHub Issues

none

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,12 @@ endif()
 file(READ ${CMAKE_CURRENT_SOURCE_DIR}/scripts/version.txt ver)
 project(doctest VERSION ${ver} LANGUAGES CXX)
 
-option(DOCTEST_WITH_TESTS               "Build tests/examples" ON)
+set(MAIN_PROJECT OFF)
+if(CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
+  set(MAIN_PROJECT ON)
+endif()
+
+option(DOCTEST_WITH_TESTS               "Build tests/examples" ${MAIN_PROJECT})
 option(DOCTEST_WITH_MAIN_IN_STATIC_LIB  "Build a static lib (cmake target) with a default main entry point" ON)
 option(DOCTEST_NO_INSTALL  "Skip the installation process" OFF)
 option(DOCTEST_USE_STD_HEADERS  "Use std headers" OFF)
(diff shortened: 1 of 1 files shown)


YarikTH:
For reference, here is the same purpose code from [{fmt}](https://github.com/fmtlib/fmt/blob/master/CMakeLists.txt)
```cmake
# Determine if fmt is built as a subproject (using add_subdirectory)
# or if it is the master project.
set(MASTER_PROJECT OFF)
if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
  set(MASTER_PROJECT ON)
  message(STATUS "CMake version: ${CMAKE_VERSION}")
endif ()
...
...
...
# Options that control generation of various targets.
option(FMT_DOC "Generate the doc target." ${MASTER_PROJECT})
option(FMT_INSTALL "Generate the install target." ${MASTER_PROJECT})
option(FMT_TEST "Generate the test target." ${MASTER_PROJECT})
option(FMT_FUZZ "Generate the fuzz target." OFF)
option(FMT_CUDA_TEST "Generate the cuda-test target." OFF)
option(FMT_OS "Include core requiring OS (Windows/Posix) " ON)
```

So it might be a good idea at least copy explanatory comment and maybe use `CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR` check instead of `CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR`.

globberwops:
@YarikTH I am having a déjà vu ;) Done 👍🏻

onqtam:
Thanks for this! merging :)

PR #461, opened 2021-01-25
Use function-like macros for prefixless macro names

tbleher:
The short macros (without "DOCTEST_" prefix) are very useful, but can
collide with existing names.

Example:

```
  /* LogLevel.h */
  enum class LogLevel {
      FATAL,
      WARN,
      NOTICE,
      INFO,
      DEBUG
  };
```

This code may break with doctest without
_DOCTEST_CONFIG_NO_SHORT_MACRO_NAMES_, since doctest defines macros named
"WARN" and "INFO". These macros will then replace the enum enties WARN by
DOCTEST_WARN (and INFO by DOCTEST_INFO). This will break compilation in
the following case:

```
  #include "LogLevel.h"
  #include "doctest.h"
  static const LogLevel sDefaultLogLevel = LogLevel::INFO;
```

Fix this by using function-like macros for prefixless macros, so the
preprocessor will only replace the names if followed by braces.


tbleher:
Sorry, pull request was against wrong branch.

PR #462, opened 2021-01-25
Use function-like macros for prefixless macro names

tbleher:
The short macros (without "DOCTEST_" prefix) are very useful, but can
collide with existing names.

Example:

```
  /* LogLevel.h */
  enum class LogLevel {
      FATAL,
      WARN,
      NOTICE,
      INFO,
      DEBUG
  };

```
This code may break with doctest without
_DOCTEST_CONFIG_NO_SHORT_MACRO_NAMES_, since doctest defines macros named
"WARN" and "INFO". These macros will then replace the enum enties WARN by
DOCTEST_WARN (and INFO by DOCTEST_INFO). This will break compilation in
the following case:

```
  #include "LogLevel.h"
  #include "doctest.h"
  static const LogLevel sDefaultLogLevel = LogLevel::INFO;
```

Fix this by using function-like macros for prefixless macros, so the
preprocessor will only replace the names if followed by braces.


tbleher:
@onqtam Two checks fail, but it looks like this is an external error:
E.g. one of the failing tests says:
> URL http://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-expat-2.2.9-1-any.pkg.tar.xz is not valid

Anything I can do to fix this?

onqtam:
@tbleher This PR seems like a good idea, thanks!

Those CI failures can be ignored - I haven't figured out how to fix them.

tbleher:
@onqtam Wow, that was quick :) Thanks for merging this so fast!

Issue #370, opened 2020-05-10
Build error with MinGW (Mingw-w64) due to missing Windows.h (with capital W)


jrosdahl:
## Description

After PR #336 it's no longer possible to use doctest for Win32 cross-compilation using e.g. [Mingw-w64](http://mingw-w64.org) on a case sensitive file system.

### Steps to reproduce

On a Linux system (I'm currently using Ubuntu 19.10):

1. `git clone https://github.com/onqtam/doctest.git`
2. `cd doctest`
3. `printf '#define DOCTEST_CONFIG_IMPLEMENT\n#include "doctest/doctest.h"\n' >test.cpp`
4. `i686-w64-mingw32-g++-posix -c test.cpp`
5. Watch the following output:
```In file included from test.cpp:2:
doctest/doctest.h:2766:10: fatal error: Windows.h: No such file or directory
 2766 | #include <Windows.h>
      |          ^~~~~~~~~~~
compilation terminated.
```

Using the lowercase version of `windows.h` works:

1. `sed -i 's/Windows\.h/windows.h/' doctest/doctest.h`
2. `i686-w64-mingw32-g++-posix -c test.cpp && echo success` (prints `success`)

### Extra information

* doctest version: **v2.3.7** or latest master (932a2ca50666 at the time of writing)
* Operating System: **Ubuntu 19.10**
* Compiler+version: **i686-w64-mingw32-g++-posix (GCC) 9.2-posix 20191008**

Maybe of interest:

* ["Lowercase windows.h and uppercase Windows.h difference?" answer on Stack Overflow](https://stackoverflow.com/a/15466951).
* [Catch2 uses the lowercase version](https://github.com/catchorg/Catch2/blob/f0e596e2522f64451f6444edab5c59b8b9e8e082/single_include/catch2/catch.hpp#L7973).

onqtam:
Oh gosh...... I'll fix this in the coming days - I hope I don't ever forget that and switch it back because of "compiler warnings"...

Thanks for reporting!

onqtam:
pushed in the dev branch

u3shit:
Okay, I think I've also got hit by this change. The only problem is that I use clang so I don't have those fancy mingw defines. Why can't we just use the lowercase windows.h always? Even microsoft itself isn't consistent about the casing of this file.

onqtam:
@u3shit Gosh... such exciting computer science problems ))
[...]

PR #463, opened 2021-01-28
Fix compilation on case-sensitive filesystems

jhasse:
## Description

When compiling with a case-sensitive filesystem (e.g. cross-compiling from Linux) Windows.h can't be found in this example.

## GitHub Issues

See #370.


onqtam:
Thanks! just merged it. Will also add `DOCTEST_CLANG_SUPPRESS_WARNING("-Wnonportable-system-include-path")` so that it compiles with clang as well.

Issue #302, opened 2019-10-17
TEST_SUITE integration with CTest


Mirko-von-Leipzig:
## Description
I'm grouping my tests using TEST_SUITEs and using **doctest_discover_tests** to register them with ctest.

I want to run only some group of tests using ctest, which can be done based on test name or labels.

Is there a way to automatically prepend the test suite to the test name when registering with ctest? Or adding it as a label?

I can obviously ignore the TEST_SUITE and manually prefix my TEST_CASEs with the suite name, but that seems easy to screw up.

I can also manually register each test suite using **doctest_discover_tests** with the PREFIX option (I think), but it would be nice to have some automatic once-off way. An additional flag to add suite names as labels/prefix maybe?

Is there an easy way to achieve this that I am missing?

onqtam:
From a brief glance I see that ```doctest_discover_tests``` uses ```--list-test-cases``` - doctest also has ```--list-test-suites```. Or perhaps there needs to be a new option to extract the names of the test suites while listing the test case names... To be honest I have no clue how ```doctest_discover_tests``` works and I won't have the time to look into this anytime soon...

I guess it won't help but just as some extra information - you could set the test suite on a per-test-case-basis with a [decorator](https://github.com/onqtam/doctest/blob/master/doc/markdown/testcases.md#decorators).

Actually another option is to:
- extend the ```--list-test-cases``` option at least for the xml reporter to also have them inside of test suites - the current xml output is like [this](https://github.com/onqtam/doctest/blob/master/examples/all_features/test_output/list_test_cases_xml.txt) but the test cases could be nested into their appropriate test suites
- change the ```doctest_discover_tests``` cmake function to work with the xml output and parse that.

@RedDwarf69 introduced the ctest discovery so hopefully he could shed some light.

Mirko-von-Leipzig:
The xml reporter option looks like it would be the simplest. 

The alternative is to implement a test-case + test-suite matcher from the outside. I'm not sure if it can be done without ambiguity?

Something like:
1. --list-test-suites
2. --list-test-cases for each test suite (enabling only the one test suite)
3. Disable all test suites to get the no-test-suite list

I don't know the workings of doctest well enough to know if this would have corner cases or not. Extending this to nested test suites would also be a pain.

onqtam:
What you described should work :)

onqtam:
[...]

PR #464, opened 2021-01-29
Automatically add TEST_SUITE labels to discovered tests

shivupa:
The PR is against dev, and I have read the `CONTRIBUTING.md`. I understand that usually a issue is opened for discussion, but I figured the draft PR would illustrate the changes better.

## Description
What:
When using `doctest_discover_tests`, the `LABELS` property is set for the generated test based on the `TEST_SUITE`.

Why:
In a downstream project using CMake, I found it useful to group each `TEST_CASE` in a `TEST_SUITE`. I found myself wanting to selectively run tests based on what `TEST_SUITE` they belong to, e.g, `ctest -L TEST_SUITE_NAME`. The test discovery returns the list of test, but modifying each one would be tedious.

## GitHub Issues
#302 is related, but the discussion doesn't seem to indicate that this is the desired solution.

### DIFF
--- a/scripts/cmake/doctest.cmake
+++ b/scripts/cmake/doctest.cmake
@@ -32,6 +32,7 @@ same as the doctest name; see also ``TEST_PREFIX`` and ``TEST_SUFFIX``.
                          [TEST_PREFIX prefix]
                          [TEST_SUFFIX suffix]
                          [PROPERTIES name1 value1...]
+                         [ADD_LABELS value]
                          [TEST_LIST var]
                          [JUNIT_OUTPUT_DIR dir]
     )
@@ -85,6 +86,9 @@ same as the doctest name; see also ``TEST_PREFIX`` and ``TEST_SUFFIX``.
     Specifies additional properties to be set on all tests discovered by this
     invocation of ``doctest_discover_tests``.
 
+  ``ADD_LABELS value``
+    Specifies if the test labels should be set automatically.
+
   ``TEST_LIST var``
     Make the list of tests available in the variable ``var``, rather than the
     default ``<target>_TESTS``.  This can be useful when the same test
@@ -106,7 +110,7 @@ function(doctest_discover_tests TARGET)
     ""
     ""
     "TEST_PREFIX;TEST_SUFFIX;WORKING_DIRECTORY;TEST_LIST;JUNIT_OUTPUT_DIR"
-    "TEST_SPEC;EXTRA_ARGS;PROPERTIES"
+    "TEST_SPEC;EXTRA_ARGS;PROPERTIES;ADD_LABELS"
     ${ARGN}
   )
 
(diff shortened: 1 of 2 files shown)


onqtam:
Hi,

In a world where we had native XML parsing support in CMake it would have been better to switch to parsing the XML output in `doctestAddTests.cmake` instead of the console output - there you'll be able to get all test cases and their test suites (as an attribute to the test cases - see [here](https://github.com/onqtam/doctest/blob/master/examples/all_features/test_output/list_test_cases_xml.txt)) with a single `execute_process` call (as compared to the current approach - with one for each test case). But we don't live in that world...

My concern with the current approach is that if there are thousands of test cases there would be thousands of invocations of the binary just to discover the test suite of the test cases. Perhaps this labeling shouldn't be on by default and should be guarded in an `if()`...

I'll leave this PR open for now (hopefully will get more feedback from others as I'm not a user of these CMake scripts) but you should be free to use these changes locally for now.

shivupa:
> Hi,
> 
> In a world where we had native XML parsing support in CMake it would have been better to switch to parsing the XML output in `doctestAddTests.cmake` instead of the console output - there you'll be able to get all test cases and their test suites (as an attribute to the test cases - see [here](https://github.com/onqtam/doctest/blob/master/examples/all_features/test_output/list_test_cases_xml.txt)) with a single `execute_process` call (as compared to the current approach - with one for each test case). But we don't live in that world...
> 
> My concern with the current approach is that if there are thousands of test cases there would be thousands of invocations of the binary just to discover the test suite of the test cases. Perhaps this labeling shouldn't be on by default and should be guarded in an `if()`...

Sorry for the delay I had some deadlines. I also was worried about this. In my use case there are only a few (<100) tests so there is no perceptible overhead, but this clearly could vary. I hope this method can serve as a stand in until there is a cleaner XML based method. I wrapped this in an `if()` so now `doctest_discover_tests(main_test ADD_LABELS 1)` will add the labels.

> I'll leave this PR open for now (hopefully will get more feedback from others as I'm not a user of these CMake scripts) but you should be free to use these changes locally for now.

I'm fine with leaving this open for as long as you'd like. I am working from the branch for the time being so it doesn't hold me up!

onqtam:
thanks for the contribution!

jecassis:
Seeing a failure in test discovery when upgrading to 2.4.6 due to this feature when labels are not explicitly set. Example:

```
doctest_discover_tests(MyTests)
```

This is the error:

```console
 CMake Error at .../scripts/cmake/doctestAddTests.cmake:60 (if):
    if given arguments:
  
      "EQUAL" "1"
  
    Unknown arguments specified
```

To workaround, need to explicitly set `ADD_LABELS` to `0`:

```
doctest_discover_tests(MyTests ADD_LABELS 0)
```

Please take care of the scenario where `ADD_LABELS` is an empty string so that the original and non-argumented test discovery directive maintains its original behavior.

shivupa:
> doctest_discover_tests(MyTests)

Hi Thanks for reporting this I opened #489. I'll update there once the issue is fixed.

shivupa:
> Seeing a failure in test discovery when upgrading to 2.4.6 due to this feature when labels are not explicitly set. Example:
> 
> ```
> doctest_discover_tests(MyTests)
> ```
> 
> This is the error:
> 
> ```
>  CMake Error at .../scripts/cmake/doctestAddTests.cmake:60 (if):
>     if given arguments:
>   
>       "EQUAL" "1"
>   
>     Unknown arguments specified
> ```
> 
> To workaround, need to explicitly set `ADD_LABELS` to `0`:
> 
> ```
> doctest_discover_tests(MyTests ADD_LABELS 0)
> ```
> 
> Please take care of the scenario where `ADD_LABELS` is an empty string so that the original and non-argumented test discovery directive maintains its original behavior.

#490 fixes this.

jharmer95:
This is still broken, it can be fixed by changing it to the CMake-preferred syntax `if(${add_labels})` rather than compare a string to 1. See PR #527 (from issue #524) for the fix

PR #468, opened 2021-02-11
REQUIRE does not compile when operator== in different namespace #443 .

### LINKED ISSUE #443: REQUIRE does not compile when operator== in different namespace
Hi,

I have a simple equality that compiles normally but fails when placed within a `REQUIRE`.   This is a surprise to me, since I was hoping to replace many of my manual tests (which are based on  `f ( a ==  b)` ) by moving them into a REQUIRE.  (Aside, this same conditional compiles when  inside the equivalent CppUnit  macro).

To reproduce, please compile the file below.   I used gcc 7.5.0 on Ubuntu 18, doctest 2.4.1

```c++
#include "doctest.h"

namespace user {

struct label {
  label() : i(0) {}
  int i;
};

}

/* deliberately outside the user:: namespace */
bool operator==(const user::label& lhs, const user::label& rhs) {
  return lhs.i == rhs.i;
}

void foo() {
  user::label a;
  user::label b;

  if (a == b) return; /* this compiles */

  REQUIRE(a == b); /* does not compile */
}
```

Error:

```shell
g++   -Wfatal-errors  -pedantic -Wall --std=c++1z main.cc  
In file included from main.cc:2:0:
doctest.h: In instantiation of ‘doctest::detail::Result doctest::detail::Expression_lhs<L>::operator==(const R&) [with R = user::label; L = const user::label&]’:
main.cc:26:3:   required from here
doctest.h:1149:32: error: no match for ‘operator==’ (operand types are ‘const user::label’ and ‘const user::label’)
 #define DOCTEST_CMP_EQ(l, r) l == r
                                ^
doctest.h:1044:20: note: in expansion of macro ‘DOCTEST_CMP_EQ’
         bool res = op_macro(lhs, rhs);                                                             \
                    ^~~~~~~~
doctest.h:1186:9: note: in expansion of macro ‘DOCTEST_DO_BINARY_EXPRESSION_COMPARISON’
         DOCTEST_DO_BINARY_EXPRESSION_COMPARISON(==, " == ", DOCTEST_CMP_EQ) //!OCLINT bitwise operator in conditional
```

The declaration of the `operator==` outside the `user` namespace is significant; if I move it into `user`, everything does compiles.

navinp0304:
## Description

- This was a valid test case when you have global operator the comparision succeeds but doctest fails.
- This is due to forced instantiation of template member function `Expression_lhs.operator==(const R& rhs).`
- The template member function contains `lhs  == rhs` for equality operator. This comparison fails since the global operator is defined after the template and it doesn't take part in lookup.
-  In that case the function itself is not instantiated due to sfinae and the global operator is visible and due to user defined conversion the operator succeeds.

## GitHub Issues
Closes #443

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -443,6 +443,7 @@ DOCTEST_MSVC_SUPPRESS_WARNING_POP
 #ifdef DOCTEST_CONFIG_INCLUDE_TYPE_TRAITS
 #include <type_traits>
 #endif // DOCTEST_CONFIG_INCLUDE_TYPE_TRAITS
+#include<utility>
 
 namespace doctest {
 
@@ -1048,9 +1049,15 @@ namespace detail {
         return toString(lhs) + op + toString(rhs);
     }
 
+// This will check if there is any way it could find a operator like member or friend and uses it.
+// If not it doesn't find the operator or if the operator at global scope is defined after
+// this template, the template won't be instantiated due to SFINAE. Once the template is not
+// instantiated it can look for global operator using normal conversions.
+#define SFINAE_OP(ret,op) decltype(std::declval<const L&>() op std::declval<const R&>(),static_cast<ret>(0))
+
 #define DOCTEST_DO_BINARY_EXPRESSION_COMPARISON(op, op_str, op_macro)                              \
     template <typename R>                                                                          \
-    DOCTEST_NOINLINE Result operator op(const DOCTEST_REF_WRAP(R) rhs) {                           \
+    DOCTEST_NOINLINE SFINAE_OP(Result,op) operator op(const DOCTEST_REF_WRAP(R) rhs) {             \
         bool res = op_macro(lhs, rhs);                                                             \
         if(m_at & assertType::is_false)                                                            \
             res = !res;                                                                            \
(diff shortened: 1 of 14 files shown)


onqtam:
This solution adds SFINAE on the common instantiation path for all asserts and that would slow compile times - I'll leave this PR open for now and decide a bit later what to do with it.

I also wonder if `declval` can be implemented without including `<utility>` - perhaps something like what's posted [here](https://stackoverflow.com/questions/56212117/understanding-declval-optimized-implementation). Not including any headers would be really great.

navinp0304 (commented):
>I also wonder if declval can be implemented using only C++ without including <utility> - perhaps something like what's posted here. Not including any headers would be really great.

I removed utility and used the faster version of declval as per the link. Please pull the changes.

navinp0304:
>I also wonder if declval can be implemented using only C++ without including <utility> - perhaps something like what's posted here. Not including any headers would be really great.

I removed utility and used the faster version of declval as per the link. Please pull the changes.

onqtam:
sorry for the delay - will look at this again soon.

navinp0304:
> sorry for the delay - will look at this again soon.

This fixes 399 as well. Using rvalue refs than existing lvalue in func arguments.

onqtam:
Thanks for this - I'll merge it as it is and fix 1 tiny thing later on (doctest already has `remove_reference` in namespace `detail`)

PR #471, opened 2021-02-14
Fix MSVC linter warning about uninitialized TestSuite variables

Reedbeta:
In MSVC 2019 the Intellisense linter pops a warning whenever I used `TEST_SUITE`. It was unhappy that the `static doctest::detail::TestSuite data` was uninitialized. Fixed by default-initializing that in two places where it appears. Thanks!

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -1933,10 +1933,12 @@ int registerReporter(const char* name, int priority, bool isReporter) {
             static DOCTEST_NOINLINE doctest::detail::TestSuite& getCurrentTestSuite() {            \
                 DOCTEST_MSVC_SUPPRESS_WARNING_WITH_PUSH(4640)                                      \
                 DOCTEST_CLANG_SUPPRESS_WARNING_WITH_PUSH("-Wexit-time-destructors")                \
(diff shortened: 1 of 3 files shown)


Reedbeta:
OK, the CI is failing and I am not sure why this is not working when compiled as C++11...

Reedbeta:
I think the CI is just running this on some really old and buggy versions of gcc. 😕

onqtam:
I guess we should be able to get the old GCC builds running by silencing `-Wmissing-field-initializers` around the problematic pieces of code like this:
```c++
DOCTEST_GCC_SUPPRESS_WARNING_WITH_PUSH("-Wexit-time-destructors")
// ... code with ={};
DOCTEST_GCC_SUPPRESS_WARNING_POP
```

Reedbeta:
Alright, I put the warning suppress in, but it's still failing gcc 4.8 and 4.9. 🤪 Doesn't make any sense. I even tried it on godbolt to verify that warning can be suppressed, and it works fine there.

onqtam:
Let's merge this and I'll try to figure out how to get gcc 4.8 & 4.9 not to complain. Thanks!

PR #475, opened 2021-02-23
Using templated operator== inside TEST_CASE changes deduced types of forwarding references #399

### LINKED ISSUE #399: Using templated operator== inside TEST_CASE changes deduced types of forwarding references
I was writing some generic code that among other things allows customization of operators for user provided classes. This involved a lot of templates and forwarding references, so my tests also included cases that ensured that my templates are deducing the types that I wanted. However, it looked like the behavior was subtly different between the test code and production code which was very confusing. Ultimately, I manged to reduce my working example to the following (dummy implementations for brevity):
```C++
  template <typename T>
  bool operator==(T&&, T&&)
  {
    return true;
  }

  struct X {};

  TEST_CASE("test")
  {
    X x{};
    CHECK(x == X{});
  }
```
This code compiles (doctest 2.4.0 with gcc 10.0.1), but it should not. Indeed, if I use the comparison `x == X{}` outside of `TEST_CASE`, gcc says `deduced conflicting types for parameter ‘T’ (‘X&’ and ‘X’)` which sounds exactly right to me. The same thing happens with other comparison operators like `operator<`, but nothing else as far as I can tell.

I assume that doctest does some magic behind the scenes to parse expressions inside `CHECK` which is a cause of this and I can appreciate that it might be difficult to eliminate this side effect. But it would be great if there was a way to fix this or at least have some workaround.

navinp0304:
Please close this pull request. I planned to merge from dev to dev. Instead it came from master ->dev.
Using rvalue refs instead of const L& for function arguments and using forward for both lvalue and rvalue.

Closes #399


onqtam:
Ok, closing it, although I think you should be able to close it too.

I was going to respond soon on this PR - was considering merging it - are you going to open a new one?

PR #478, opened 2021-03-05
Fix compile error when targeting macOS version earlier and macOS 10.12

### LINKED ISSUE #466: std::uncaught_exceptions is only available if compiling for macOS 10.12 or above
Using doctest version 2.4.5 I'm getting a compile error for saying that std::uncaught_exceptions is only available for macos 10.12 or above. I need to target older macos systems so in my cmake project I use:

`set(CMAKE_OSX_DEPLOYMENT_TARGET "10.9" CACHE STRING "Minimum OS X deployment version" FORCE)`

This option adds an argument to each compilation unit specifying it should compile for 10.9 support. The code in doctest currently checks for if the compiler has the std::uncaught_exceptions feature, but not if the macos version supports it too. 

My current workaround is to just ignore std::uncaught_exceptions on apple since the deprecation warning is suppressed anyways. Here's the code for that:

```
#if defined(__cpp_lib_uncaught_exceptions) && __cpp_lib_uncaught_exceptions >= 201411L && !defined(__APPLE__)
        if(std::uncaught_exceptions() > 0) {
#else
        if(std::uncaught_exception()) {
#endif
```

SamWindell:
## Description
Doctest does not compile if you are targeting a macOS version earlier than 10.12. This patch fixes that.

std::uncaught_exceptions only available on 10.12 and above (despite it being in the standard library). It results in the compiler sending an error. Targeting an earlier version of macOS is done with the flag -mmacosx-version-min=x, where x is something like 10.9.

When you use the -mmacosx-version-min flag, a define is set: __MAC_OS_X_VERSION_MIN_REQUIRED. The patch checks against this number.

## GitHub Issues
Closes #466

### DIFF
--- a/doctest/parts/doctest.cpp
+++ b/doctest/parts/doctest.cpp
@@ -1062,7 +1062,7 @@ namespace detail {
                 g_cs->subcasesPassed.insert(g_cs->subcasesStack);
             g_cs->subcasesStack.pop_back();
 
-#if defined(__cpp_lib_uncaught_exceptions) && __cpp_lib_uncaught_exceptions >= 201411L
+#if defined(__cpp_lib_uncaught_exceptions) && __cpp_lib_uncaught_exceptions >= 201411L && (!defined(__MAC_OS_X_VERSION_MIN_REQUIRED) || __MAC_OS_X_VERSION_MIN_REQUIRED >= 101200)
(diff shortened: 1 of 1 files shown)


onqtam:
Thanks!

PR #480, opened 2021-03-11
Convert to bool by casting, rather than double negation

kitegi:
## Description

doctest allows asserting the validity of objects that are convertible to `bool`, this works for things like pointers, `optional<T>`, etc., which define an `explicit operator bool() const;`

this is done by `!!arg` which works for that purpose. but doesn't do the right thing in some other cases. for example, bool-like objects that are known at compile time can be implemented like this

```cpp
template <bool B>
struct constant_bool {
  constexpr constant_bool<!B> operator!() const noexcept { return {}; }
  explicit constexpr operator bool() const noexcept { return B; }
};
```

given some `constant_bool<true> b;`, `!b` has type `constant_bool<false>`, and `!!b` again has type `constant_bool<true>`.  
so `bool x = !!b;` doesn't work since the conversion operator is explicit

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -1184,7 +1184,7 @@ namespace detail {
                 , m_at(at) {}
 
         DOCTEST_NOINLINE operator Result() {
-            bool res = !!lhs;
+            bool res = static_cast<bool>(lhs);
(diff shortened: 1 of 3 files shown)


onqtam:
Makes sense. Thanks :)

PR #482, opened 2021-03-13
Improve speed with attached debugger (Windows)

### LINKED ISSUE #481: Assertions are slow when running on Windows with a debugger attached
## Description

Assertions are slow when running on Windows with a debugger attached.

The reason is the unconditional call of `OutputDebugStringA` which takes a comparatively very long time when a debugger is attached.

In my case the difference in execution time was dramatic. E.g. one test took over a minute when started from the IDE (with the debugger attached), while it takes less than a second when running without a debugger (same Debug build).

Also, in Visual Studio, the frequent calls of `OutputDebugStringA` with an empty string seem to overflow some buffer every now and then - Visual Studio periodically prints `*** Output truncated by debugger ***` in the debug window.

When I added a check so that `OutputDebugStringA` is only called when there's actually something to print, the execution time went down to about the time when running without a debugger (< 1 second in my case) -- and the `*** Output truncated by debugger ***` messages are gone too.

I've prepared a pull request.

### Steps to reproduce

* Write a simple test that does a lot of assertions (100k+) in otherwise trivial & fast code.
* Compile for Windows with Visual Studio, Debug configuration
* Start outside of Visual Studio without any debugger attached
* Start from inside Visual Studio with debugger attached

### Extra information

* doctest version: **v2.4.3**
* Operating System: **Windows 10 x64 1909**
* Compiler+version: **Visual Studio 2019 16.8.6**
* Configuration: **x64-Debug**

pgroke:
## Description

**What**: Only call `OutputDebugStringA` if there's actually something to print.

**Why**: This speeds up execution with a debugger attached dramatically.

## GitHub Issues

Closes #481

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -5900,8 +5900,10 @@ namespace {
         bool with_col = g_no_colors;                                                               \
         g_no_colors   = false;                                                                     \
         ConsoleReporter::func(arg);                                                                \
-        DOCTEST_OUTPUT_DEBUG_STRING(oss.str().c_str());                                            \
-        oss.str("");                                                                               \
+        if(oss.tellp() != std::streampos{}) {                                                      \
+            DOCTEST_OUTPUT_DEBUG_STRING(oss.str().c_str());                                        \
+            oss.str("");                                                                           \
+        }                                                                                          \
         g_no_colors = with_col;                                                                    \
     }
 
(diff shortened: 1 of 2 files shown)


onqtam:
Thanks for the contribution! Don't worry about the 2 failing builds.

Issue #398, opened 2020-07-19
-tc does not work with comma in names


martinus:
A simple reproducer is this:

```cpp
TEST_CASE("a,b") {}
```

It doesn't seem to be possible to run this test with `-tc`. I can run `./app -ltc` which prints

```
a,b
```

But running `./app -tc='a,b'` tries to run test `a` and test `b`, which both don't exist.

That's not a big problem in that case because I can just rename the test, but it becomes a problem with `TEST_CASE_TEMPLATE`. E.g. see this reproducer:

```cpp
template <typename X, typename Y>
class Foo {};
TYPE_TO_STRING(Foo<int, int>);

TEST_CASE_TEMPLATE("foo", T, Foo<int, int>) {}
```

Running `./app -ltc` prints 

```
foo<Foo<int, int>>
```

But since the test contains a `,` it is not possible to run just this test :cry:

onqtam:
Sorry for the late reply! This issue is the same for filtering test cases and [subcases](https://github.com/onqtam/doctest/issues/135) - the best I could offer for now is to use `?` (will match 1 character) instead of the commas (`,`), so in your case you could use `a?b`. Maybe I should put this in the FAQ & the relevant places in the docs for now... and also maybe escaping should be possible. Will keep this open until the problem is resolved.

onqtam:
a PR which allows commas to be escaped (with `\` as a prefix) was just merged in `dev`, however, I do think that this solution (as proposed in #135 ):
> or change the argument syntax so that if you want to run multiple subcases, you would type --sc='ab' --sc='cd'

would be better in the long term, so I'll keep this issue open

Saalvage:
Primary issue fixed, #135 will remain as the potential rest of this is figured out.

Issue #297, opened 2019-09-23
Test cases containing a comma cannot be run individually


Tradias:
## Description
Trying to run a test binary which contains a test_case that has a comma in its name with `"--test-case=[name with comma]"` causes no test to run.

### Steps to reproduce
Create a test case with a comma, like   
`TEST_CASE("a,b") {}`
Run it with `"--test-case=a,b` and expect that test to execute.  
Likewise when running it with `"--test-case=a\,b"` which means `ctest` is also unable to run any tests when using the `doctest_discover_tests` cmake function.
Actual result: All tests are skipped.

### Extra information
I noticed that there is some special handling for commas in the `doctest_discover_tests` cmake script, yet even without it, I cannot get the test to run using `ctest`.
I am using this `main` entry point, even though I doubt it

* doctest version: **v2.3.5**
* Operating System: **Ubuntu 19.04**
* Compiler+version: **GCC 9.1 and Clang 8.0**

onqtam:
here is a related issue around subcases and filtering of names with commas: https://github.com/onqtam/doctest/issues/135

so basically filtering has an issue around commas, but you can use ```?``` instead of ```,``` for now which basically means "match 1 character no matter what it is".

onqtam:
a PR which allows commas to be escaped (with `\` as a prefix) was just merged in `dev`, however I do think that this solution (as proposed in #135 ):
> or change the argument syntax so that if you want to run multiple subcases, you would type --sc='ab' --sc='cd'

would be better in the long term, so I'll keep this issue open

bevinhex:
Not only comma but also other special characters like ":",";" cause test to get skipped

Tradias:
@bevinhex works fine for me with version 2.4.5 and 2.4.6

bevinhex:
@Tradias Perhaps it is fixed on 2.4.5 then, I was using 2.4.4, because that is the latest version on conan center at the moment

Saalvage:
Primary issue fixed, #135 will remain as the potential rest of this is figured out.

PR #493, opened 2021-04-05
Handle escaped commas in parsed arguments

### LINKED ISSUE #135: --sc does not work if SUBCASE contains commas
## Description
I have a subcase which contains a comma and I can't filter it with --sc. Moreover it was very troubling because the wrong test was run.

### Steps to reproduce

Here's a simple program:

```cpp
#include "doctest/doctest.h"

TEST_CASE("tc") {
    SUBCASE("ab") {
        std::cout << "ab" << std::endl;
    }
    SUBCASE("ab, c") {
        std::cout << "abc" << std::endl;
    }
}
```

Run it with

    clang++ test.cpp -o test -std=c++14 && ./test --sc='ab, c'

You get "ab" instead of "ab, c"

### Extra information

This seems to be a feature to run multiple subcases, but it was quite misleading to me, I lost some time just to see that the wrong test was run.

I have two alternative suggestions to fix that:

- if the string is "ab, c" run all "ab", " c", and "ab, c" subcases
- or change the argument syntax so that if you want to run multiple subcases, you would type --sc='ab' --sc='cd'

---

* doctest version: master **cd1d747**
* Operating System: **debian unstable**
* Compiler+version: **clang 5.0.0**

friendlyanon:
## Description
Arguments to CLI flags will now handle escaped commas properly.
```sh
# A command similar to what the CMake test discovery module generates
example_test "--test-case=example group\, case"
# Will run test declared as TEST_CASE("example group, case")
```

## GitHub Issues
What prompted this PR https://github.com/Dobiasd/FunctionalPlus/issues/233
Closes #135 #297 #398

### DIFF
--- a/doctest/parts/doctest.cpp
+++ b/doctest/parts/doctest.cpp
@@ -3333,18 +3333,42 @@ namespace {
                            std::vector<String>& res) {
         String filtersString;
         if(parseOption(argc, argv, pattern, &filtersString)) {
-            // tokenize with "," as a separator
-            // cppcheck-suppress strtokCalled
-            DOCTEST_CLANG_SUPPRESS_WARNING_WITH_PUSH("-Wdeprecated-declarations")
-            auto pch = std::strtok(filtersString.c_str(), ","); // modifies the string
-            while(pch != nullptr) {
-                if(strlen(pch))
-                    res.push_back(pch);
-                // uses the strtok() internal state to go to the next token
-                // cppcheck-suppress strtokCalled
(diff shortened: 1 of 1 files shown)


onqtam:
I wonder if we should also change the `--list-test-cases` option to print the commas with a backslash for escaping the commas...

friendlyanon:
I think the documentation should be updated to say that a sequence of `\,` characters is tokenized as `,`, while a single comma is a list separator.

PR #505, opened 2021-05-17
Fix build with GCC 11

jktjkt:
GCC 11 started warning about unused-but-set variables. Here's how it
fails:
```
 examples/all_features/stringification.cpp: In function ‘void _DOCTEST_ANON_FUNC_20()’:
 examples/all_features/stringification.cpp:106:26: error: variable ‘bla2’ set but not used [-Werror=unused-but-set-variable]
   106 |     MyTypeInherited<int> bla2;
       |                          ^~~~
```
I don't think that that assignment is actually doing anything; in fact,
there's that other variable which is assigned in a similar manner, but
that one is also used for an assert.


jktjkt:
A gentle ping to see if I can get this one merged.

PR #512, opened 2021-06-05
Fix 'ignoring return value' warning with gcc 9.3

zeux:
Without this fix, GCC 9.3.0 generates the following warning when
comparing shared_ptr to nullptr:

```
extern/doctest.h:1239:49: error: ignoring return value of ‘bool std::operator!=(const std::shared_ptr<_Tp>&, std::nullptr_t) [with _Tp = Luau::Module; std::nullptr_t = std::nullptr_t]’, declared with attribute nodiscard [-Werror=unused-result]
 1239 |         DOCTEST_DO_BINARY_EXPRESSION_COMPARISON(!=, " != ", DOCTEST_CMP_NE) //!OCLINT bitwise operator in conditional
```


onqtam:
Just pushed the same changes instead of merging the PR as there were other changes that conflicted with it and this was easier. Thanks!

PR #518, opened 2021-06-24
specifying working directory for execute_process in doctest_discover_tests

philbucher:
## Description
quoting from https://github.com/catchorg/Catch2/pull/2039 :

> In a CMake project, when using `catch_discover_tests(WORKING_DIRECTORY [...])` it usually means that the test binary needs resources that are relative to said working directory. This could lead to UB or a crash when running `--list-test-names-only` or `--list-reporters` on the produced test binary i.e when a .dll is not lazy loaded and not available in the current PATH.
> This commit simply forwards the `WORKING_DIRECTORY` argument to the `execute_process()` calls.
> 
> This also remove the need of using CMake's own magic with `CMAKE_MSVCIDE_RUN_PATH` and the like...

## GitHub Issues
See https://github.com/catchorg/Catch2/pull/2039

### DIFF
--- a/scripts/cmake/doctestAddTests.cmake
+++ b/scripts/cmake/doctestAddTests.cmake
@@ -39,6 +39,7 @@ execute_process(
   COMMAND ${TEST_EXECUTOR} "${TEST_EXECUTABLE}" ${spec} --list-test-cases
   OUTPUT_VARIABLE output
   RESULT_VARIABLE result
+  WORKING_DIRECTORY "${TEST_WORKING_DIR}"
 )
 if(NOT ${result} EQUAL 0)
   message(FATAL_ERROR
@@ -63,6 +64,7 @@ foreach(line ${output})
       COMMAND ${TEST_EXECUTOR} "${TEST_EXECUTABLE}" --test-case=${test} --list-test-suites
       OUTPUT_VARIABLE labeloutput
       RESULT_VARIABLE labelresult
+      WORKING_DIRECTORY "${TEST_WORKING_DIR}"
     )
     if(NOT ${labelresult} EQUAL 0)
       message(FATAL_ERROR


philbucher:
For anyone reading this, here is how I use it:
`octest_discover_tests(tests_exe WORKING_DIRECTORY $<TARGET_FILE_DIR:my_lib>)`
Note that I am using generator expressions but multi config builds

I struggled with this a bit when using Visual Studio, but I think the generator expression is the correct usage

friendlyanon:
> For anyone reading this, here is how I use it:
> `doctest_discover_tests(tests_exe WORKING_DIRECTORY $<TARGET_FILE_DIR:my_lib>)`

This only works if you have a single shared library dependency. This is still not sufficient if you have multiple shared library dependencies in various locations.

You'd need to pass a list of paths to add to `PATH` in the doctest script before `execute_process()`.

philbucher:
> > For anyone reading this, here is how I use it:
> > `doctest_discover_tests(tests_exe WORKING_DIRECTORY $<TARGET_FILE_DIR:my_lib>)`
> 
> This only works if you have a single shared library dependency. This is still not sufficient if you have multiple shared library dependencies in various locations.
> 
> You'd need to pass a list of paths to add to `PATH` in the doctest script before `execute_process()`.

Sure you are right

However my usecase is only an example, I can imagine it also to be useful for other cases to specify the working dir

BTW do you know how to set the `PATH` with CMake? Is there a native/good solution for this? I couldn't find it

friendlyanon:
https://cmake.org/cmake/help/latest/command/set.html#set-environment-variable

friendlyanon:
@philbucher FYI, I have implemented this in my fork of [utest.h](https://github.com/friendlyanon/utest.h/commit/3337e926073c6cc323b8ced96209cd863d524b81) and you can see it in action [here](https://github.com/friendlyanon/cmake-init-gif-engine/tree/2185e51daa61d2809362ace7ff0264356e5aa146).

If you want to later also make a PR for this, then this might be a good start for that.

Issue #524, opened 2021-07-22
CMAKE: doctest_discover_tests() error when ADD_LABELS is not specified


ArcEarth:
## Description
Regression introduced by https://github.com/onqtam/doctest/pull/464
When `doctest_discover_tests(<target>)` does not have `ADD_LABELS ...` options, test discovery breaks with the following error: 
```
[build] CMake Error at doctest/2.4.6/lib/cmake/doctestAddTests.cmake:60 (if):
[build]   if given arguments:
[build] 
[build]     "EQUAL" "1"
[build] 
[build]   Unknown arguments specified
```

### Steps to reproduce
Building anything with `doctest_discover_tests(<target>)` without extra options.

* doctest version: **v2.4.6**
* cmake version: **v3.19.4**
* Operating System: **OSX 11.4**
* Compiler+version: **Clang 11.0.4**

jharmer95:
I have encountered this as well. Particularly annoying because I can't easily change the file for my CI/CD
Line 60 can be changed to either

`if (add_labels EQUAL 1)`

or

`if (${add_labels})`

I can prepare a PR for this fix

onqtam:
should be fixed in `dev`!

PR #490, opened 2021-03-30
Set variable to 0 if not set

shivupa:
## Description
Set a variable in the CMake to zero if it is not defined.

## GitHub Issues
Closes #489

PR #527, opened 2021-08-02
Fixing issue with doctestAddTests.cmake

jharmer95:
Newer versions of CMake won't work with line 60: `if(${add_labels} EQUAL 1)`, changed it to simply `if (${add_labels})`

## Description

Builds using the CMake `doctest_discover_tests` function are failing due to a syntax error. This is likely due to newer versions of CMake syntax. This fixes those issues by updating the syntax.

## GitHub Issues

Addresses #524

### DIFF
--- a/scripts/cmake/doctestAddTests.cmake
+++ b/scripts/cmake/doctestAddTests.cmake
@@ -57,7 +57,7 @@ foreach(line ${output})
   endif()
   set(test ${line})
   set(labels "")
-  if(${add_labels} EQUAL 1)
+  if(${add_labels})
     # get test suite that test belongs to
     execute_process(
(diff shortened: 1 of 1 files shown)


jharmer95:
Not sure if the cmake scripts are tested in the project's CI/CD, but it could be done by creating a basic CMake project that uses these scripts to detect tests. This would allow for future regressions to be spotted

shivupa:
Hi Can you also revert the changes in #490 in this PR. Those are not necessary with this fix. Thanks

b1ackviking:
Is there a way to speed up the inclusion of this PR?

onqtam:
> Not sure if the cmake scripts are tested in the project's CI/CD, but it could be done by creating a basic CMake project that uses these scripts to detect tests. This would allow for future regressions to be spotted

PRs welcome :)

PR #544, opened 2021-11-01
cmake: Add deduplication of test cases when they have the same name

nextsilicon-itay-bookstein:
Because in doctest multiple test cases may share the same name
regardless of suite scope (if any), and they therefore cannot
always be addressed individually,  improve CTest discovery
behavior by giving up parallelism and exposing them as one test
to CTest. This will result in running N such name-colliding tests
once in series rather than the current behavior of exposing them
as N CTest tests which each run these N doctest tests in series.

Signed-off-by: Itay Bookstein <itay.bookstein@nextsilicon.com>

### DIFF
--- a/scripts/cmake/doctestAddTests.cmake
+++ b/scripts/cmake/doctestAddTests.cmake
@@ -50,12 +50,31 @@ endif()
 
 string(REPLACE "\n" ";" output "${output}")
 
+function(is_doctest_line line result)
+  set(_result FALSE)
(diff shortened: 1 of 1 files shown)


onqtam:
Not sure this should be the default behavior - would like some input from others that are using these scripts...

nextsilicon-itay-bookstein:
I agree that it's a bit peculiar and wasn't sure what the 'best' way to tackle this is; but I think running N^2 tests instead of N is more of a misbehavior than this wonky deduplication. I arrived at this solution as a consequence of the fact that tests are (understandably) not uniqued by name, even when you consider suite scoping.

mitchgrout:

PR #550, opened 2021-11-09
Remove unused list_query_results() function

ivankochin:
## Description

The list_query_results() function is never used, so it may be removed.


onqtam:
Thanks for pointing this out - indeed it's not needed!

However, I'll close the PR and manually push the change as this needs to be edited in `doctest.cpp` form which `doctest.h` is generated.

PR #555, opened 2021-12-03
Add Intel Compiler support

isaevil:
## Description
Support for compilation with Intel Compiler


doctest/doctest.h
```diff
@@ -1313,6 +1329,11 @@ DOCTEST_CLANG_SUPPRESS_WARNING_POP
         int         m_expected_failures;
         double      m_timeout;
 
+        // Fix segfault on Intel Compiler 2021 during the DOCTEST_TEST_SUITE_END execution
```
onqtam, line 1332: Weird - I can't think of why it would segfault and there are no issues with all the other compilers & sanitizers & valgrind... Any clue why this might be happening?

doctest/doctest.h
```diff
@@ -311,6 +317,12 @@ DOCTEST_MSVC_SUPPRESS_WARNING(26812) // Prefer 'enum class' over 'enum'
 #define DOCTEST_ALIGNMENT(x) __attribute__((aligned(x)))
 #endif
 
+// Work-around for the warning: 'routine is both "inline" and "noinline"'
```
onqtam, line 320: I suppose this is because of this line: `inline DOCTEST_NOINLINE void der::f()` (here: https://github.com/doctest/doctest/blob/master/doctest/doctest.h#L1872 )

The reason it's both inline & noinline is because it needs to be a weak symbol (inline) and thus be able to be defined in multiple translation units (if coming from a header file) but we still don't want it to be inline-able (as a compiler optimization) wherever it's called (can't remember why I made it non-inline-able though...).

Other compilers don't have a problem with it - perhaps the ICC warning should be suppressed in another way and not by completely undefining `DOCTEST_NOINLINE`? `DOCTEST_NOINLINE` helps a lot with compile speeds.

isaevil: The error line as follows:
```
/common/doctest.h(1251): error #2196: routine is both "inline" and "noinline"
          DOCTEST_DO_BINARY_EXPRESSION_COMPARISON(==, " == ", DOCTEST_CMP_EQ) //!OCLINT bitwise operator in conditional
```

doctest/doctest.h
isaevil: I have rechecked without it. I guess it was a workaround for older versions of Intel Compiler 2021. It is not needed.

doctest/doctest.h
onqtam: judging by the documentation here:
https://community.intel.com/t5/Intel-C-Compiler/Where-is-the-online-documentation-for-suppressing-warnings/m-p/1111943

seems that those lines with `DOCTEST_DO_BINARY_EXPRESSION_COMPARISON` can be surrounded with:
```c++
#if DOCTEST_ICC
#pragma warning (push)
#pragma warning (disable:2196)
#endif // ICC
...
#if DOCTEST_ICC
#pragma warning (pop)
#endif // ICC
```

But I can't test this... Can you?

isaevil: It's weird, but warning still exists until I remove `#pragma warning (pop)` stuff.

yeputons: My understanding is that Intel C++ Compiler's warning suppression plays weirdly with macros. E.g. it's possible that suppression does not work at the macro expansion point (because the warning is actually inside a macro) and simultaneously does not work at the macro definition point (because it's expanded elsewhere). For example, if I take the latest released doctest 2.4.7 and [the factorial example](https://github.com/doctest/doctest/blob/a384864346f48c7ff5608051c176e7bd505712ae/doc/markdown/tutorial.md#a-simple-example), disable 2196 for the whole TU and enable it at the very bottom, I still get warnings:

```
a.cpp(13): warning #2196: routine is both "inline" and "noinline"

doctest.h(1320): warning #2196: routine is both "inline" and "noinline"
          DOCTEST_DO_BINARY_EXPRESSION_COMPARISON(==, " == ", DOCTEST_CMP_EQ) //!OCLINT bitwise operator in conditional
          ^
```

yeputons:
By the way, it's fairly straightforward to obtain Intel C++ Compiler Classic (`icc`, not to be confused with newer Clang-based DPC++/C++ compilers like `icx`, see [here](https://www.intel.com/content/www/us/en/developer/articles/technical/adoption-of-llvm-complete-icx.html)) as it's a part of [Intel oneAPI Base Toolkit](https://www.intel.com/content/www/us/en/developer/tools/oneapi/base-toolkit.html), which is available for Windows, Linux and macOS.

E.g. for Ubuntu (should be possible on GitHub Actions as well, I think):

1. Add Intel APT repositories, see ["Pre-installation Steps" here](https://www.intel.com/content/www/us/en/develop/documentation/installation-guide-for-intel-oneapi-toolkits-linux/top/installation/install-using-package-managers/apt.html).
2. `sudo apt update && sudo apt install intel-oneapi-compiler-dpcpp-cpp-and-cpp-classic` (few GB of 64 bit compilers instead of full-blown 15GB `intel-basekit` with bells and whistles)
3. Set up `PATH` and other environment variables by `. /opt/intel/oneapi/setvars.sh` (details are [here](https://www.intel.com/content/www/us/en/develop/documentation/get-started-with-intel-oneapi-base-linux/top/run-a-sample-project-using-the-command-line.html))
4. Run `icc main.cpp` to compile to `a.out`, basic flags are similar to GCC's

doctest/doctest.h
```diff
@@ -1162,7 +1174,11 @@ DOCTEST_CLANG_SUPPRESS_WARNING_WITH_PUSH("-Wunused-comparison")
 // If not it doesn't find the operator or if the operator at global scope is defined after
 // this template, the template won't be instantiated due to SFINAE. Once the template is not
 // instantiated it can look for global operator using normal conversions.
+#if !DOCTEST_ICC
 #define SFINAE_OP(ret,op) decltype((void)(doctest::detail::declval<L>() op doctest::detail::declval<R>()),ret{})
+#else
+#define SFINAE_OP(ret,op) ret
+#endif
```
Saalvage, line 1181: This actually breaks for me using ICC (19.2). Simply removing the `#if` fixes it.

onqtam: @isaevil would it work without these last few lines of changes?

isaevil: @onqtam @Saalvage No, it doesn't work for me. I am getting this kind of error:
```
doctest.h(853): error: too many recursive substitutions of function template signatures
      template<typename T> auto declval() DOCTEST_NOEXCEPT -> decltype(declval<T>(0)) ;
```

Saalvage: So this will need another switch based on whether it's running on Windows

Saalvage:
So this'll be included in 2.4.10 which will hopefully be released within the week.

However, the issue with the `SFINAE` workaround that was originally in this PR is still present: https://github.com/doctest/doctest/actions/runs/3176827254/jobs/5176570218
Therefore, I have decided to remove it for now. There has to be something that is different in your configuration, as it is working fine for me both on Windows and in the Ubuntu CI. Do you have any idea what that could be @isaevil? I'd really like to branch on that difference so that it will work for everyone and add CI for it as well.

On a side note, is there any better way to make sure the Intel compiler is in the path other than executing the script that puts it there in every step? That is definitely cumbersome but I've found nothing else that works (e.g. `bashrc` doesn't work because it is not being used for GH Actions).

yeputons:
> On a side note, is there any better way to make sure the Intel compiler is in the path other than executing the script that puts it there in every step?

I'm not aware of any. The script is somewhat complicated, and GitHub Actions [does not preseve environment variables between steps](https://github.com/orgs/community/discussions/26407). You can set environment variables that persist between steps (see [here](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable)), but I'm not sure it's feasible with `setvars.sh`'s setting multiple variables.

Saalvage:
So I've decided on making the workaround an optional config option, even on Intel compilers. It doesn't seem to always occur, and I'd generally prefer just manually suppressing the warning, but it doesn't seem like ICC correctly supports that in-code atm.

isaevil:
@Saalvage many thanks for working on this. Unfortunately, I didn't have time to address all the comments and track this by myself.
Is there something I should do?

Saalvage:
@isaevil I'm sorry for my late response. The only remaining issue that I can see is `SFINAE_OP` causing issues in your setup, but its removal causing issues in all of mine (see e.g. [here](https://github.com/doctest/doctest/actions/runs/3176827254/jobs/5176570218)). For the sake of getting 2.4.10 out of the door I am going to merge this for now, as some support will be better than none, but I would very much appreciate you opening another issue if the `SFINAE_OP` issue persists.

PR #556, opened 2021-12-03
Fix error: missing initializer for member doctest::detail::TestSuite

isaevil:
## Description
Fixing error: missing initializer for member doctest::detail::TestSuite: <member name>


onqtam:
Thanks - they were being initialized in `TestSuite& TestSuite::operator*(const char* in)` but that's actually ugly and error prone.

PR #557, opened 2021-12-03
Fix redefinition error while using double time DOCTEST_ANONYMOUS(DOCTEST_CAPTURE_)

isaevil:
## Description
Double `DOCTEST_ANONYMOUS(DOCTEST_CAPTURE_)` leads to redefinition error on MSVS like this:
`error C2082: redefinition of formal parameter 'DOCTEST_CAPTURE_54'`

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -2050,7 +2050,8 @@ int registerReporter(const char* name, int priority, bool isReporter) {
 
 // for logging
 #define DOCTEST_INFO(...)                                                                          \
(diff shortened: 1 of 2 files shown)


onqtam:
I guess this is when either `__COUNTER__` isn't available or somehow doesn't work properly?

Merging, will later add clang-format off & on suppressions around it so that it doesn't get re-formatted to what it was previously in the future. Thanks!

isaevil:
> I guess this is when either `__COUNTER__` isn't available or somehow doesn't work properly?

Correct :)

PR #558, opened 2021-12-03
Add Universal Windows Platform support

isaevil:
Support compilation with Universal Windows Platform by fixing compilation errors like these:
```
doctest.h(4272): error C3861: 'SetThreadStackGuarantee': identifier not found
doctest.h(4305): error C2065: 'SEM_FAILCRITICALERRORS': undeclared identifier
doctest.h(4305): error C2065: 'SEM_NOALIGNMENTFAULTEXCEPT': undeclared identifier
doctest.h(4306): error C2065: 'SEM_NOGPFAULTERRORBOX': undeclared identifier
doctest.h(4306): error C2065: 'SEM_NOOPENFILEERRORBOX': undeclared identifier
doctest.h(4324): error C3861: 'SetThreadStackGuarantee': identifier not found

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -254,6 +254,10 @@ DOCTEST_MSVC_SUPPRESS_WARNING(4623) // default constructor was implicitly define
 // MSVC++ 9.0       _MSC_VER == 1500 (Visual Studio 2008)
 // MSVC++ 8.0       _MSC_VER == 1400 (Visual Studio 2005)
 
+// Universal Windows Platform support
+#ifdef WINAPI_FAMILY
+#define DOCTEST_CONFIG_NO_WINDOWS_SEH
+#endif // WINAPI_FAMILY
 #if DOCTEST_MSVC && !defined(DOCTEST_CONFIG_WINDOWS_SEH)
 #define DOCTEST_CONFIG_WINDOWS_SEH
 #endif // MSVC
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by onqtam** [comment]
I think a cleaner way to do this would be to add the follownig:
```c++
#ifdef WINAPI_FAMILY
#define DOCTEST_CONFIG_NO_WINDOWS_SEH
#endif // WINAPI_FAMILY
```

in doctest_fwd.h - right around line 230 before `#if DOCTEST_MSVC && !defined(DOCTEST_CONFIG_WINDOWS_SEH)`.

Along with a comment - I'm not familiar at all with what `WINAPI_FAMILY` is and the relations with `WINAPI_FAMILY_PHONE_APP` & `WINAPI_PARTITION_DESKTOP`.

isaevil:
> I think a cleaner way to do this would be to add the follownig:
> 
> ```c++
> #ifdef WINAPI_FAMILY
> #define DOCTEST_CONFIG_NO_WINDOWS_SEH
> #endif // WINAPI_FAMILY
> ```
> 
> in doctest_fwd.h - right around line 230 before `#if DOCTEST_MSVC && !defined(DOCTEST_CONFIG_WINDOWS_SEH)` - along with a comment.
> 
> I'm not familiar at all with `WINAPI_FAMILY` and the relations with `WINAPI_FAMILY_APP` & `WINAPI_FAMILY_PHONE_APP` & `WINAPI_PARTITION_DESKTOP`, but there are such checks already in doctest.cpp (`#if defined(WINAPI_FAMILY) && (WINAPI_FAMILY == WINAPI_FAMILY_APP)`) which are specific - much better than something as broad as `WINAPI_FAMILY`.
> 
> And btw thanks for the contributions - even though I request changes in each and every single of your PRs - more compiler/platform support is welcome!

I guess this check you mentioned `#if defined(WINAPI_FAMILY) && (WINAPI_FAMILY == WINAPI_FAMILY_APP)` is actually applicable here too because as I can see we define  `/DWINAPI_FAMILY=WINAPI_FAMILY_APP` in our project.

PR #559, opened 2021-12-03
Mark variables as volatile to suppress `unused variable` warnings

isaevil:
Examples of errors:
```
doctest.h(4021): error #177: variable "doctest::<unnamed>::dummy_init_console_colors" was declared but never referenced
      int dummy_init_console_colors = colors_init();
```
```
error #2415: variable "DOCTEST_ANON_VAR_36" of static storage duration was declared but never referenced
```
It can be reproduced on Windows 10 with Intel Compiler(icl)

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -4031,7 +4031,7 @@ namespace {
         return 0;
     }
 
-    int dummy_init_console_colors = colors_init();
+    volatile int dummy_init_console_colors = colors_init();
 #endif // DOCTEST_CONFIG_COLORS_WINDOWS
 
     DOCTEST_CLANG_SUPPRESS_WARNING_WITH_PUSH("-Wdeprecated-declarations")
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -385,7 +385,7 @@ DOCTEST_MSVC_SUPPRESS_WARNING(4623) // default constructor was implicitly define
 #define DOCTEST_GLOBAL_NO_WARNINGS(var)                                                            \
     DOCTEST_CLANG_SUPPRESS_WARNING_WITH_PUSH("-Wglobal-constructors")                              \
     DOCTEST_CLANG_SUPPRESS_WARNING("-Wunused-variable")                                            \
-    static const int var DOCTEST_UNUSED // NOLINT(fuchsia-statically-constructed-objects,cert-err58-cpp)
+    static const volatile int var DOCTEST_UNUSED // NOLINT(fuchsia-statically-constructed-objects,cert-err58-cpp)
 #define DOCTEST_GLOBAL_NO_WARNINGS_END() DOCTEST_CLANG_SUPPRESS_WARNING_POP
 
 #ifndef DOCTEST_BREAK_INTO_DEBUGGER
--- a/doctest/parts/doctest.cpp
+++ b/doctest/parts/doctest.cpp
@@ -1204,7 +1204,7 @@ namespace {
         return 0;
     }
 
-    int dummy_init_console_colors = colors_init();
+    volatile int dummy_init_console_colors = colors_init();
 #endif // DOCTEST_CONFIG_COLORS_WINDOWS
 
     DOCTEST_CLANG_SUPPRESS_WARNING_WITH_PUSH("-Wdeprecated-declarations")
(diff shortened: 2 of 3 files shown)


isaevil:
> How do you get warnings for unused variables? Which compiler/OS?
> 
> Currently, the `DOCTEST_UNUSED` macro handles it fine for all tested compilers. Maybe you're using ICC (judging by the rest of the PRs)? Maybe `DOCTEST_UNUSED` should be extended to work with ICC - I don't like this use of `volatile` just to silence a warning in some circumstances. If `DOCTEST_UNUSED` can't be extended maybe the ICC-specific warning can be suppressed some other way?

I have updated the PR description with info about OS and compiler with error examples. `volatile` solution is more of a workaround. We definitely should look for more nice solution.

onqtam:
You can try adding this at the top/bottom of the 2 parts of the doctest header:

```c++
#if DOCTEST_ICC
#pragma warning (push)
#pragma warning (disable:177)
#endif // ICC

// doctest code

#if DOCTEST_ICC
#pragma warning (pop)
#endif // ICC
```

let me know if it works that way!

onqtam:
Oh... just noticed that one of the problems is the `DOCTEST_GLOBAL_NO_WARNINGS` macro and pragmas around the entire contents of the header won't work for that because it will be expanded by the preprocessor outside of the header... Maybe we can add this:

```c++
#ifndef DOCTEST_GLOBAL_VAR_EXTRA_ATTR
#define DOCTEST_GLOBAL_VAR_EXTRA_ATTR
#endif
```

and stick it instead of the volatile keyword - that way you'd be able to define it locally to `volatile` for yourself before the 
including the header but still not affect others - I really don't like the use of volatile for suppressing a warning...

Btw I wonder if `__pragma` would work for ICC - like for MSVC:
https://github.com/doctest/doctest/blob/master/doctest/doctest.h#L123-L134

isaevil:
> Oh... just noticed that one of the problems is the `DOCTEST_GLOBAL_NO_WARNINGS` macro and pragmas around the entire contents of the header won't work for that because it will be expanded by the preprocessor outside of the header... Maybe we can add this:
> 
> ```c++
> #ifndef DOCTEST_GLOBAL_VAR_EXTRA_ATTR
> #define DOCTEST_GLOBAL_VAR_EXTRA_ATTR
> #endif
> ```
> 
> and stick it instead of the volatile keyword - that way you'd be able to define it locally to `volatile` for yourself before the including the header but still not affect others - I really don't like the use of volatile for suppressing a warning...
> 
> Btw I wonder if `__pragma` would work for ICC - like for MSVC: https://github.com/doctest/doctest/blob/master/doctest/doctest.h#L123-L134

I actually still don't understand why this variable is needed. All what `colors_init` returns is `0`...

isaevil:
> Oh... just noticed that one of the problems is the `DOCTEST_GLOBAL_NO_WARNINGS` macro and pragmas around the entire contents of the header won't work for that because it will be expanded by the preprocessor outside of the header... Maybe we can add this:
> 
> ```c++
> #ifndef DOCTEST_GLOBAL_VAR_EXTRA_ATTR
> #define DOCTEST_GLOBAL_VAR_EXTRA_ATTR
> #endif
> ```
> 
> and stick it instead of the volatile keyword - that way you'd be able to define it locally to `volatile` for yourself before the including the header but still not affect others - I really don't like the use of volatile for suppressing a warning...
> 
> Btw I wonder if `__pragma` would work for ICC - like for MSVC: https://github.com/doctest/doctest/blob/master/doctest/doctest.h#L123-L134

`__pragma` works only for the second warning and only without `pop`...
`dummy_init_console_colors ` can be workarounded for `ICC` case by casting it to void somewhere. For example, in the next declared function:
```c++
        DOCTEST_CLANG_SUPPRESS_WARNING_WITH_PUSH("-Wdeprecated-declarations")
            void color_to_stream(std::ostream& s, Color::Enum code) {
#if defined(DOCTEST_CONFIG_COLORS_WINDOWS) && DOCTEST_ICC
            static_cast<void>(dummy_init_console_colors);
#endif
            static_cast<void>(s);    // for DOCTEST_CONFIG_COLORS_NONE or DOCTEST_CONFIG_COLORS_WINDOWS
            static_cast<void>(code); // for DOCTEST_CONFIG_COLORS_NONE
```

onqtam:
These 2 PRs have silenced the warnings for the intel compiler as well:
https://github.com/doctest/doctest/pull/575
https://github.com/doctest/doctest/pull/576

Closing this!

datadiode:
> These 2 PRs have silenced the warnings for the intel compiler as well: #575 #576
> 
> Closing this!

@mitchgrout This PR had the advantage of not breaking TEST_CASE_CLASS. Please reconsider.

mitchgrout:
@datadiode would you mind creating an issue to describe what the need is here; half since I'm not familiar with this change, half to avoid pinging the original authors on such an old PR. Thanks!

PR #560, opened 2021-12-03
Workaround access violation

isaevil:
## Description
Due to race between exiting the process and starting of a new detached thread in `Windows`, thread local variables, which constructors or destructors have calls to runtime functions (e.g.` free()` function) can cause access violation since `TBB` along with runtime library may have been unloaded by the time these variables are constructed or destroyed. The workaround is do not call constructors and/or destructors for such variables if they are not used.


doctest/doctest.h
```diff
@@ -454,6 +454,36 @@ namespace doctest {
 
 DOCTEST_INTERFACE extern bool is_running_in_test;
 
+#if _MSC_VER
```
onqtam, line 457: This breaks on other compilers when it's undefined - all builds besides windows clang/cl fail because of this - maybe use DOCTEST_MSVC? Builds must pass before this can get merged.

Also, it would be good to have a good fat comment explaining why this code is needed so that people don't have to look at git history & git blame.

onqtam:
Is this still an issue? The code has been refactored since

doctest/parts/doctest_fwd.h
```diff
@@ -476,6 +476,36 @@ namespace doctest {
 
 DOCTEST_INTERFACE extern bool is_running_in_test;
 
+#if DOCTEST_MSVC
```
iboB, line 479: this is a Windows-specific problem, so builds with mingw or clang should also be affected. Perhaps the correct check here is `DOCTEST_PLATFORM_WINDOWS`

doctest/parts/doctest_fwd.h
```diff
@@ -476,6 +476,36 @@ namespace doctest {
...
+template <typename T>
+struct doctest_thread_local_wrapper {
```
iboB, line 481: this type is only used in the implementation. There doesn't seem to be a need for it to be in the fwd part. It can be added to doctest.cpp instead.

iboB (commented):
doctest doesn't start any threads on its own. Can you please provide scenario which is affected by this PR?

belkiss:
Heya!

Any way this could get integrated? This patch is the only way I found to get version 2.49 to work without issue on windows...

Thanks!

mitchgrout:

Issue #503, opened 2021-05-02
[FEATURE REQUEST] Quiet flag


gamagan:
Please add a 'quiet' flag where Doctest shows none of its usual output.

My particular usecase is that I use Doctest for testing a 3D Renderer. I run each testcase individually, and each test runs twice: once for setting up the test frame data, and another time to verify the rendered frame. Since each test is ran individually, and each test runs twice, the console contains lots of noise from Doctest's usual output. 

* doctest version: 2.4.5
* Operating System: Windows 10, 64bit
* Compiler+version: MSVC 2019, latest

gamagan:
I managed to get the behavior that I wanted(on Windows 10) by doing:
`doctest::Context::setOption("out", "NUL")`
on the Context instance.

onqtam:
That redirects the output to a file and indeed can be used as a workaround. Leaving this open for now

onqtam:
2 new flags pushed in `dev` branch (soon in official version in `master):
- `-quiet` - quiets everything regardless
- `-minimal` - outputs only on failure

Closing this - let us know if these aren't sufficient!

PR #561, opened 2021-12-07
Quiet flag

Saalvage:
## Description
Added a quiet cli flag that disables all output and a method to the context class to allow for redirection of output.

I personally needed the latter since my application was thrashing `std::cout` and felt the former was easy enough to implement with it.

I was not entirely sure where the flag should go in the documentation, and thus decided to place it at an arbitrary location I deemed fit. If this is deemed incorrect let me know and I will move it. The same can be said about the `DiscardOStream` class.

## GitHub Issues
Implements feature #503


onqtam:
Looks good!

However, I wonder if there should be some output on failure - depends on what one would want to see from a quiet flag...

There should also be a test for this - I can add one after merge.

onqtam:
I guess quiet-ness can be further developed either with additional flags or changes - merging this now!

onqtam:
Just added a test and noticed that the xml/junit reports are also quiet... perhaps the quiet flag should not be compatible when specifying a reporter different from console as that's nonsense, but let's leave it like that for now - just adding a comment for that here for future reference.

PR #562, opened 2021-12-09
Add -minimal flag

Saalvage:
## Description
As discussed in #561 a less extreme form of -quiet is probably pretty useful as well, excludes the header and summary, leaving only all failing tests.

P.S.: Sorry for not including a test with my previous PR!

### DIFF
--- a/doctest/parts/doctest.cpp
+++ b/doctest/parts/doctest.cpp
@@ -886,6 +886,7 @@ void Context::setOption(const char*, const char*) {}
 bool Context::shouldExit() { return false; }
 void Context::setAsDefaultForAssertsOutOfTestCases() {}
 void Context::setAssertHandler(detail::assert_handler) {}
+void Context::setCout(std::ostream* out) {}
 int  Context::run() { return 0; }
 
 IReporter::~IReporter() = default;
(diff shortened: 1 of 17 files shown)


doctest/parts/doctest.cpp
```diff
@@ -3076,9 +3077,15 @@ namespace {
...
 
-        void test_run_start() override { printIntro(); }
+        void test_run_start() override {
+            if(!opt.minimal)
+                printIntro();
+        }
 
         void test_run_end(const TestRunStats& p) override {
+            if(opt.minimal)
+                return;
```
onqtam, line 3087: I think it makes sense to print the outro if anything has failed - as a summary. There's a variable a few lines below called `anythingFailed` which could be checked for that

Saalvage: I'm wondering if this should only depend on the failure of test cases, since assertions could fail with `doctest::{may, should}_fail` and perhaps even `doctest::no_output` in which case I wouldn't expect a summary with minimal.

If you disagree I'll keep it as is (and fix the test).

onqtam: You are right, I hadn't considered those cases!

PR #564, opened 2021-12-09
Make String::operator+ non-member

Saalvage:
## Description
Having operator+ as a non-member allows for type coercion on the lhs value.


onqtam:
LGTM!

PR #566, opened 2021-12-11
Clean up warning suppression a bit; Fixes #565

### LINKED ISSUE #565: MSVC warnings leak through when using the library as a single header with /Wall
## Description
doctest seemingly has a list of 4 warnings that "should be disabled globally", however, these warnings only ever appear to be disabled when building via cmake, which you obviously don't use when utilizing the single header feature. At least for my compiler and a minimal project this leads to warning C4514 being thrown around a whole bunch.

I'm not exactly sure on how this should be adressed, which is why I raised this issue instead of making a PR, I'd imagine just adding the 4 warnings at the top of the all-encompasing warning push should be enough. If this is indeed the case I'll happily craft up a PR if this isn't considered too trivial.

### Steps to reproduce
Create a new project, set the warning level to /Wall, include the `doctest.h` header in a source file and compile.

### Extra information
* doctest version: 68fab34c (dev head at the time of posting)
* Operating System: Windows 10 Pro 19043
* Compiler+version: MSVC 22 17.0.2

Saalvage:
## Description
Suppress the warnings that are to be suppressed globally in the header itself as well. Also clean up the suppressions by merging common ones into a seperate macro.
I've taken the liberty of removing the comments since they were cluttering the macro a lot and were only present for half of the warning numbers anyways.
Of course if there's anything I should change I'll happily modify. ^^

## GitHub Issues
#565

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -170,30 +170,30 @@
                                                                                                    \
     DOCTEST_MSVC_SUPPRESS_WARNING_PUSH                                                             \
     /* these 4 also disabled globally via cmake: */                                                \
-    DOCTEST_MSVC_SUPPRESS_WARNING(4514)                                                            \
-    DOCTEST_MSVC_SUPPRESS_WARNING(4571)                                                            \
-    DOCTEST_MSVC_SUPPRESS_WARNING(4710)                                                            \
-    DOCTEST_MSVC_SUPPRESS_WARNING(4711)                                                            \
+    DOCTEST_MSVC_SUPPRESS_WARNING(4514) /* unreferenced inline function has been removed */        \
+    DOCTEST_MSVC_SUPPRESS_WARNING(4571) /* SEH related */                                          \
+    DOCTEST_MSVC_SUPPRESS_WARNING(4710) /* function not inlined */                                 \
+    DOCTEST_MSVC_SUPPRESS_WARNING(4711) /* function selected for inline expansion*/                \
     /* */                                                                                          \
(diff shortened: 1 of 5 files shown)


doctest/doctest.h
```diff
@@ -137,67 +145,74 @@
...
+    DOCTEST_MSVC_SUPPRESS_WARNING(4512)                                                            \
+    DOCTEST_MSVC_SUPPRESS_WARNING(4127)                                                            \
+    DOCTEST_MSVC_SUPPRESS_WARNING(4820)                                                            \
+    DOCTEST_MSVC_SUPPRESS_WARNING(4626)                                                            \
+    DOCTEST_MSVC_SUPPRESS_WARNING(5026)                                                            \
+    DOCTEST_MSVC_SUPPRESS_WARNING(5027)                                                            \
+    DOCTEST_MSVC_SUPPRESS_WARNING(4625)                                                            \
+    DOCTEST_MSVC_SUPPRESS_WARNING(4640)                                                            \
+    DOCTEST_MSVC_SUPPRESS_WARNING(5045)                                                            \
+    /* static analysis */                                                                          \
```
onqtam, line 191: I think all MSVC warnings could keep their comments for what they are for with such comments in the multiline macro - I think it's pretty useful for future reference and would avoid needing to google numbers

onqtam:
1 more thing to change would be to remove the following from `update_stuff.py`:
```
    elif line.startswith("#define DOCTEST_VERSION_STR "):
        doctest_contents += "#define DOCTEST_VERSION_STR \"" + version + "\"\n"
```

onqtam:
LGTM!

Issue #567, opened 2021-12-11
`g_oss` is causing incorrect stringification results


Saalvage:
# Edit: I have now published a possible fix at #569!

I'm currently working on a stringification utility for STL classes, and I've discovered an obscure bug that manifests in two slightly different ways:
### 1.
When an argument to a `MessageBuilder` is stringified via being written to an `ostream` (`operator<<`) the result of that stringification is accidentally prepended to the result of the `MessageBuilder`.
**Example:**
```cpp
struct A { }; struct B { };
static std::ostream& operator<<(std::ostream& o, const A&) { o << "A"; return o; }
template<> struct doctest::StringMaker<B> {
	static String convert(const B&) { return toString(A()) + "B"; }
};
TEST_CASE("") { FAIL(B()); }
```
Expected result: `AB`
Actual result: `AAB`

This can be avoided by first writing the `String` into an object like so:
```cpp
doctest::String s = doctest::toString(B());
FAIL(s);
```

### 2.
When an `operator<<` utilizes `toString` which in turn executes its function via writing to a stream.
Arguably this is a bit contrived, since you usually wouldn't use `toString` in an `operator<<` overload, but it still shows a weakness in design, which demonstrates that simply giving `MessageBuilders`s their own thread local osstream won't solve this problem entirely.
There might be a less contrived instance of this bug lurking somewhere!
**Example:**
```cpp
struct A { }; struct B { };
static std::ostream& operator<<(std::ostream& o, const A&) { o << "A"; return o; }
static std::ostream& operator<<(std::ostream& o, const B&) { o << "C" << "B" + doctest::toString(A()); return o; }
TEST_CASE("") { String s = toString(B()); FAIL(s); /* Failure is not due to MessageBuilder! */ }
```
Expected result: `CBA`
Actual result: `ACA`

The stream that is used to stringify `B` needs to be used again to stringify `A`, this messes things up.

I'm honestly not entirely sure on how to solve this, I think the idea of having a single string buffer will need to be abandoned.
My first approach was simply moving the logic of resetting the buffer into the result function, this solves 1. but modifies 2.'s result to be `BCA`, which is still incorrect!

My current progress can be found on https://github.com/Saalvage/doctest/tree/tlss
Some input would be appreciated!

### Extra information
* doctest version: a384864 (dev head at the time of posting)
* Operating System: Windows 10 Pro 19043
* Compiler+version: MSVC 22 17.0.2

onqtam:
[...]

Issue #568, opened 2021-12-12
Compilation errors on MSVC 2015 after doctest update to 2.4.7


rmisev:
## Description

Code below doesn't compile on MSVC 2015 when doctest 2.4.7 is used (but compiles fine when doctest is 2.4.6).
Compiler output:
```
d:\doctest_bug\doctest.h(821): error C2440: 'return': cannot convert from 'const char [4]' to 'const char (&&)[4]'
d:\doctest_bug\doctest.h(821): note: You cannot bind an lvalue to an rvalue reference
d:\doctest_bug\doctest.h(1320): note: see reference to function template instantiation 'T (&&doctest::detail::forward<const R>(const char (&)[4]) noexcept)' being compiled
        with
        [
            T=char [4],
            R=char [4]
        ]
d:\doctest_bug\doctest_bug.cpp(7): note: see reference to function template instantiation 'doctest::detail::Result doctest::detail::Expression_lhs<const std::string &>::operator ==<char[4],0x0>(const R (&))' being compiled
        with
        [
            R=char [4]
        ]
d:\doctest_bug\doctest_bug.cpp(7): note: see reference to function template instantiation 'doctest::detail::Result doctest::detail::Expression_lhs<const std::string &>::operator ==<char[4],0x0>(const R (&))' being compiled
        with
        [
            R=char [4]
        ]
```

### Steps to reproduce

1. Create file doctest_bug.cpp with following content:
~~~c++
#define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
#include "doctest.h"
#include <string>

TEST_CASE("string comparison") {
    std::string str = "ABC";
    CHECK(str == "ABC");
}
~~~

2. Compile it with MSVC 2015 and Doctest 2.4.7

### Extra information

* doctest version: **2.4.7**
* Operating System: **Windows 10**
* Compiler+version: **MSVC 2015**

onqtam:
Turns out the latest release was without MSVC 2015 testing due to the migration to a github org & relying solely on github actions - re-enabling appveyor now...

Just committed a test like your snippet and holy cow:
https://github.com/doctest/doctest/runs/4509284329?check_suite_focus=true

seems that GCC & new versions of MSVC handle it properly, but clang & MSVC 2015 error on it, and in different ways.
- MSVC cannot compile
- clang constructs the stringified version incorrectly (for `CHECK(str == "ABC2");`) which might be related to #567 :
```
-  values: CHECK( ABC == ABC2 )
[...]

PR #569, opened 2021-12-12
Fix the thread-local string-stream

Saalvage:
## Description
I have reimplemented the thread-local string stream as a stack-like class.

One potentially remaining issue is the following:
`MessageBuilder` `a` gets created, it causes a push to the stack. `MessageBuilder` `b` gets created, it causes a push to the stack.
`a` gets `log`ed, causing a pop to the stack. However, this pop incorrectly pops the contents of `b`, causing `b` to pop `a`'s content from the stack. This could be solved by either: 1. simply be ignored, considering that this likely isn't possible when using the library correctly; 2. introducing a thread-local/static variable which dictates that only 1 `MessageBuilder` may exist at one time; 3. giving `MessageBuilder`s a stringstream member that they own, although considering the tlss was added in the first place suggests there's something that speaks against this.

I've also taken the liberty of adjusting/refactoring a few minor aspects of related classes, which I hope is all well and good.
As always, if there's anything I could improve on, don't fret to let me know!

## GitHub Issues
#567


doctest/parts/doctest_fwd.h
```diff
@@ -862,10 +870,8 @@ namespace detail {
     template<class T>
     using has_insertion_operator = has_insertion_operator_impl::check<const T>;
 
-    DOCTEST_INTERFACE void my_memcpy(void* dest, const void* src, unsigned num);
-
-    DOCTEST_INTERFACE std::ostream* getTlsOss(bool reset=true); // returns a thread-local ostringstream
-    DOCTEST_INTERFACE String getTlsOssResult();
+    DOCTEST_INTERFACE std::ostream& tlssPush();
```
onqtam, line 873: I still haven't processed the entire changeset, but 1 quick comment would be that some MSVC version couldn't compile when a reference was used instead of a pointer (in case you were wondering), even though the reference was just passed around and no method was being called... But the builds seem to pass so I guess it's time to switch back to a ref.

onqtam: Actually MSVC 2015 doesn't seem to be included in the github actions build matrix - it was being tested on appveyor until that worked but I didn't set it up properly now that doctest moved to an org... so the latest version was released without testing 2015...

Saalvage: Oh damn, that really sounds suboptimal. So I think the best course of action for this PR is to wait for the dev branch to be fixed and then rerun the builds on this?

onqtam: yeah... also this issue:
https://github.com/doctest/doctest/issues/568

I'll be looking closer at commit 02af672f0d3093157daecb55ce4e109ffaf2bc53

onqtam:
I'll be away for the next 4 days but I'll come back to this after that

onqtam:
This seems good! Regarding the use of nested `MessageBuilder`s - let's ignore that.

Unfortunately VS 2015 fails (not with compiler errors but with different runtime behavior around the stringifications): https://ci.appveyor.com/project/onqtam/doctest-6cp6x/builds/41915462
Example failure:
```
         <Message type="WARNING" filename="subcases.cpp" line="0">
-          <Text>
-            print me twice
-          </Text>
+          <Text/>
         </Message>
```

Issue #571, opened 2021-12-28
Regression between 2.4.6 and 2.4.7


YarikTH:
## Description

Following minimized code works perfectly on 2.4.6, but fails on 2.4.7

```c++
// #define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN shouldn't be defined in compilation unit to reproduce
#include <doctest/doctest.h>

#include <string>

namespace {

class B
{
public:
    const std::string& getName() const
    {
        return m_name;
    }

    std::string m_name = "B";
};

}  // namespace

TEST_CASE("Wtf?") {
    B b;

    CHECK(b.getName() == "B");
}
```
https://godbolt.org/z/Eq48fjsed (due to the godbolt limitations see "undefined reference to `main'" linker warnings as success)

Flags: "-std=c++11 -Wall -Wextra -Wpedantic -Werror"

It fails on GCC version 4.9 - 10.3 with -fpermissive warning. (GCC version 11.2 compiles it without a problem though 🤔)

```
doctest/doctest.h: In instantiation of 'void doctest::detail::fillstream(const T (&)[N]) [with T = char; long unsigned int N = 2]':
doctest/doctest.h:904:31:   required from 'static void doctest::detail::filldata<T [N]>::fill(const T (&)[N]) [with T = const char; long unsigned int N = 2]'
doctest/doctest.h:918:65:   required from 'void doctest::detail::filloss(const T (&)[N]) [with T = char; long unsigned int N = 2]'
doctest/doctest.h:933:20:   required from 'static doctest::String doctest::detail::StringMakerBase<true>::convert(const T&) [with T = char [2]]'
doctest/doctest.h:978:35:   required from 'doctest::String doctest::toString(const T&) [with T = char [2]; typename doctest::detail::enable_if<(! doctest::detail::is_enum<T>::value), bool>::type <anonymous> = true]'
doctest/doctest.h:1142:45:   required from 'doctest::String doctest::detail::stringifyBinaryExpr(const L&, const char*, const R&) [with L = std::__cxx11::basic_string<char>; R = char [2]]'
[...]

PR #572, opened 2021-12-28
Solve #571 issue

YarikTH:
## GitHub Issues

See #571.

Maybe it would be better to hack doctest self-testing code to reproduce the problem, but I have no idea how to do so.

### DIFF
--- a/doctest/parts/doctest_fwd.h
+++ b/doctest/parts/doctest_fwd.h
@@ -440,6 +440,16 @@ class basic_ostream;
 typedef basic_ostream<char, char_traits<char>> ostream;
 template <class... Types>
 class tuple;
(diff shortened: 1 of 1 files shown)


YarikTH:
Once upon a time, I'll do it correctly...

YarikTH:
Fixed in the dev branch somehow. Thanks @onqtam for zero feedback

onqtam:
Sorry for not replying earlier - I have limited bandwidth and was also unsure how to respond precisely because the issue is no longer present in `dev`.

YarikTH:
> was also unsure how to respond precisely because the issue is no longer present in dev
The fact that it is no longer reproduced in the dev branch it is valuable information by itself. Because it seems that I didn't checked it.

Issue #496, opened 2021-04-11
[Feature request] CHECK could return the value of expression


Lectem:
## Description

It would be nice to have `CHECK` (or a new macro) returning the result of the expression.
This is something I've for assertions in quite a few code bases where I have a "VERIFY" assert that is evaluated no matter what and can participate in the flow of the program.
In the case of doctest, simlarly to `VERIFY` vs `ASSERT`, CHECK does not stop the execution (while REQUIRE does) I think it would make sense that `CHECK` also returns the evaluation of the condition so that we can write code such as this:

```cpp
if(CHECK(somePtr != nullptr))
{
     CHECK(somePtr->someMember == someValue);
}
```

So that one can avoid this kind of code:
```cpp
CHECK(somePtr != nullptr);
if(somePtr != nullptr)
{
     CHECK(somePtr->someMember == someValue);
}
```

This is very helpful when dealing with optional data when writing a test that can handle different inputs with the same code.

Is this something you would be interesting in supporting ?

onqtam:
This will be quite hard to achieve (if possible at all) while maintaining the expression decomposing semantics of the `CHECK` macro. Maybe an easy solution is to have a global (thread local) variable set by the framework which can then be queried - something like `if(doctest::LastAssertFailed)`

onqtam:
This was just implemented by @Saalvage & merged in `dev`! Closing this issue.

PR #574, opened 2022-01-04
Assertions returning booleans

Saalvage:
## Considerations
- This should be benchmarked. (I don't know how to!) If using lambdas over `do while (false)` causes performance degradation this feature should be toggleable via config macro.
- Maybe note this in the documentation somewhere?

## GitHub Issues
#496


onqtam:
This is great!

I just measured it by compiling a .cpp file with `time` using g++ 10.3:

- baseline is without any asserts & test cases
- `with asserts` is 250 test cases with 20 asserts in each comparing 2 integers with a CHECK:

This is without this PR:

| *old*        | `-O0`  | `-O3`   |
|--------------|--------|---------|
| baseline     | 1,668s | 3,634s  |
| with asserts | 7,663s | 52,086s |

so 6 seconds for the asserts with `-O0` and 48.5 with `-O3`.

And this is with this PR:

| *new*        | `-O0`   | `-O3`   |
|--------------|---------|---------|
| baseline     | 1,627s  | 3,614s  |
| with asserts | 12,263s | 59,061s |

so 10.5 seconds for the asserts with `-O0` and 55.5 with `-O3`.

So the slowdown for 5k asserts for `-O0` is about 75% and 14% for `-O3` which seems very acceptable to me!

Tradias:
This breaks usage of the `co_await` operator within CHECK macros:

https://godbolt.org/z/xP4vWWse3

Besides, I find the compilation time slowdown quite substantial, given that debug mode is probably the most relevant build type when working with tests.

@Saalvage do you mind adding a config macro like you mention or find a different solution?

Saalvage:
Totally agree, gonna open an issue on this.

examples/all_features/assertion_macros.cpp
```diff
@@ -191,3 +191,16 @@ static void someAssertsInFunction() {
 TEST_CASE("some asserts used in a function called by a test case") {
     someAssertsInFunction();
 }
+
+TEST_CASE("macro return values") {
+    int a = 4;
+    int b = 2;
+    DOCTEST_MSVC_SUPPRESS_WARNING_WITH_PUSH(4702)
+    if (CHECK(a == b)) { MESSAGE("should not be reached!"); }
```
YarikTH, line 199: All the checks return false. Need to duplicate them and cover the case when the return value of CHECK is true. Otherwise possible error when check always returns false is not covered.

Saalvage: Already covered in:
- https://github.com/doctest/doctest/blob/dev/examples/all_features/assert_returns_disabled.cpp
and
- https://github.com/doctest/doctest/blob/dev/examples/all_features/assert_returns_disabled_evaluate.cpp

PR #575, opened 2022-01-04
Alternative approach to Windows color initialization

Saalvage:
Another approach to suppressing the unused variable warning is simply removing the unusued variable!

Of course the possibility exists that this introduced a different warning with Intel's compiler, so could @isaevil be so kind to test this branch and report back? ^^

## GitHub Issues
Alternative to #559

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -4015,29 +4015,6 @@ namespace {
         return suiteOrderComparator(lhs, rhs);
     }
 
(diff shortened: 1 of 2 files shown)


onqtam:
This seems like a good improvement! Will merge it.

However, his PR also adds volatile in the `DOCTEST_GLOBAL_NO_WARNINGS` macro which is used in every test case so #559 is not completely solved.

PR #576, opened 2022-01-05
Fix unused variable 2

Saalvage:
For real this time!

Same approach as last time, suppress the unused variable warning by using the unusued variable.

I've benchmarking using clang but could detect virtually no difference.

## GitHub issues
https://github.com/doctest/doctest/pull/575
I've tested it using the Intel compiler and it correctly suppresses the warning!


onqtam:
This is cool! I wonder when some compiler would start complaining in Release that the variable is still never used :D

Issue #577, opened 2022-01-07
TEST_CASE_TEMPLATE causes "-Wunused-local-typedef" warning on Clang


helmesjo:
## Description
Building with `-Wall` on Clang (tested on 13.0) causes warning `-Wunused-local-typedef` to trigger when a `TEST_CASE_TEMPLATE_INVOKE` is placed inside a `SCENARIO` scope:
```c++
TEST_CASE_TEMPLATE_DEFINE("test", type_t, test_template)
{
}

SCENARIO("scenario")
{
    TEST_CASE_TEMPLATE_INVOKE(test_template, int); // Warning: `-Wunused-local-typedef`
}
TEST_CASE_TEMPLATE_INVOKE(test_template, int); // No warning
```
```
<source>:10:5: error: unused typedef 'DOCTEST_ANON_FOR_SEMICOLON_12' [-Werror,-Wunused-local-typedef]
    TEST_CASE_TEMPLATE_INVOKE(test_template, int);
    ^
/opt/compiler-explorer/libs/doctest/trunk/doctest/doctest.h:2647:44: note: expanded from macro 'TEST_CASE_TEMPLATE_INVOKE'
#define TEST_CASE_TEMPLATE_INVOKE(id, ...) DOCTEST_TEST_CASE_TEMPLATE_INVOKE(id, __VA_ARGS__)
                                           ^
/opt/compiler-explorer/libs/doctest/trunk/doctest/doctest.h:2052:17: note: expanded from macro 'DOCTEST_TEST_CASE_TEMPLATE_INVOKE'
    typedef int DOCTEST_ANONYMOUS(DOCTEST_ANON_FOR_SEMICOLON_)
                ^
/opt/compiler-explorer/libs/doctest/trunk/doctest/doctest.h:347:30: note: expanded from macro 'DOCTEST_ANONYMOUS'
#define DOCTEST_ANONYMOUS(x) DOCTEST_CAT(x, __COUNTER__)
                             ^
/opt/compiler-explorer/libs/doctest/trunk/doctest/doctest.h:345:29: note: expanded from macro 'DOCTEST_CAT'
#define DOCTEST_CAT(s1, s2) DOCTEST_CAT_IMPL(s1, s2)
                            ^
/opt/compiler-explorer/libs/doctest/trunk/doctest/doctest.h:344:34: note: expanded from macro 'DOCTEST_CAT_IMPL'
#define DOCTEST_CAT_IMPL(s1, s2) s1##s2
                                 ^
[...]

PR #578, opened 2022-01-08
Fix semicolon enforcement

Saalvage:
Woo, C++11!

## More-or-Less-Fun Fact
I would like to draw attention to the peculiarity that `DOCTEST_TEST_SUITE_END` poses. It is usable both as a function-like macro and as a regular macro. This is because in order to enforce the macro invocation be followed by a semicolon it has a typedef of the following form at the end: `typedef int DOCTEST_INTERNAL_STH`. Now, this is just adds a simple alias for int, doesn't it? (It does!)
But now watch what happens when we append parentheses: `typedef int DOCTEST_INTERNAL_STH()`. Now, I wouldn't blame anyone for believing this to be invalid syntax, but it actually introduces `DOCTEST_INTERNAL_STH` as an alias, for a ***function*** returning int and taking no parameters! Now isn't that crazy?

## TLDR
All macros that use the existing form of semicolon enforcement outside of functions may have their invocation followed by a set of parantheses inside which type parameters can be placed.
`DOCTEST_TEST_SUITE_END` is a special case because it is the only affected macro which is itself not function-like. This means backwards compatibility with both invocations should preferably be upheld for it, and it alone (who would ever do `DOCTEST_TEST_SUITE_BEGIN(...)()`?). Maintaining the old syntax in this case should not be a big deal either, as I don't think a test suite can ever be ended locally. 

## Related issue
#577

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -2131,10 +2131,10 @@ int registerReporter(const char* name, int priority, bool isReporter) {
     static_assert(true, "")
 
 // for ending a testsuite block
-#define DOCTEST_TEST_SUITE_END()                                                                   \
+#define DOCTEST_TEST_SUITE_END                                                                     \
     DOCTEST_GLOBAL_NO_WARNINGS(DOCTEST_ANONYMOUS(DOCTEST_ANON_VAR_),                               \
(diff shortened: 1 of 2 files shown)


onqtam:
In the future some compiler will complain that this static assert is unnecessary :P

PR #580, opened 2022-01-09
Continuous Integration Refactor

Saalvage:
Getting rid of AppVeyor. (Damn thing sometimes took 4 hours to build!!)

## Considerations
- [x] AppVeyor is still present in the PR tests here, might need to be disabled manually?
- [x] Potentially reinvestigate GCC on Windows?
- [x] Potentially reinvestigate [x86 MinGW](https://github.com/doctest/doctest/runs/4750696358?check_suite_focus=true) (only some tests fail)?
- [x] Potentially add more intermediate Xcode versions or macOS gcc builds?
- [x] Current .yml contained commented-out references to openMPI, might be worth reinvestigating?
- [ ] It might be worth looking into utilizing Ccache to cut down on runtimes?
- [x] Clang x86 builds depend on gcc-multilib being present, is there no clang alternative to that?
- [x] Due to the nature of testing the CI this branch/PR contains an ungodly amount of commits. It seems like PRs get squashed before merging, but just incase it might still pose a concern I'm mentioning it here

These points should probably be adressed/converted to issues before merging.


Saalvage:
Can't get x86 clang-cl using Ninja to work. Really don't wanna fall back on scoop. Honestly think we have more than enough redundancy to just ignore this.

onqtam:
This stuff is great!

> * [ ]  AppVeyor is still present in the PR tests here, might need to be disabled manually?

I just removed it.

> * [ ]  Potentially reinvestigate GCC on Windows?

Isn't mingw precisely that? Not sure there's another way of using GCC on Windows.

> * [ ]  Potentially reinvestigate [x86 MinGW](https://github.com/doctest/doctest/runs/4750696358?check_suite_focus=true) (only some tests fail)?

I think the x64 mingw builds are enough - I doubt the 32 bit builds fail because of doctest - let's not spend any more time on obscure toolchain issues.

> * [ ]  Potentially add more intermediate Xcode versions or macOS gcc builds?

Nah, this is fine :)

> * [ ]  Current .yml contained commented-out references to openMPI, might be worth reinvestigating?

@BerengerBerthoul @philbucher might want to add tests/builds for it, but this PR is good to go in that regard.

> * [ ]  It might be worth looking into utilizing Ccache to cut down on runtimes?

I had integrated it previously on travis (don't remember the gains), it could be added for github actions too - I did it once for a GitHub project a year ago: https://github.com/vectorizedio/redpanda/pull/141

> * [ ]  Clang x86 builds depend on gcc-multilib being present, is there no clang alternative to that?

Not sure - doesn't bother me.

> * [ ]  Due to the nature of testing the CI this branch/PR contains an ungodly amount of commits. It seems like PRs get squashed before merging, but just incase it might still pose a concern I'm mentioning it here

I always squash commits when merging PRs, no worries!

One issue is that currently, Valgrind builds are running only for x86 (judging by the `run_test(configuration, "VALGRIND", x86_flag)` line) - previously they were run for both x64 and x86. This can be addressed in a follow-up PR.

Btw what is the difference between `clang` & `clang-cl` on Windows?

Saalvage:
> Isn't mingw precisely that? Not sure there's another way of using GCC on Windows.

There were commented out builds using GCC with Ninja on Windows (they compiled but the tests failed!)
I assume they're probably redundant with MinGW though.

> I had integrated it previously on travis (don't remember the gains), it could be added for github actions too - I did it once for a GitHub project a year ago: vectorizedio/redpanda#141

Gonna investigate!

> Btw what is the difference between clang & clang-cl on Windows?

According to the Ninja CMake console output it's "Clang 13.0.0 with GNU-like command-line" as opposed to "Clang 13.0.0 with MSVC-like command-line", whatever that means. [Probably a good source if you actually care.](https://llvm.org/devmtg/2014-04/PDFs/Talks/clang-cl.pdf) (I have to admit, I don't really.)

PR #584, opened 2022-01-12
Feature: Better NaN

Saalvage:
No longer separate macros!

### DIFF
--- a/doctest/parts/doctest_fwd.h
+++ b/doctest/parts/doctest_fwd.h
@@ -697,17 +697,6 @@ DOCTEST_INTERFACE const char* assertString(assertType::Enum at);
 DOCTEST_INTERFACE const char* failureString(assertType::Enum at);
 DOCTEST_INTERFACE const char* skipPathFromFilename(const char* file);
 
-template <typename F>
-struct IsNaN
-{
-    F val;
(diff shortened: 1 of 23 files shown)


YarikTH:
It looks like a breaking change. It should be added in addition to the old macroses if the change is planned for minor version, or it should be postponed for the next major release.

But change itself looks cool. Better result with a wat less code.

Saalvage:
The previous NaN macros are (thankfully) not in any stable release yet (only on the dev branch) so having a breaking change should be fine. Thank you for the review!

doctest/parts/doctest_fwd.h
```diff
@@ -1131,6 +1130,28 @@ DOCTEST_INTERFACE String toString(const Approx& in);
 
 DOCTEST_INTERFACE const ContextOptions* getContextOptions();
 
+#if DOCTEST_MSVC
+#define DOCTEST_INTERFACE_DECL
```
onqtam, line 1134: Isn't this the same as what's near line 320? Seems that this is a redundant redefinition.

Saalvage: Yep! Must've missed removing this when moving it to the top

PR #585, opened 2022-01-12
Refactor stringification

### LINKED ISSUE #571: Regression between 2.4.6 and 2.4.7

Saalvage:
So `StringMaker` and `StringStream` act as a pretty much completely parallel structure. This structure will be eradicated.
For the first draft I have simply removed the `StringMaker`'s body, attaching it's head to `StringStream`'s body.

The customization points are broken, as well as a few other features, namely raw memory stringification likely among others.

This (hopefully) finally fixes #571 for good!


YarikTH:
Some thoughts about the situation with stringification. Imho first of all, missing unit tests should be added. Either tests for full output with -s (covering all stringification cases) or for output of toString functions for unknown types and for supported types. If doctest have it already, then the whole issue with stringification would non exist.

Also, at least such tests (if not all of them) should be added several times in different configuration matrix:
* main function in the separate cpp / main function  in the same cpp as tests
* with and without DOCTEST_CONFIG_USE_STD_HEADERS
* other similar options that can break the things

Valgrind and sanitiser configuration should be added on CI if they are not. At least they would have detected the overflow problem if they existed. Or doctest just lack of tests that fail under them.

I think that at least part of the problem is
```c++
    // cppcheck-suppress noExplicitConstructor
    String(const char* in);
```
It's a landmine waiting to explode. And it explodes in 2.4.7 and 2.4.8.

Saalvage:
Totally agreed, it's quite ironic that a unit testing library does not utilize unit tests to test itself!

Valgrind is already being used and [sanitizers already managed to catch an issue when the appropriate tests are used](https://github.com/doctest/doctest/runs/4796631949?check_suite_focus=true)!

From the many stringification issues I don't think that that String constructor is one, I don't even think String constructors are at all considered on stringification, although I'm not sure.

doctest/parts/doctest_fwd.h
```diff
@@ -1120,13 +1039,86 @@ class DOCTEST_INTERFACE Approx
...
+    struct StringStreamBase
+    {
+        template <typename T>
+        static void convert(std::ostream* s, const T&) {
+            writeChars(s, "{?}");
+        }
+
+        // always treat char* as a string in this context - no matter
+        // if DOCTEST_CONFIG_TREAT_CHAR_STAR_AS_STRING is defined
+        static void convert(std::ostream* s, const char* in) { *s << in; }
```
Saalvage, line 1058: Is this used at this point?

Saalvage:
TODO:
- [ ] More regression tests
- [ ] See if we can somehow get `toStream` working in MessageBuilder
- [ ] Test treating char star as string

Saalvage:
Here's a diagram of the current stringification process, green representing customization points.
diagram

YarikTH:
This branch has crashed in case of pointer stringification (invalid pointer actually).
https://godbolt.org/z/jW7sbj6T7

Previous releases work fine
https://godbolt.org/z/1dKozvq7T

YarikTH:
Also, my stringification test points me to some improvements:
```
Fundamental types:
    float:                          1.1f
    double:                         2.2
    long double:                    3.3
С arrays:
    const char[]:                   foo
    char[]:                         bar
    const int[]:                    123
    int[]:                          456
```
=>
```
Fundamental types:
    float:                          1.1f
    double:                         2.2
    long double:                    3.3L
С arrays:
    const char[]:                   foo
    char[]:                         bar
    const int[]:                    [1, 2, 3]
    int[]:                          [4, 5, 6]
```
Great work, I almost reported strange stringification of non-char arrays 👍

YarikTH:
toString ignored overloaded ostream operator for enum class values.
https://godbolt.org/z/sYPEb3Y1h

YarikTH:
Call docstring::toString with pointer types spoils internal tloss. Reproducible in 2.4.8 and in dev, fixed in this PR. Great work!
https://godbolt.org/z/fc4cfMj3K

YarikTH:
I think that define to disable tloss optimization is very desirable. It's unstable feature that causes too much problems. Imho it should be disabled by default and switched on manually if someone really wants it to speed up tests runtime.

YarikTH:
Yet another find. The toString output of wchar_t, char16_t and char32_t depend on the cxx standard. It outputs integer representation until c++17, but in c++20 it becomes unknown for doctest.
https://godbolt.org/z/7axaaoTKx

Tested both on GCC and clang.

Saalvage:
> This branch has crashed in case of pointer stringification (invalid pointer actually). https://godbolt.org/z/jW7sbj6T7
> 
> Previous releases work fine https://godbolt.org/z/1dKozvq7T
> 
> It crashed in
> 
> ```c++
>     template <typename T>
>     String rawMemoryToString(const T& object) {
>         return rawMemoryToString(&object, sizeof(object));
>     }
> ```
> 
> It seems that the presumption that not null pointer points to valid data is principally incorrect.
> 
> My example is kind of syntetic, but for example there is such thing as an end iterator for std::array or C array. It points past the array data so, dereferencing of it, it is an error.
> 
> ```c++
> int a[] = { 1, 2, 3 };
> auto it = std::find( std::begin(a), std::end(a), 2 );
> CHECK( it != std::end(a) ); // both it and end can point at the invalid memory past the buffer
> ```

This is really weird because I didn't really change anything about the stringification of pointers.. Good catch!

YarikTH:
I've made stringify test or more or less sandbox where it can be easily checked
https://github.com/YarikTH/doctest_stringify_test

It uses doctest itself and ApprovalTests.cpp to continuously compare differences in outputs. Test contains all fundamental types, custom classes and most of container and utility std classes. It optionally uses stl_stringifier.h to check how it does its work.

All of the finds were done in the process of developing it.

YarikTH:
About pointer stringification. There is already ostream operator for pointer types. But doctest uses its own approach to their stringification. And this approach is not consistent with stringification of shared_prt and unique_ptr because doctest doesn't handle them)
https://godbolt.org/z/dfxsW5jG4

doctest/doctest.h
```diff
@@ -445,7 +455,6 @@ DOCTEST_GCC_SUPPRESS_WARNING_POP
 #ifndef DOCTEST_CONFIG_INCLUDE_TYPE_TRAITS
 #define DOCTEST_CONFIG_INCLUDE_TYPE_TRAITS
 #endif // DOCTEST_CONFIG_INCLUDE_TYPE_TRAITS
-#include <cmath>
 #include <cstddef>
 #include <ostream>
 #include <istream>
```
YarikTH, line 460: `#include<iosfwd>` might be sufficient here. Because it declares but not defines both ostream and istream and its inclusion times are non-existing comparing to ostream and istream see https://artificial-mind.net/projects/compile-health/
But I'm not sure if DOCTEST_CONFIG_USE_STD_HEADERS is properly tested.

Saalvage:
Honestly the rat's tail of this is unimaginable, I am almost certain there is more weird and broken stuff waiting in the nooks and crannies here, but at some point we'll just have to bite the bullet and go through with it. Can't have people waiting on 2.4.9 forever!

Issue #248, opened 2019-06-24
Vector contents show up as question mark


RPiAwesomeness:
## Description
I recently switched from the Catch2 library due to the improved compile times and everything seems to have gone flawlessly. However, I noticed that when comparing vector values fails, unlike Catch2 which will print the contents of the vectors, I just get this:

```
===============================================================================
test/test-encoding.cpp:73:
TEST CASE:  std::string to_utf16le_bytes(std::u16string)

test/test-encoding.cpp:79: FATAL ERROR: REQUIRE( result == expected ) is NOT correct!
  values: REQUIRE( {?} == {?} )

===============================================================================
```

whereas with Catch2 it provides the contents of the vectors:

```
test/test-encoding.cpp:90: FAILED:
  REQUIRE( result == expected )
with expansion:
  { '�', '2', '�', '2' }
  ==
  { '2', '�', '2', '�' }

===============================================================================
```

I know #170 matches this fairly exactly, but in that case he was dealing with templates and localization was the issue that was resolved, which I don't believe (off the top of my head) should apply in this situation. The contents of the vectors are `uint8_t` (`unsigned char` on most x64 systems), so I don't believe it has to do with the contents.

### Steps to reproduce

Compile the following with `doctest.h` (`g++ test.cpp doctest.h -o vector_test`) and see result:

```c++
// Tell doctest library to create a main for this file - only include this line once!
#define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN

#include <vector>

#include "doctest.h"

using namespace std;

TEST_CASE("Compare vectors")
{
  vector<uint8_t> expected{ 0x32, 0xdb, 0x32, 0x8d };
  vector<uint8_t> result{ 0x32, 0x00, 0x32, 0x00 };

  REQUIRE(expected == result);
}
```

#### Result

```
[doctest] doctest version is "2.3.3"
[doctest] run with "--help" for options
===============================================================================
test/test-test.cpp:10:
TEST CASE:  Compare vectors

test/test-test.cpp:15: FATAL ERROR: REQUIRE( expected == result ) is NOT correct!
  values: REQUIRE( {?} == {?} )

===============================================================================
[doctest] test cases:      1 |      0 passed |      1 failed |      0 skipped
[doctest] assertions:      1 |      0 passed |      1 failed |
[doctest] Status: FAILURE!
```

* doctest version: **v2.3.3**
* Operating System: **Ubuntu 19.04 on Linux 5.0.0-17-generic**
[...]

Issue #192, opened 2019-03-11
std::array doesn't have a default printer.


DenisYaroshevskiy:
Hi.

For std::array comparison I get:
REQUIRE( {?} == {?} )
I believe you support this for other containers. Can you add this when/if you have the time?

onqtam:
@DenisYaroshevskiy This isn't supported for any container from the STL - Catch2 does that. I might provide a second header with extensions (because the main header will never include any STL/whatever header) and provide stringification for the most common containers/STL Types.

A related issue is printing ENUM values which will also go into such an extension header - see this: **https://github.com/onqtam/doctest/issues/121**

DenisYaroshevskiy:
@onqtam Oh - I see.

You do rely on ADL though for looking up types <=> for std ones it's not nice to do so.
Can I specialize smth instead?

* How about for generic <is this a range> test? I don't know what your minimal language requirements, but if you do it based on having nested iterator type, you should be OK even for 98.
* What if 1 million? Good point. Well - dump all is a very decent default if you could override it.

onqtam:
C++11 is the requirement for doctest 2.x (no longer C++98).

I think [this is how Catch2 detects ranges](https://github.com/catchorg/Catch2/blob/16dc2197046dd1362e26027e789f15d65275a185/include/internal/catch_tostring.h#L472).

I might implement matchers at some point and in that case when comparing containers you would get automatically the index that points to the different elements - this would be very hard to do with the current stringification of the values.

But currently I'm focusing on finishing the XML reporter.

Issue #170, opened 2018-12-20
Print vector content.


slepasteur:
## Description
I am trying to print std::vector content on failure.
E.g.:
`REQUIRE(std::vector<int>{1} == std::vector<int>{1,2});`

Should produce to following output.
```
FATAL ERROR: REQUIRE( std::vector<int>{1} == std::vector<int>{1,2} ) is NOT correct!
  values: REQUIRE( [ 1 ] == [ 1 2 ] )
```
Here is my attempt:
```c++
namespace doctest {
template <typename T>
struct StringMaker<std::vector<T>>
{
    static String convert(const std::vector<T>& vec)
    {
        std::ostringstream oss;
        oss << "[ ";
        std::copy(vec.begin(), vec.end(), std::ostream_iterator<T>(oss, " "));
        oss << "]";
        return oss.str().c_str();
    }
};
} // namespace doctest
```
It works well at first but as soon as I add more translation units with more tests it starts failing and printing the default way:
```
FATAL ERROR: REQUIRE( std::vector<int>{1} == std::vector<int>{1,2} ) is NOT correct!
  values: REQUIRE( {?} == {?} )
```

### Steps to reproduce
It is kind of hard to reproduce as the output is shown only on failure and it keeps failing after some time.

### Extra information
* doctest version: **v2.0.1**
* Operating System: **Archlinux**
* Compiler+version: **clang 7.0.0 / gcc 8.2.1**

onqtam:
Is the specialization of ```StringMaker<std::vector<T>>``` visible in all translation units? It should be - not just in one of them. Perhaps you can make a proxy header for doctest called ```doctest_proxy.h``` and have the following:

```c++
#pragma once
#include <doctest.h>

// StringMaker<std::vector<T>>
```

Perhaps I should edit the documentation if this is the problem (the specialization being present in only one of the translation units) - this might be due to ODR.

slepasteur:
That was indeed the problem. Thank-you.

u3shit:
Should I suggest some documentation changes wrt this issue? I've spent numerous hours trying to figure out why my custom stringification functions seem to randomly work in some files but not in others (well, I learned that you can do ODR violations with function overloads and template specializations, so it wasn't a completely wasted time, but well...).
[...]

PR #587, opened 2022-01-13
Support pretty printing of container based on heuristics

YarikTH:
## Description

There are a lot of complaints about the fact, that vector values and other containers output as "{?}" because their stringification is not supported.

I think that it should not be much hard to support standard containers (or at least most of them) without including nothing but type_traits.

```c++
template<typename T, typename = void> struct has_begin_end_overloads : std::false_type {};
template<typename T>                  struct has_begin_end_overloads<T, std::void_t<
    decltype( std::begin( std::declval<T>() ) != std::end( std::declval<T>() ) )
    >> : std::true_type {};
template<typename T> constexpr bool has_begin_end_overloads_v = has_begin_end_overloads<T>::value;

template<typename T, typename = void> struct has_mapped_type : std::false_type {};
template<typename T>                  struct has_mapped_type<T, std::void_t<
    typename T::mapped_type
    >> : std::true_type {};
template<typename T> constexpr bool has_mapped_type_v = has_mapped_type<T>::value;

template<typename T> constexpr bool is_element_container_v = has_begin_end_overloads_v<remove_all_t<T>> && !has_mapped_type_v<remove_all_t<T>>;
// output as [X, Y, Z]
// detects vector, list, deque, set etc. works even with old good C arrays

template<typename T> constexpr bool is_key_value_container_v = has_begin_end_overloads_v<remove_all_t<T>> && has_mapped_type_v<remove_all_t<T>>;
// output as {X: XX, Y: YY, Z: ZZ}
// works with map and unordered_map
```

I have written such code for my closed sourced project and it took less than 300 LOC to support stringification of pairs, tuples, pointers, etc. Unfortunately, it is ported to C++17 but with some effort, code can be adapted to C++11.

I'm sure that this stringification would be a great optional feature (enabled by default or not) and doesn't bloat the code at all. But benchmarks should measure it better as soon there is something to measure.

I often use doctest here and there and I'm bored to copy-paste vertor stringification again and again and again. I prefer it to be in the doctest, even in a disabled state by default.

## GitHub Issues

#121, #170, #192, #248

### DIFF
--- a/scripts/playground/CMakeLists.txt
+++ b/scripts/playground/CMakeLists.txt
@@ -5,3 +5,4 @@ doctest_add_test(NAME playground NO_OUTPUT COMMAND $<TARGET_FILE:playground> -nv
 
 #add_custom_command(TARGET playground POST_BUILD COMMAND $<TARGET_FILE:${PROJECT_NAME}>)
 #add_custom_command(TARGET playground POST_BUILD COMMAND ctest -C $<CONFIGURATION> --output-on-failure)
+


YarikTH:
Looked through https://github.com/Saalvage/doctest/blob/dev/doctest/extensions/stl_stringifier.h

Looks like a lot of well-adjusted work to process all the possible things std throws at you. A generalized approach, unfortunately, couldn't cover all the listed classes, but I think it covers at least half of them. Others are more rarely used anyway.

Saalvage:
`DOCTEST_CONFIG_INCLUDE_TYPE_TRAITS` is already a config option!

YarikTH:
Waiting for #585. There is no point in adding additional stringification, while the current one is broken a little.

Some little preparation is done in godbolt https://godbolt.org/z/bKhsfcxns

elad-eyal-ns:
#585 is merged - this ticket should be re-opened?

PR #589, opened 2022-01-14
Feature config ret vals

### LINKED ISSUE #588: DOCTEST_CONFIG_ASSERT_RETURN_VALUES
#574
> This breaks usage of the `co_await` operator within CHECK macros:
> 
> https://godbolt.org/z/xP4vWWse3
> 
> Besides, I find the compilation time slowdown quite substantial, given that debug mode is probably the most relevant build type when working with tests.

Will probably have to implement `DOCTEST_FUNC_SCOPE_BEGIN` `DOCTEST_FUNC_SCOPE_END` that either use lambdas or do while.
Then don't actually return when using do while and change the assert macros to empty when doctest is disabled.

I'm wondering whether having the return values enabled by default was the right decision. It's a feature you have to actively utilize and it has a negative impact on performance. I'm leaning towards disabling it by default, but would like to hear some more opinions.

Saalvage:
Closes #588.

Also fixes a bug where disabled but evaluated exception related lambdas don't capture and added regression tests for it.

### DIFF
--- a/doc/markdown/assertions.md
+++ b/doc/markdown/assertions.md
@@ -113,8 +113,6 @@ This is a combination of ```<LEVEL>_THROWS_WITH``` and ```<LEVEL>_THROWS_AS```.
 CHECK_THROWS_WITH_AS(func(), "invalid operation!", std::runtime_error);
 ```
 
-All `THROWS_WITH` macros are unaffected by `DOCTEST_CONFIG_EVALUATE_ASSERTS_EVEN_WHEN_DISABLED`.
-
 - ```<LEVEL>_NOTHROW(expression)```
 
 Expects that no exception is thrown during evaluation of the expression.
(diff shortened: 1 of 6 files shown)


Saalvage:
@Tradias Any additional thoughts?

Tradias:
Works for me. Technically this is a breaking change to the previous release since `check-returns-value` used to be enabled by default. According to the dont-pay-for-what-you-dont-use premise I would be fine with it being disabled by default, seeing the negative impact on compile time that it has. People that have already jumped on the new feature are probably fine with making another change: adding the new preprocessor definition `DOCTEST_CONFIG_ASSERTS_RETURN_VALUES`.

Saalvage:
Well, you could argue the previous change was also breaking when using`co_return` and this just remedies that!

PR #590, opened 2022-01-16
Add tests for default stringification result of doctest

YarikTH:
## Description

Use doctest to self-test itself. Stringification tests cannot be added in the "all_features" trash can, because code from it poisons new tests (namely StringMaker<std::list<T>>).

Also, the approach of other tests is not quite applicable here, because they are actually approval tests that work on program output level. I was really surprised when I added my test that should fall, with NO_OUTPUT
```cmake
doctest_add_test(NO_OUTPUT NAME stringification_all COMMAND $<TARGET_FILE:stringification>)
```
but it wasn't and I saw such output in the console:
```
133: [doctest] doctest version is "2.4.8"
133: [doctest] run with "--help" for options
133: ===============================================================================
133: main.cpp(0):
133: TEST CASE:  C_arrays
133: 
133: main.cpp(0): ERROR: CHECK_EQ( toString(constCharArray), "foo" ) is NOT correct!
133:   values: CHECK_EQ( foo╞ , foo )
133: 
133: main.cpp(0): ERROR: CHECK_EQ( toString(constIntArray), "[1, 2, 3]" ) is NOT correct!
133:   values: CHECK_EQ( 123, [1, 2, 3] )
133: 
133: main.cpp(0): ERROR: CHECK_EQ( toString(intArray), "[4, 5, 6]" ) is NOT correct!
133:   values: CHECK_EQ( 456, [4, 5, 6] )
133: 
133: ===============================================================================
133: main.cpp(0):
133: TEST CASE:  Enumerations
133: 
133: main.cpp(0): ERROR: CHECK_EQ( toString(EnumWithOss::E_VALUE_WITH_OSS), "E_VALUE_WITH_OSS" ) is NOT correct!
133:   values: CHECK_EQ( 2, E_VALUE_WITH_OSS )
133: 
133: main.cpp(0): ERROR: CHECK_EQ( toString(EnumClassWithOss::A), "A" ) is NOT correct!
133:   values: CHECK_EQ( 4, A )
133: 
133: ===============================================================================
133: main.cpp(0):
133: TEST CASE:  Structs
133: 
133: main.cpp(0): ERROR: CHECK_EQ( toString(StructOss{777}), "777" ) is NOT correct!
133:   values: CHECK_EQ( 309, 777 )
133: 
133: ===============================================================================
133: [doctest] test cases:  12 |   9 passed | 3 failed | 0 skipped
133: [doctest] assertions: 110 | 104 passed | 6 failed |
133: [doctest] Status: FAILURE!
```

Some tests falls now, but it is ok because currently doctest itself is a little broken and tests just point at this fact.

Based on work commented here https://github.com/doctest/doctest/pull/585.

And on a separate project https://github.com/YarikTH/doctest_stringify_test.

### DIFF
--- a/examples/stringification/CMakeLists.txt
+++ b/examples/stringification/CMakeLists.txt
@@ -1,0 +1,44 @@
+# separated executable to ensure there is no specializations of StringMaker leaked
+# In opposite to all_features/stringification.cpp this test intended to check all the bundled
+# output rules for all the fundamental types, std containers etc
+# it might overlap with other tests, but better to test twice, than miss some cases
+add_executable(stringification main.cpp)
+target_link_libraries(stringification doctest)
+target_compile_definitions(stringification PRIVATE SIZEOF_VOID_P=${CMAKE_SIZEOF_VOID_P})
+if(MSVC)
+    # See https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/
+    target_compile_options(stringification PUBLIC "/Zc:__cplusplus")
+endif()
+
+doctest_add_test_2(NAME stringification_all COMMAND $<TARGET_FILE:stringification>)
+
+# Unfortunately tests could and actually do affect each other, so rerun each test case separately
+doctest_detect_test_cases(main.cpp stringification_test_cases)
+foreach(test_case IN LISTS stringification_test_cases)
+    string(MAKE_C_IDENTIFIER "${test_case}" test_case_identifier)
+    string(TOLOWER "${test_case_identifier}" test_case_identifier)
+    doctest_add_test_2(NAME "stringification_${test_case_identifier}" COMMAND $<TARGET_FILE:stringification> --test-case=${test_case})
+endforeach()
+
+if(MSVC)
+    target_compile_options(stringification PRIVATE /wd4505) # unreferenced local function has been removed
+    target_compile_options(stringification PRIVATE /wd4100) # unreferenced formal parameter
+    target_compile_options(stringification PRIVATE /wd4189) # local variable is initialized but not referenced
+    target_compile_options(stringification PRIVATE /wd4312) # 'reinterpret_cast': conversion from 'unsigned int' to 'int *' of greater size
+    target_compile_options(stringification PRIVATE /wd5245) # '`anonymous-namespace'::operator <<': unreferenced function with internal linkage has been removed
+endif()
+
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+    target_compile_options(stringification PRIVATE -Wno-unknown-warning-option)
+    target_compile_options(stringification PRIVATE -Wno-unneeded-internal-declaration)
+    target_compile_options(stringification PRIVATE -Wno-unused-function)
+    target_compile_options(stringification PRIVATE -Wno-unused-parameter)
+    target_compile_options(stringification PRIVATE -Wno-unused-variable)
+    target_compile_options(stringification PRIVATE -Wno-unused-template)
+elseif(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
+    target_compile_options(stringification PRIVATE -Wno-unused-function)
+    target_compile_options(stringification PRIVATE -Wno-unused-parameter)
+    target_compile_options(stringification PRIVATE -Wno-unused-variable)
+    target_compile_options(stringification PRIVATE -Wno-switch-default)
+    target_compile_options(stringification PRIVATE -Wno-useless-cast) # static_casts used to explicitly set used type
+endif()
--- a/examples/stringification/CMakeLists.txt
+++ b/examples/stringification/CMakeLists.txt
@@ -10,14 +10,17 @@ if(MSVC)
     target_compile_options(stringification PUBLIC "/Zc:__cplusplus")
 endif()
 
-doctest_add_test_2(NAME stringification_all COMMAND $<TARGET_FILE:stringification>)
+# omit the version and the num test cases skipped from the summary - this way the output will change less often
+set(common_args COMMAND $<TARGET_FILE:stringification> --no-skipped-summary --no-version)
+
+doctest_add_test(NAME stringification_all COMMAND ${common_args})
 
 # Unfortunately tests could and actually do affect each other, so rerun each test case separately
 doctest_detect_test_cases(main.cpp stringification_test_cases)
 foreach(test_case IN LISTS stringification_test_cases)
     string(MAKE_C_IDENTIFIER "${test_case}" test_case_identifier)
     string(TOLOWER "${test_case_identifier}" test_case_identifier)
-    doctest_add_test_2(NAME "stringification_${test_case_identifier}" COMMAND $<TARGET_FILE:stringification> --test-case=${test_case})
+    doctest_add_test(NAME "stringification_${test_case_identifier}" COMMAND ${common_args} --test-case=${test_case})
 endforeach()
 
 if(MSVC)
--- a/examples/stringification/main.cpp
+++ b/examples/stringification/main.cpp
@@ -1,0 +1,437 @@
+#define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
+#include <doctest/doctest.h>
+
+DOCTEST_MAKE_STD_HEADERS_CLEAN_FROM_WARNINGS_ON_WALL_BEGIN
+// Pass
+#if __cplusplus >= 202002L
+#   define CXX_STANDARD 20
+#elif __cplusplus >= 201703L
+#   define CXX_STANDARD 17
+#elif __cplusplus >= 201402L
+#   define CXX_STANDARD 14
+#elif __cplusplus >= 201103L
+#   define CXX_STANDARD 11
+#else
+#   define CXX_STANDARD 3
+#endif
+
+#include <ostream>
+
+#include <string>
+#include <utility>
+#include <tuple>
+#include <array>
+#include <valarray>
+#include <initializer_list>
+#include <vector>
+#include <deque>
+#include <list>
+#include <forward_list>
+#include <stack>
+#include <queue>
+#include <set>
+#include <map>
+
+#if CXX_STANDARD >= 14
+#   include <string_view>
+#endif
+
+#if CXX_STANDARD >= 17
+#   include <optional>
+#   include <variant>
+#endif
+
+#if CXX_STANDARD >= 20
+#   include <span>
+#endif
+DOCTEST_MAKE_STD_HEADERS_CLEAN_FROM_WARNINGS_ON_WALL_END
+
+namespace
+{
+
+enum Enum
+{
+    E_VALUE = 1
+};
+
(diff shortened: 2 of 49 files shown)


examples/stringification/main.cpp
```diff
@@ -0,0 +1,432 @@
+#define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
+#include <doctest/doctest.h>
+
+DOCTEST_MAKE_STD_HEADERS_CLEAN_FROM_WARNINGS_ON_WALL_BEGIN
+// Pass
+#if __cplusplus >= 202002L
```
Saalvage, line 6: This is broken on MSVC, it doesn't define `__cplusplus` properly.

Saalvage: See my stringification branch for fix

YarikTH: See CMakeLists.txt. It is already handled
```cmake
if(MSVC)
    # See https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/
    target_compile_options(stringification PUBLIC "/Zc:__cplusplus")
endif()
```

Saalvage: https://github.com/Saalvage/doctest/blob/a9a9c23083111dac60afd7810f18ef9a762f2b36/doctest/doctest.h#L91

YarikTH: It might be not ideal solution for a header-only library, but for test executable it should work just fine

YarikTH:
Need some time to fix all problems on CI. Cursed compiler warnings and compiler zoo. One compiler warns me when I handle default switch for enum, where all enum values are handled, the other warns me about the opposite, how dare I not to have default switch. And in GCC 4.9 "unused-function" warning even is not suppressible inside of the source code.

Saalvage:
I know the feeling!

YarikTH:
Ok, now it's done. Enabling broken test, so all detected problems will be highlighted.

scripts/cmake/common.cmake
```diff
@@ -72,6 +72,50 @@ function(doctest_add_test)
     doctest_add_test_impl(${ARGN} JUNIT_OUTPUT)
 endfunction()
 
+# Unlike doctest_add_test we need to add doctest as a regular test
+# I mean test which output is not checked, but instead we check retcode
```
Saalvage, line 76: Why not just check for no output?

YarikTH: Because NO_OUTPUT mode is either broken or intended for something else.
See https://github.com/doctest/doctest/pull/590#issue-1105153070
I tried to use it, and as a result, I see "Status: FAILURE!" in the console, but CTest entry is completed just fine, no error is emitted. Because it seems that only two types of errors are allowed now:
1. Compilation of test is failed (possibly due to warnings)
2. Executable output differs from approved one (it seems that NO_OUTPUT disables exactly this)
The case when we have doctest based executable and if some of the tests fail, then CTest fails is not an option currently.

YarikTH: Actually, I can accept the game rules and make approval test. For which only 
```
[doctest] doctest version is "2.4.8"
[doctest] run with "--help" for options
[doctest] test cases: 0 | 0 passed | 0 failed | 12 skipped
[doctest] assertions: 0 | 0 passed | 0 failed |
[doctest] Status: SUCCESS!
```
output is allowed. But I don't like such approach.

Saalvage: Imo that approach is preferable since I think cmake should remain minimal.

YarikTH: Ok. I'll try to get it work.

Saalvage:
This should be redirected to the `fix-string` branch. I won't merge it into dev before that is done and over with to make things simpler.

YarikTH:
Meh. Now CI is broken because I have a different amount of test cases depending on standard and compiler. As a result, I either have to add stubs instead of disabled test cases to keep their amount, or have to move disabled tests into other files and executables, or have to add "cxx_standard_<X>" suffix to all test names that depend on the standard.

Silly problem out of nowhere(

YarikTH:
I think I'll move with test case stubs if there are no better and easier solutions.

scripts/cmake/common.cmake
YarikTH: Done. But at what cost? 😭

YarikTH:
Currently is not a big problem, but I can't see set cxx standard on CI. So some cxx standard dependent paths could be not tested.

scripts/cmake/common.cmake
```diff
@@ -72,6 +72,25 @@ function(doctest_add_test)
     doctest_add_test_impl(${ARGN} JUNIT_OUTPUT)
 endfunction()
 
+# In order not to miss subcases, parse a file with regex to find them
+function(doctest_detect_test_cases FILE_NAME OUTPUT_NAME)
```
onqtam, line 76: Not sure if relevant, but just wanted to post about `doctest_discover_tests` here in case you're not aware of it (I haven't used it myself):
https://github.com/doctest/doctest/blob/master/scripts/cmake/doctest.cmake#L108

YarikTH:
I'm aware of it, and considered using it. But I fear it wouldn't work in
this context. At the very least it moves tests compilation from building
time to the configuration time, that I think is not desirable. If these
tests have compilation errors, then cmake configuration is broken and you
can't work with other targets.

пн, 17 янв. 2022 г., 15:45 Viktor Kirilov ***@***.***>:

Saalvage:
Two issues that kind of slow down the reviewing of this:
- I am pretty unfamiliar with CMake and will need to read into it
- From my testing with the stl stringification I'm fairly certain the features C++>11 don't seem to get tested iirc, even if they're available

YarikTH:
> I am pretty unfamiliar with CMake and will need to read into it

There is no complicated logic in CMake code.
- doctest_detect_test_cases performers a naive regexp search of strings inside TEST_CASE macro in the given file
- stringification executable is based on all_features code. It differs in
  -  "/Zc:__cplusplus" part that forces MSVC to pass correct version in the __cplusplus
  - passing SIZEOF_VOID_P, so I don't have to find correct macro to detect it inside the code
  - automatically adding tests for detected test cases, rather than list them manually as in all_features

> From my testing with the stl stringification I'm fairly certain the features C++>11 don't seem to get tested iirc, even if they're available
I don't get it. As I understand, there is no CXX_STANDARD configuration in the build matrix on CI. So only C++11/14 is tested on CI (MSVC has dropped C++11 support eventually). If you talking about this, then yes, this is a problem I have no idea how to solve. Adding CXX_STANDARD into the build matrix multiplies the number of configurations by 4. It's insane.

Anyway, testing of std classes stringification is more or less optional, because currently, doctest doesn't support it. Tests are just declaring it. It's more like preparation for the future.

examples/stringification/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
+# separated executable to ensure there is no specializations of StringMaker leaked
+# In opposite to all_features/stringification.cpp this test intended to check all the bundled
+# output rules for all the fundamental types, std containers etc
+# it might overlap with other tests, but better to test twice, than miss some cases
+add_executable(stringification main.cpp)
+target_link_libraries(stringification doctest)
+target_compile_definitions(stringification PRIVATE SIZEOF_VOID_P=${CMAKE_SIZEOF_VOID_P})
```
Saalvage, line 7: Just calculate `SIZEOF_VOID_P` at runtime?

examples/stringification/main.cpp
```diff
@@ -0,0 +1,491 @@
...
+
+#else
+    add_stub_asserts(7);
+#endif  // CXX_STANDARD >= 17
+
+#if CXX_STANDARD >= 20
+    {
+        std::array<int, 3> ar {{ 9, 1, 5 }};
+        std::span<int> value { ar };
+        CHECK_EQ(toString(value), "{?}");
```
Saalvage, line 460: I want to merge the `string-fix` branch into dev for 2.4.9, while I would like to postpone the stl stringification extension header (for which tests already exist) to 2.5, so either this PR will need to be postponed until 2.5 or you should split it into two parts

examples/stringification/main.cpp
```diff
@@ -0,0 +1,491 @@
...
+    add_stub_asserts(17);
+#endif  // CXX_STANDARD >= 11
+
+#if CXX_STANDARD >= 14
+    {
+        std::integer_sequence<int, 9, 2, 5, 1, 9, 1, 6> value {};
+        CHECK_EQ(toString(value), "{?}");
+    }
+#else
+    add_stub_asserts(1);
```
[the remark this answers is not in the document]
YarikTH: Both cases are invalid. I've just tried them both.

-  "-minimal" completely removes output in console output format, but it doesn't affect XML and junit output formats. They still contain assert statistics
- NO_OUTPUT is either broken or intended for compile-time checks only. Because the output is ignored and not compared, retcode is completely ignored. So any test no matter of failed asserts passed as a result. In the initial version, I added alternative function doctest_add_test_2 just to workaround this behaviour

examples/stringification/CMakeLists.txt
YarikTH: Then instead of more or less cool-looking compile-time strings `POINTER_PREFIX "abadf00d"` I will need string concatenation. Ew.

This won't be a problem if I rebase this branch against the fix-string branch where pointer "abadf00d" is stringified as "0xabadf00d" independently of SIZEOF_VOID_P.

examples/stringification/main.cpp
YarikTH: I still see no problems here. We check that *currently* stl stringification is absent. All is fine. When it is added, this test will be changed with the actual stringification result.

Saalvage: Testing that something is absent seems like nothing but bloat to me. I'd really like to have everything but the working tests (so that includes everything relating to a specific standard version) completely removed from this PR before merging. If you don't want to put in the work for it then I'd happily do it for you.

examples/stringification/main.cpp
```diff
@@ -0,0 +1,491 @@
...
+#   define POINTER_PREFIX "0x"
+#elif SIZEOF_VOID_P == 8
+#   define POINTER_PREFIX "0x00000000"
+#endif
+    
+    // doctest overloads toString() for pointer types. So string doesn't match ostreamed result
+    // NULL is handled separatly somewhat
+    // other value representation has fixed size string representation depending on sizeof(void*)
+    CHECK_EQ(toString(static_cast<int*>(nullptr)), "NULL");
+    CHECK_EQ(toString(reinterpret_cast<int*>(0xABADF00D)), POINTER_PREFIX "abadf00d");
```
Saalvage, line 241: Correct me if I'm wrong, but I'm fairly certain stringification of pointers is implementation defined (comes down to `printf` with the `%p` specifier). Doctest only handles `nullptr`s seperately. We could implement pointer stringification implementation-independently but I'm honestly not sure if we should bother. (E.g. it would require template metaprogramming to get the amount of padding 0s needed.)

PR #592, opened 2022-01-18
Ignore CMake and MacOS generated files

LeonBrands:
Ignores macOS .DS_Store files, Xcode project files, and CMake generated files. These files keep showing up as "changes" when doctest is added in as a git submodule and included in a CMake project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,9 +44,32 @@ bazel-*
 *.xcodeproj/
 XCBuildData/
 
+# Visual Studio
+*.vcxproj
+*.vcxproj.filters
+*.sln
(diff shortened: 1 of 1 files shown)


Saalvage:
The exclusion of a folder called `build` in the root of the repo is intended to be the catch-all for basically all of these kinds of issues. Is that not sufficient? If it actually is not then this lgtm!

LeonBrands:
build/ works fine if cmake is configured to output inside of build/ but this isnt guaranteed

Saalvage:
Thanks for your contribution!

PR #603, opened 2022-01-25
Add IsNaN operator!

Saalvage:


Saalvage:
As always, needs docs

PR #611, opened 2022-02-04
Fix mpi subcase

BerengerBerthoul:
## Description

I am the orginal author of the MPI extension of doctest with my colleague @bmaugars. The extension was integrated about two years ago, see #418. It appears that there is a bug when using several `SUBCASE` in an `MPI_TEST_CASE`

Before this PR, the following MPI_TEST_CASE would deadlock if launched on strictly more than 2 MPI processes.
```

MPI_TEST_CASE("Parallel tests with subcases",2) {

  SUBCASE("sub_case 0") {
    CHECK( test_nb_procs == 2 );
  }
  SUBCASE("sub_case 1") {
    CHECK( test_rank == 0 ); // should fail on proc 1
  }
}

```
This PR corrects the behavior.

Also now, the test is skipped if there is not enought processes to run it, and only a warning message is issued (before, the test would have been reported as failed).

### DIFF
--- a/doc/markdown/extensions.md
+++ b/doc/markdown/extensions.md
@@ -62,7 +62,7 @@ You need to launch the unit tests with an ```mpirun``` or ```mpiexec``` command:
 mpirun -np 2 unit_test_executable.exe
 ```
 
-```doctest::mpi_init_thread()``` should be called before running the unit tests, and ```doctest::mpi_finalize()``` at the end of the program. Also, using the default console reporter will result in each process writing everything in the same place, which is not what we want. Two reporters are provided and can be enabled. A complete ```main()``` would be:
+```doctest::mpi_init_thread()``` must be called before running the unit tests, and ```doctest::mpi_finalize()``` at the end of the program. Also, using the default console reporter will result in each process writing everything in the same place, which is not what we want. Two reporters are provided and can be enabled. A complete ```main()``` would be:
 
 
 ```c++
--- a/doc/markdown/extensions.md
+++ b/doc/markdown/extensions.md
@@ -33,16 +33,16 @@ int my_function_to_test(MPI_Comm comm) {
 
 MPI_TEST_CASE("test over two processes",2) { // Parallel test on 2 processes
   int x = my_function_to_test(test_comm);
-  
+
   MPI_CHECK( 0,  x==10 ); // CHECK for rank 0, that x==10
   MPI_CHECK( 1,  x==11 ); // CHECK for rank 1, that x==11
 }
 ```
 
-An ```MPI_TEST_CASE``` is like a regular ```TEST_CASE```, except it takes a second argument, which is the number of processes needed to run the test.  If the number of processes is less than 2, the test will fail. If the number of processes is greater than or equal to 2, it will create a sub-communicator over 2 processes, called ```test_comm```, and execute the test over these processes. Three objects are provided by ```MPI_TEST_CASE```: 
+An ```MPI_TEST_CASE``` is like a regular ```TEST_CASE```, except it takes a second argument, which is the number of processes needed to run the test.  If the number of processes is less than 2, the test will fail. If the number of processes is greater than or equal to 2, it will create a sub-communicator over 2 processes, called ```test_comm```, and execute the test over these processes. Three objects are provided by ```MPI_TEST_CASE```:
  * ```test_comm```, of type ```MPI_Comm```: the mpi communicator on which the test is running,
  * ```test_rank``` and ```test_nb_procs```, two ```int``` giving respectively the rank of the current process and the size of the communicator for ```test_comm```. These last two are just here for convenience and could be retrieved from ```test_comm```.
- 
+
 We always have:
 
 ```c++
@@ -57,12 +57,12 @@ It is possible to use regular assertions in an ```MPI_TEST_CASE```. MPI-specific
 
 ## The main entry points and mpi reporters
 
-You need to launch the unit tests with an ```mpirun``` command:
+You need to launch the unit tests with an ```mpirun``` or ```mpiexec``` command:
 ```
 mpirun -np 2 unit_test_executable.exe
 ```
 
-```MPI_Init``` should be called before running the unit tests. Also, using the default console reporter will result in each process writing everything in the same place, which is not what we want. Two reporters are provided and can be enabled. A complete ```main()``` would be:
+```doctest::mpi_init_thread()``` should be called before running the unit tests, and ```doctest::mpi_finalize()``` at the end of the program. Also, using the default console reporter will result in each process writing everything in the same place, which is not what we want. Two reporters are provided and can be enabled. A complete ```main()``` would be:
 
 
 ```c++
@@ -71,7 +71,7 @@ mpirun -np 2 unit_test_executable.exe
 #include "doctest/extensions/doctest_mpi.h"
 
 int main(int argc, char** argv) {
-  MPI_Init(&argc, &argv);
+  doctest::mpi_init_thread(argc,argv,MPI_THREAD_MULTIPLE); // Or any MPI thread level
 
   doctest::Context ctx;
   ctx.setOption("reporters", "MpiConsoleReporter");
@@ -81,7 +81,7 @@ int main(int argc, char** argv) {
 
   int test_result = ctx.run();
 
-  MPI_Finalize();
+  doctest::mpi_finalize();
 
   return test_result;
 }
@@ -103,7 +103,7 @@ std_e_mpi_unit_tests
 [doctest] run with "--help" for options
 ===============================================================================
 path/to/test.cpp:30:
-TEST CASE: my test case 
+TEST CASE: my test case
 
 On rank [2] : path/to/test.cpp:35: CHECK( x==-1 ) is NOT correct!
   values: CHECK( 0 == -1 )
@@ -113,11 +113,35 @@ On rank [2] : path/to/test.cpp:35: CHECK( x==-1 ) is NOT correct!
 [doctest] assertions:      2 |      2 passed |      0 failed |
 [doctest] Status: SUCCESS!
 ===============================================================================
-[doctest] glob assertions:      5 |      4 passed |      1 failed |
+[doctest] assertions on all processes:   5 |   4 passed |      1 failed |
 ===============================================================================
-[doctest] fail on rank:     
+[doctest] fail on rank:
     -> On rank [2] with 1 test failed
 [doctest] Status: FAILURE!
+```
+
+If the test executable is launch with less processes than the number of processes required by one test, the test is skipped and marqued as such in the mpi console reporter:
+
+
+```c++
+MPI_TEST_CASE("my_test",3) {
+  // ...
+}
+```
+
+```
+mpirun -np 2 unit_test_executable.exe
+```
+
+```
+===============================================================================
+[doctest] test cases:      1 |      1 passed |      0 failed |      1 skipped
+[doctest] assertions:      1 |      1 passed |      0 failed |
+[doctest] Status: SUCCESS!
+===============================================================================
+[doctest] assertions on all processes:   1 |   1 passed |      0 failed |
+[doctest] WARNING: Skipped 1 test requiring more than 2 MPI processes to run
+===============================================================================
 ```
 
 ### MpiFileReporter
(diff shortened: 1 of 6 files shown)


doctest/extensions/doctest_mpi.h
```diff
@@ -2,80 +2,104 @@
...
+// we need a sub-communicator of N procs to execute it.
+// It is then registered here and can be re-used
+// by other tests that requires a sub-comm of the same size
+std::unordered_map<int,mpi_sub_comm> sub_comms_by_size;
+
+// Record if at least one MPI_TEST_CASE was registered "skipped"
+// because there is not enought procs to execute it
+int nb_test_cases_skipped_insufficient_procs = 0;
+
+// Can be safely called before MPI_Init()
```
Saalvage, line 21: Shouldn't this be `doctest::mpi_init_thread` instead of `MPI_Init`? (same as a few lines below)

Saalvage:
> Also now, the test is skipped if there is not enought processes to run it, and only a warning message is issued (before, the test would have been reported as failed).
I feel like it might be preferable to have a config option to toggle this, with the default behavior being the existing one?

doc/markdown/extensions.md
```diff
@@ -57,12 +57,12 @@ It is possible to use regular assertions in an ```MPI_TEST_CASE```. MPI-specific
...
 ## The main entry points and mpi reporters
 
-You need to launch the unit tests with an ```mpirun``` command:
+You need to launch the unit tests with an ```mpirun``` or ```mpiexec``` command:
 ```
 mpirun -np 2 unit_test_executable.exe
 ```
 
-```MPI_Init``` should be called before running the unit tests. Also, using the default console reporter will result in each process writing everything in the same place, which is not what we want. Two reporters are provided and can be enabled. A complete ```main()``` would be:
+```doctest::mpi_init_thread()``` should be called before running the unit tests, and ```doctest::mpi_finalize()``` at the end of the program. Also, using the default console reporter will result in each process writing everything in the same place, which is not what we want. Two reporters are provided and can be enabled. A complete ```main()``` would be:
```
Saalvage, line 65: Is this a breaking change or simply a fix to the docs?

doctest/extensions/doctest_mpi.h
BerengerBerthoul: I think we could put `MPI_Init_thread ` here since this is the one used by `doctest::mpi_init_thread`, but both `MPI_Init` and `MPI_Init_thread `  essentially do the same thing and `MPI_Init `is better known, which is why I put this name here

BerengerBerthoul:
> > Also now, the test is skipped if there is not enought processes to run it, and only a warning message is issued (before, the test would have been reported as failed).
> 
> I feel like it might be preferable to have a config option to toggle this, with the default behavior being the existing one?

As you prefer. I was wondering the same, but decided against it because the old behavior was essentially here by default because I didn't know the `mpi_comm_world_size` init trick that now allows us to mark the test as skipped. I think that everyone prefers the new behavior (at least for the users I know)

doc/markdown/extensions.md
BerengerBerthoul: This is a breaking change.

Saalvage:
In that case the way it's currently implemented is probably fine!

Saalvage:
So the reason I can't merge this as of yet is because I'm personally not at all familiar with MPI, so I'd like to have someone who is to have at least a second glance at this.
@philbucher or @onqtam does one of you think they could dedicate the time necessary to give this a review?

doctest/extensions/doctest_mpi.h
```diff
@@ -2,80 +2,104 @@
...
+mpi_comm_world_size() {
+  #if defined(OPEN_MPI)
+    const char* size_str = std::getenv("OMPI_COMM_WORLD_SIZE");
+  #elif defined(I_MPI_VERSION) || defined(MPI_VERSION)
+    const char* size_str = std::getenv("PMI_SIZE"); // see https://community.intel.com/t5/Intel-oneAPI-HPC-Toolkit/Environment-variables-defined-by-intel-mpirun/td-p/1096703
+  #else
+    #error "Unknown MPI implementation"
+  #endif
+  if (size_str==nullptr) return 1; // not launched with mpirun/mpiexec, so assume only one process
+  return std::stoi(size_str);
```
philbucher, line 35: maybe add a check here to make sure it is > 0? Just to be safe

doctest/extensions/doctest_mpi.h
```diff
@@ -2,80 +2,104 @@
...
-
-  ~mpi_sub_comm() {
-    if(comm != MPI_COMM_NULL){
-      MPI_Comm_free(&comm);
-    }
+}
+int mpi_init_thread(int argc, char *argv[], int required_thread_support) {
+  int provided_thread_support;
+  MPI_Init_thread(&argc, &argv, required_thread_support, &provided_thread_support);
+  if (provided_thread_support!=MPI_THREAD_MULTIPLE) {
```
philbucher, line 50:
```suggestion
  if (provided_thread_support!=required_thread_support) {
```

should this be correct?

doctest/extensions/mpi_sub_comm.h
```diff
@@ -0,0 +1,81 @@
...
+
+      if(comm != MPI_COMM_NULL){
+        MPI_Comm_rank(comm, &rank);
+        assert(rank==comm_world_rank);
+      }
+    }
+  }
+
+  void destroy_comm() {
+    if(comm != MPI_COMM_NULL){
```
philbucher, line 55: what about `MPI_COM_SELF`? Not sure if applicable
we have a similar wrapper for `MPI_Comm`:
https://github.com/KratosMultiphysics/Kratos/blob/1ce05c6a3eb8769f03a47646f78ec23fdbb89926/kratos/mpi/sources/mpi_data_communicator.cpp#L238-L245

doctest/extensions/doctest_mpi.h
```diff
@@ -2,80 +2,104 @@
...
+//   to prevent execution of tests where there is not enough procs,
+//   but doctest::skip() is called during test registration, that is, before main(), and hence before MPI_Init()
+int
+mpi_comm_world_size() {
+  #if defined(OPEN_MPI)
+    const char* size_str = std::getenv("OMPI_COMM_WORLD_SIZE");
+  #elif defined(I_MPI_VERSION) || defined(MPI_VERSION)
+    const char* size_str = std::getenv("PMI_SIZE"); // see https://community.intel.com/t5/Intel-oneAPI-HPC-Toolkit/Environment-variables-defined-by-intel-mpirun/td-p/1096703
+  #else
+    #error "Unknown MPI implementation"
```
philbucher, line 32: I guess OpenMPI and IntelMPI are the two most commonly used, but we should keep our eyes open also to support others like MPICH

doc/markdown/extensions.md
philbucher, line 65:
```suggestion
```doctest::mpi_init_thread()``` must be called before running the unit tests, and ```doctest::mpi_finalize()``` at the end of the program. Also, using the default console reporter will result in each process writing everything in the same place, which is not what we want. Two reporters are provided and can be enabled. A complete ```main()``` would be:
```

this is not optional is it?

doctest/extensions/doctest_mpi.h
```diff
@@ -2,80 +2,104 @@
...
+std::unordered_map<int,mpi_sub_comm> sub_comms_by_size;
+
+// Record if at least one MPI_TEST_CASE was registered "skipped"
+// because there is not enought procs to execute it
+int nb_test_cases_skipped_insufficient_procs = 0;
+
+// Can be safely called before MPI_Init()
+//   This is needed for MPI_TEST_CASE because we use doctest::skip()
+//   to prevent execution of tests where there is not enough procs,
+//   but doctest::skip() is called during test registration, that is, before main(), and hence before MPI_Init()
```
philbucher, line 24: Another usecase: `doctest_discover_tests` (used for registering the tests with [CTest](https://cmake.org/cmake/help/latest/manual/ctest.1.html))

=> there the binary is also executed to list the test cases

(I made it work with MPI-tests for my stuff but didn't make a PR yet as it requires some cleanup etc.)

philbucher (commented):
Thx for the fixes @BerengerBerthoul !
I have some minor comments, but overall it looks good to me!

I don't love the `mpi_comm_world_size` function but for now this seems to be the only way to properly integrate the MPI-tests, so I would leave it as is and keep it in mind, maybe a better solution can be found in the future

doctest/extensions/doctest_mpi.h
philbucher: another idea:
We could optionally ask the user for a custom env-var in case we don't directly support the MPI-implementation
i.e. sth on the line of the following:
~~~c++
for (const auto& env_var_name : {"DOCTEST_COMM_WORLD_SIZE", "OMPI_COMM_WORLD_SIZE", "PMI_SIZE"})  { // can be easily extended if we know the env-var names from other MPI-implementations
    const char* size_str = std::getenv(env_var_name);
    if (size_str) {
        int size = std::stoi(size_str);
        assert (size > 0);
        return size;
    }
}

error: please define "DOCTEST_COMM_WORLD_SIZE" to match the number of cores running with
~~~

and then inside `doctest::mpi_init_thread()` check if it is consistent with what is returned from `MPI_Comm_size`

doctest/extensions/doctest_mpi.h
BerengerBerthoul: On this suggestion, I have done the following : we record the value in a global variable, then test it against `MPI_Comm_size(MPI_COMM_WORLD)` during `doctest::mpi_init_thread`. So we should have an exception if this brittle `mpi_comm_world_size  `function has a problem (I don't like the function either... but don't know of any other way...)

doctest/extensions/doctest_mpi.h
BerengerBerthoul: good catch

doctest/extensions/mpi_sub_comm.h
BerengerBerthoul: the communicator is created by `MPI_Comm_split`, so it cannot be `MPI_COMM_SELF ` nor `MPI_COMM_WORLD`

doctest/extensions/doctest_mpi.h
BerengerBerthoul: Yes I would like to support other implementations but I am not using them personally (and installing MPI is not something I particularly long for). I will definitely upstream a patch if I ever get my hands on one of them.
Regarding your `DOCTEST_COMM_WORLD_SIZE` suggestion, I was going to implement it, but I think that we are better off failing with a clear message that encourages to upstream the problem rather than silently allowing people to get around with an incomplete support. Plus, I am wondering if `PMI_SIZE `isn't the name for MPICH and MVAPICH anyway since they all share a common root. What do you think ?

doc/markdown/extensions.md
BerengerBerthoul: yes good catch

doctest/extensions/doctest_mpi.h
BerengerBerthoul: Good to know. Do you want me to add this use case to the source comment ? Or you can do it with your PR ?

philbucher:
I tried this now myself with OpenMPI and MPICH and have the following conclusion:
- MPICH is already supported, it uses the same env-vars as IntelMPI :tada: 
- can you please also update [examples/mpi/main.cpp](https://github.com/doctest/doctest/blob/master/examples/mpi/main.cpp) to use `doctest::mpi_init_thread` and `doctest::mpi_finalize`? => otherwise it crashes (as expected) at exiting bcs `MPI_comm_free` is called after `MPI_Finalize`

Otherwise for me this is ready to be merged!

BerengerBerthoul:
@philbucher Ok the warnings and example have been fixed.
Thanks @Saalvage and @philbucher for the quick review and feedback !

philbucher (approved):
prefect, thx now it works perfectly!

philbucher:
@Saalvage @onqtam should I merge this, or how should we proceed?

Saalvage:
I would like to have a crack at adding CI for MPI, but I guess that shouldn't stop this from being merged. :)

philbucher:
OpenMPI is quite straight forward, IntelMPI is a bit more annoying to get working

BTW does doctest have a regular CI with the Intel compiler already?

Saalvage:
Not yet, #555 adds support for ICC, but there were some issues with it requiring different configurations, depending on the platform it is running on which I haven't had the time to look into properly.

PR #612, opened 2022-02-09
Fix MPI extension to work if launched without mpirun/mpiexec

BerengerBerthoul:
## Description
A bug was introduced by PR #611: now if the doctest mpi extension is lauched without mpirun/mpiexec, it triggers an internal error, whereas it really should execute sequential tests and report parallel tests as skipped.

Turns out this is a one character change... Sorry for not noticing it


Saalvage:
CI failing seems unrelated to the PR

Saalvage:
Thanks for noticing and fixing it :)

PR #616, opened 2022-02-14
Allow escaping backslash with backslash in filters (#614)

### LINKED ISSUE #614: -tc does not work with names ending in backslash or having `\,` in their names
## Description
#493 has introduced comma escaping for `-tc` and other parameters which use `parseCommaSepArgs` (as a fix for #398): one can run test named `A,improved` with `-tc=A\,improved`. The code finds all `\,` substrings and treats them as a `,` inside test name regardless of context. Unfortunately,that gives `\` a special meaning which it did not have previously, resulting in more similar issues.

### Steps to reproduce
Try to run exactly two tests from below: one `a*`, followed by `b`:

```c++
#include "doctest.h"
#include <iostream>

TEST_CASE("a\\") {
    std::cout << "test a slash\n";
}

TEST_CASE("a\\,") {
    std::cout << "test a slash comma\n";
}

TEST_CASE("a\\\\") {
    std::cout << "test a slash slash\n";
}

TEST_CASE("a\\\\,") {
    std::cout << "test a slash slash comma\n";
}

TEST_CASE("b") {
    std::cout << "b\n";
}
```

For example, running `-tc=a\,b` or `-tc=a\\,b` runs no tests as it tries to find a test named `a,b` or `a\,b` instead of treating the comma as a separator.

### Extra information
My proposal:

1. Treat `\` as _the_ escaping character. Specifically, it should be always followed by either `,` or `\`. `\,` emits `,`, `\\` emits `\\`. It's consistent with how escaping works in other applications: it's always possible to escape the escaping symbol one way or the other.
2. All other characters after `\` are banned and emit an error. E.g. `a\,b` runs a single test called `a,b` (as in 2.4.7/2.4.8); `a\\,b` runs two tests called `a\` and `b` (runs `a\,b` in 2.4.7/2.4.8, runs `a\\` and `b` prior to that); `a\\\b` is an error (runs `a\\\b` in all previous versions).

This proposal is breaking if a user had a test which had `\` in the name (probably spelled as `\\` in C++ source code). A softer approach (but much more unusual to my taste of escaping) is to keep the current behavior: `\X` emits `\X` if `X` is neither `\` nor `,`.

In both cases each test case name has at least one representation:

1. Characters other than `\` and `,` are left as is. It's ok because they bear no special meaning.
2. `\` is replaced with `\\`.
3. `,` is replaced with `\,`.

The difference between harder and softer approach arises in test case names like `ensure that '\' is parsed correctly`, where `\` is followed by neither `\` nor `,`. It will continue to work with the softer approach and break with my original apporach.

* doctest version: **v2.4.8**

yeputons:
## Description
All filters now treat double backslash as a single backslash. It allows specifying strings like `\,` in filters and is more consistent with other escaping systems. E.g. it's now possible to run a test called `a\,` with `-tc=a\\\,`.

That way it's possible to pass any test name to the `-tc`, which is important for IDEs like CLion. They had some integration issues previously to this and #493: https://youtrack.jetbrains.com/issue/CPP-25111

## GitHub Issues
Closes #614

### DIFF
--- a/doc/markdown/commandline.md
+++ b/doc/markdown/commandline.md
@@ -9,7 +9,9 @@
 **Bool options** - they expect ```1```/```yes```/```on```/```true``` or ```0```/```no```/```off```/```false``` after the ```=``` sign - but they can also be used like flags and the ```=value``` part can be skipped - then ```true``` is assumed.  
 
 **Filters** - a comma-separated list of wildcards for matching values - where ```*``` means "match any sequence" and ```?``` means "match any one character".
-To pass patterns with intervals use ```""``` like this:  ```--test-case="*no sound*,vaguely named test number ?"```. Patterns that contain a comma can be escaped with ```\``` (example: ```--test-case=this\,test\,has\,commas```).
+To pass patterns with intervals use ```""``` like this:  ```--test-case="*no sound*,vaguely named test number ?"```. Patterns that contain a comma or a backslash can be escaped with ```\``` (example: ```--test-case=this\,test\,has\,commas\,and\,a\\\,backslash\,followed\,by\,a\,comma```).
+If a backslash is followed by neither ```\``` nor ```,``` it's left as is, e.g. ```--test-case="Test that \ works correctly"```.
+Be careful: your shell may use ```\``` for escaping as well, so `\` may actually get consumed by the shell instead of doctest.
 
 All the options can also be set with code (defaults/overrides) if the user [**supplies the ```main()``` function**](main.md).
 
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -6302,8 +6302,8 @@ namespace {
                 char character = *current++;
                 if(seenBackslash) {
                     seenBackslash = false;
-                    if(character == ',') {
-                        s.put(',');
+                    if(character == ',' || character == '\\') {
+                        s.put(character);
                         continue;
                     }
                     s.put('\\');
--- a/doctest/parts/doctest.cpp
+++ b/doctest/parts/doctest.cpp
@@ -3332,8 +3332,8 @@ namespace {
                 char character = *current++;
                 if(seenBackslash) {
                     seenBackslash = false;
-                    if(character == ',') {
-                        s.put(',');
+                    if(character == ',' || character == '\\') {
+                        s.put(character);
                         continue;
                     }
                     s.put('\\');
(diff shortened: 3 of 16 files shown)


yeputons:
Failing checks seem weird and not relevant to this PR, looks like VS 2019 is no longer installed.

[This is a faliure](https://github.com/doctest/doctest/runs/5183294240?check_suite_focus=true):

```
Run cmake -B build -S . -G "Visual Studio 16 2019" \ -A x64 -T v142
CMake Error at CMakeLists.txt:12 (project):
  Generator
-- Configuring incomplete, errors occurred!

See also "D:/a/doctest/doctest/build/CMakeFiles/CMakeOutput.log".
    Visual Studio 16 2019

  could not find any instance of Visual Studio.



Error: Process completed with exit code 1.
```

but [this is a success](https://github.com/doctest/doctest/runs/5183294395?check_suite_focus=true):

```

Run cmake -B build -S . -G "Visual Studio 17 2022" \ -A x64 -T v143
-- The CXX compiler identification is MSVC 19.30.30709.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Tools/MSVC/14.30.30705/bin/Hostx64/x64/cl.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Looking for C++ include pthread.h
-- Looking for C++ include pthread.h - not found
-- Found Threads: TRUE  
-- Checking for module 'mpi-cxx'
--   Can't find mpi-cxx.pc in any of C:/Strawberry/c/lib/pkgconfig
use the PKG_CONFIG_PATH environment variable, or
specify extra search paths via 'search_paths'
-- Could NOT find MPI_CXX (missing: MPI_CXX_LIB_NAMES MPI_CXX_HEADER_DIR MPI_CXX_WORKS) 
-- Could NOT find MPI (missing: MPI_CXX_FOUND CXX) 
-- Configuring done
-- Generating done
-- Build files have been written to: D:/a/doctest/doctest/build
```

Saalvage:
GitHub changed the `windows-latest` environment to be `windows-2022` rather than `windows-2019`. Honestly my bad, should've just used `windows-2019` explicitely from the start.

Saalvage:
I've taken the liberty of writing a small test for this, which will hopefully make all of the CI pass! ;)

A few obervations:
- I will keep the "soft" approach in order to maintain backwards compatibility for the time being, but I will readress this in 2.5. However, I'm wondering whether just having `\x` emit `x` (`,` != `x` != `\`) might be just as good, but I guess there'll be enough time to consider all alternatives!
- Is the ability to escape the other characters with a special meaning (`?` and `*`) something worth considering? (I'm aware that their parsing is done elsewhere.)
- It seems CMake is unable to output a `\` at the end of a CLI arguments in our test cases, which is why I had to postfix the test case with a semantically irrelevant `:`.

Thanks for your contribution!

PR #617, opened 2022-02-16
Add alias target for doctest_with_main

jessestricker:
## Description

This is just a small change:
I've added the alias `doctest::doctest_with_main` for the library target `doctest_with_main`.

## Motivation

There already is the alias `doctest::doctest` for the library target `doctest`, so this change will increase consistency.
Indeed, I've recently tried the alias `doctest::doctest_with_main` in a personal project and was surprised it didn't exist.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,6 +69,7 @@ endif()
 
 if(${DOCTEST_WITH_MAIN_IN_STATIC_LIB})
     add_library(${PROJECT_NAME}_with_main STATIC EXCLUDE_FROM_ALL ${doctest_parts_folder}/doctest.cpp)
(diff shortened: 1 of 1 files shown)


Saalvage:
Change looks good, I'm only wondering at the purpose of `doctest::doctest` although admittedly my experience with CMake is pretty limited!

jessestricker:
It is considered a best practise to "namespace" your public facing targets. This makes them look the the targets CMake makes available from `find_package(...)`, like `Boost::boost` or `Qt::Widgets`.
So it became a convention a lot of projects follow. Here are some sources giving more elaborate reason:

- https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1#using-a-library-defined-in-the-same-cmake-tree-should-look-the-same-as-using-an-external-library
- https://cliutils.gitlab.io/modern-cmake/chapters/install.html

Saalvage:
Thanks for your contribution!

PR #620, opened 2022-02-20
Add contains option to checks.

### LINKED ISSUE #619: Feature request: CHECK_THROWS_WITH with contains option
## Description
Currently CHECK_THROWS_WITH only checks if the string exactly matches the exception string. I would like to add a version which checks whether the exception contain a certain string. This is the same functionality as Catch2 `CHECK_THROWS_WITH(func(),Contains("needle"))`. 

I looked at the code and I think it could be implemented by adding a macro `CHECK_THROWS_WITH_CONTAINS` and adds in the ResultBuilder::log() something like (which worked locally for me): 
```
        } else if(m_at & assertType::is_throws_with_contains) { //!OCLINT bitwise operator in conditional
            m_failed = strstr(m_exception.c_str(), m_exception_string) == nullptr;
```

I am happy make a pull request if something like this would be an acceptable addition.

MFraters:
## Description
This is an implementation of what was suggested in #619. It implements a `Contains` class for the `CHECK_THROWS_WITH` (and checks in general). This allows for checking whether the throw message contains a certain string, similar to what Catch2 does. 

I first tried the union approach as suggested in the related issue, but I ran in a host of issues, including that due to non-trivial members, the default constructors and operators where deleted. I found especially the copy constructor/operator= hard to replace, since it requires knowledge about what member of the union is being used. I now use a bool to check whether it is a contains or not. This approach seems easier to me, and I can't see significant downsides of this implementation, but I am open to further suggestions.

I have not implemented test yet, because I would first like to know if this approach is acceptable, but it works well as a replacement for catch2 on my code base. I tried to format the cod with `clang-format -style=file -i doctest/parts/doctest*`, but it also changes the format of lines I haven't changed in this pull request.

## GitHub Issues
closes  #619.

### DIFF
--- a/doctest/parts/doctest.cpp
+++ b/doctest/parts/doctest.cpp
@@ -670,12 +670,25 @@ int String::compare(const String& other, bool no_case) const {
     return compare(other.c_str(), no_case);
 }
 
-Contains::Contains(const char* in){
-    string = String(in);
+
+Contains::Contains() {
+    string = "";
 }
 
 Contains::~Contains(){
     string.~String();
+}
+
+Contains::Contains(const char* in){
+    string = String(in);
+}
+
+Contains::Contains(const Contains& other)
+: string(other.string) {}
+
+Contains& Contains::operator=(const Contains& other) {
+    string = other.string;
+    return *this;
 }
 
 bool Contains::checkWith(const String& full_string) const {
--- a/doctest/parts/doctest.cpp
+++ b/doctest/parts/doctest.cpp
@@ -670,25 +670,12 @@ int String::compare(const String& other, bool no_case) const {
     return compare(other.c_str(), no_case);
 }
 
-
-Contains::Contains() {
-    string = "";
-}
-
-Contains::~Contains(){
-    string.~String();
-}
-
 Contains::Contains(const char* in){
     string = String(in);
 }
 
-Contains::Contains(const Contains& other)
-: string(other.string) {}
-
-Contains& Contains::operator=(const Contains& other) {
-    string = other.string;
-    return *this;
+Contains::~Contains(){
+    string.~String();
 }
 
 bool Contains::checkWith(const String& full_string) const {
@@ -1774,7 +1761,7 @@ namespace detail {
         m_failed           = true;
         m_threw            = false;
         m_threw_as         = false;
-        m_exception_type   = exception_type;
+        m_exception_type  = exception_type;
         m_exception_string.string = exception_string;
         m_contains         = false;
 #if DOCTEST_MSVC
--- a/doctest/parts/doctest.cpp
+++ b/doctest/parts/doctest.cpp
@@ -670,12 +670,12 @@ int String::compare(const String& other, bool no_case) const {
     return compare(other.c_str(), no_case);
 }
 
-const char *String::compare(const Contains& other) const {
-    return strstr(c_str(), other.string.c_str());
-}
-
 Contains::Contains(const char* in){
     string = String(in);
+}
+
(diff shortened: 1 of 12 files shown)


doctest/parts/doctest_fwd.h
```diff
@@ -576,12 +578,21 @@ class DOCTEST_INTERFACE String
 
     int compare(const char* other, bool no_case = false) const;
     int compare(const String& other, bool no_case = false) const;
+    const char *compare(const Contains& other) const;
```
Saalvage, line 581: Why does this return a `const char*`? Imo this should be a method on `Contains` (e.g.: `Contains::checkWith(String)`) that simply returns a bool. This will also make the `Contains` forward declaration superfluous and make the dependency between `String` and `Contains` unidirectional.

doctest/parts/doctest_fwd.h
```diff
@@ -576,12 +578,21 @@ class DOCTEST_INTERFACE String
...
+    Contains(const char* string);
+
+    String string;
 };
 
 DOCTEST_INTERFACE String operator+(const String& lhs, const String& rhs);
 
 DOCTEST_INTERFACE bool operator==(const String& lhs, const String& rhs);
 DOCTEST_INTERFACE bool operator!=(const String& lhs, const String& rhs);
+DOCTEST_INTERFACE bool operator!=(const String& lhs, const Contains& rhs);
```
Saalvage, line 595: Should have `operator==` as well. Just for general utility and maintaining symmetry.

doctest/parts/doctest.cpp
```diff
@@ -1748,7 +1757,27 @@ namespace detail {
         m_threw            = false;
         m_threw_as         = false;
         m_exception_type   = exception_type;
-        m_exception_string = exception_string;
+        m_exception_string = String(exception_string);
```
Saalvage, line 1760: `exception_string` should be made a `String` as well.

doctest/parts/doctest_fwd.h
```diff
@@ -743,7 +754,8 @@ struct DOCTEST_INTERFACE AssertData
     // for specific exception-related asserts
     bool        m_threw_as;
     const char* m_exception_type;
-    const char* m_exception_string;
+    String      m_exception_string;
```
Saalvage, line 757: I think this should be replaced with a union. Otherwise we have to construct a new `Contains` on every comparison (which includes a string copy), after we've destroyed the one we've been passed in the constructor.

Saalvage:
Aside from the few minor tweaks I'd like to see this lgtm!

MFraters:
Thanks for the review! I think I have addressed all your issues, I even managed to get the union working :)

I have added some more tests along the lines you suggested, but if you have more in mind, please let me know.

Saalvage:
If you can't figure it out just leave it to me, unions can be quite tricky to work with at times and warning suppression is also pretty yucky.

MFraters:
Yes, I am happy leave getting the union to you. It work on my computer (linux), but windows is giving a warning/error which is not very helpful and nearly everything I can find on that error is either advice to ignore it or code which upon testing doesn't solve the warning. What also doesn't help is that I can't quickly iterated do not being able to reproduce the issue locally.  

Let me know if you want me to get rid of all the "Attempt to fix windows specific warning about unions" commits, or if there is anything else I can do to move this along.

Saalvage:
All good, I'll squash when merging anyways ;)

Saalvage:
Don't bother with the test cov ci. This just needs docs and it's ready to go.

Saalvage:
Thanks for your contribution!

PR #623, opened 2022-02-28
Fix MPI extension to work with no parallel tests

BerengerBerthoul:
Fix #612 was not enough to correct #611

This PR does the following :

Instead of initializing world_size_before_init at 1 and then assigning it the right value at every MPI_TEST_CASE registration (consequently always getting 1 if there is no MPI_TEST_CASE), we initialize it immediately (and only once) with the value of mpi_comm_world_size()

Sorry for not noticing the problem, now we should be good to go


Saalvage:
Let's hope this fixes it for good!

PR #626, opened 2022-03-18
Fix properties not being passed in doctest_discover_tests

### LINKED ISSUE #621: doctest_discover_tests no longer sets ENVIRONMENT variables for discovered tests.
## Description

The CMake function `doctest_discover_tests` no longer sets ENVIRONMENT variables for discovered tests. This worked as recently as 2.4.5, but not in later versions.

### Steps to reproduce

I created a self-contained repro that checks the SOURCE_DIR environment variable isn't empty. The test passes with doctest 2.4.5 and fails with later versions.
https://github.com/quantumsteve/doctest_example

```
mkdir build
cd build
cmake ..
make
ctest
```

### Extra information

* doctest version: **v2.4.6 and later**
* Operating System: **Ubuntu 20.04**
* Compiler+version: **CMake 3.22.2, GCC 9.3.0**

quantumsteve:
## Description

This changes the position of the LABELS property so that an empty value for `${labels}` doesn't break other PROPERTY values set in `${properties}`.

I tried putting quotes around `"${labels}"`, but still couldn't successfully set an environment variable.

Here's the repo I created to test this change. 
https://github.com/quantumsteve/doctest_example

## GitHub Issues

Fixes #621

### DIFF
--- a/scripts/cmake/doctestAddTests.cmake
+++ b/scripts/cmake/doctestAddTests.cmake
@@ -105,8 +105,8 @@ foreach(line ${output})
     "${prefix}${test}${suffix}"
     PROPERTIES
     WORKING_DIRECTORY "${TEST_WORKING_DIR}"
-    LABELS ${labels}
     ${properties}
+    LABELS ${labels}
   )
   unset(labels)
   list(APPEND tests "${prefix}${test}${suffix}")


Saalvage:
Thank you for your contribution!

PR #628, opened 2022-04-07
added nolint for cert-err58

serguei-k:
Removing a clang tidy warning that appears for every test case because of [cert-err58-cpp](https://clang.llvm.org/extra/clang-tidy/checks/cert-err58-cpp.html)

Also see related [issue](https://github.com/doctest/doctest/issues/115).


onqtam:
Thanks!

Issue #495, opened 2021-04-09
Packed fields can't be accessed in 2.4.6


rhssk:
## Description
Starting with 2.4.6, members of packed structs can't be accessed in `CHECK` macros. Doing so generates a compile error: `cannot bind packed field 'X' to 'Y'`. Versions 2.4.5 and earlier do not have this issue.

### Steps to reproduce
```cpp
#define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
#include "doctest.h"

struct Foo {
	unsigned bar;
	unsigned char baz;
};

struct PackedFoo {
	unsigned bar;
	unsigned char baz;
} __attribute__((packed));

TEST_CASE("") {
	Foo foo{};
	PackedFoo packedFoo{};

	CHECK(foo.bar == 0); // OK
	CHECK(packedFoo.bar == 0); // Compile error
}
```
```bash
> g++ main.cpp -o main
In file included from main.cpp:2:
main.cpp: In function ‘void _DOCTEST_ANON_FUNC_8()’:
main.cpp:19:18: error: cannot bind packed field ‘packedFoo.PackedFoo::bar’ to ‘unsigned int&’
   19 |  CHECK(packedFoo.bar == 0);
      |        ~~~~~~~~~~^~~
doctest.h:1837:9: note: in definition of macro ‘DOCTEST_WRAP_IN_TRY’
 1837 |         x;                                                                                         \
      |         ^
doctest.h:2086:9: note: in expansion of macro ‘DOCTEST_ASSERT_IMPLEMENT_2’
 2086 |         DOCTEST_ASSERT_IMPLEMENT_2(assert_type, __VA_ARGS__);                                      \
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~
doctest.h:2104:28: note: in expansion of macro ‘DOCTEST_ASSERT_IMPLEMENT_1’
 2104 | #define DOCTEST_CHECK(...) DOCTEST_ASSERT_IMPLEMENT_1(DT_CHECK, __VA_ARGS__)
      |                            ^~~~~~~~~~~~~~~~~~~~~~~~~~
doctest.h:2573:20: note: in expansion of macro ‘DOCTEST_CHECK’
 2573 | #define CHECK(...) DOCTEST_CHECK(__VA_ARGS__)
      |                    ^~~~~~~~~~~~~
main.cpp:19:2: note: in expansion of macro ‘CHECK’
   19 |  CHECK(packedFoo.bar == 0);
[...]

PR #652, opened 2022-05-17
Cleanup of DOCTEST_DO_BINARY_EXPRESSION_COMPARISON. Fixes #651

### LINKED ISSUE #651: Comparison with implicit cast from non-const value can't be decomposed
## Description

If a class can be implicitly constructed from non-const ref and has comparison operators, the decomposer can't handle the decomposition when the the non-const ref is used.

### Steps to reproduce

```c++
struct int_pointer {
    int* p = nullptr;
    int_pointer() = default;
    
    // non-const ref used in constructor
    // we don't want to accidentally construct a dangling pointer from a temporary
    int_pointer(int& i) : p(&i) {} 
    
    explicit operator bool() { return !!p; }
    int val() const { return *p; }
};

struct int_holder {
    int i = 0;
    bool operator==(int_pointer p) const {
        if (!p) return false;
        return i == p.val();
    }
};

TEST_CASE("hp") {
    int_holder h{8};
    int i = 8;
    
    // this comparison is safe
    // int_pointer can be implicitly constructed from this int here
    CHECK(h == i);
    // CHECK((h == i)); - this compiles just fine with extra parens
}
```

### Extra information

This issue was introduced by in https://github.com/doctest/doctest/commit/02af672f0d3093157daecb55ce4e109ffaf2bc53

The change is here (pasting link as code, otherwise github squashes it): `https://github.com/doctest/doctest/commit/02af672f0d3093157daecb55ce4e109ffaf2bc53#diff-60b065fd4481eca8e17eb467e91b3f552ab8a7fb56d58abfbaf870a98cd432c6R1154` 

I have absolutely no idea why it was made. `const R&&` seems pretty suspicious. It seems to me that it can be just removed and the code can be reverted to how it was in 3.8.6:

```c++
#define DOCTEST_DO_BINARY_EXPRESSION_COMPARISON(op, op_str, op_macro)                              \
    template <typename R>                                                                          \
    DOCTEST_NOINLINE SFINAE_OP(Result,op) operator op(R&& rhs) {                             \
    bool res = op_macro(doctest::detail::forward<const L>(lhs), doctest::detail::forward<R>(rhs));                                                             \
        if(m_at & assertType::is_false)                                                            \
            res = !res;                                                                            \
        if(!res || doctest::getContextOptions()->success)                                          \
            return Result(res, stringifyBinaryExpr(lhs, op_str, rhs));                             \
        return Result(res);                                                                        \
    }                                                                                              \
```

iboB:
## Description

Removes suspicious change introduced in https://github.com/doctest/doctest/commit/02af672f0d3093157daecb55ce4e109ffaf2bc53

I have no idea why it was made and it seems pointless to me. 

## GitHub Issues

This fixes #651


onqtam:
can confirm that this does not break the compilation for which issue #495 was originally reported for: https://godbolt.org/z/cx6xs4fT6

examples/all_features/decomposition.cpp
```diff
@@ -23,3 +23,36 @@ TEST_CASE("Move Only Type") {
...
+
+DOCTEST_MSVC_SUPPRESS_WARNING_WITH_PUSH(4866) // eval order
+TEST_CASE("Impl cast from non-const value") {
+    int_holder h{ 8 };
+    int i = 8;
+
+    // this comparison is safe
+    // int_pointer can be implicitly constructed from this int here
+    CHECK(h == i);
+    CHECK((h == i));
```
onqtam, line 56: IMHO the extra parens should be removed

examples/all_features/decomposition.cpp
```diff
@@ -23,3 +23,36 @@ TEST_CASE("Move Only Type") {
...
+    int i = 0;
+    bool operator==(int_pointer p) const {
+        if (!p) return false;
+        return i == p.val();
+    }
+};
+
+DOCTEST_MSVC_SUPPRESS_WARNING_WITH_PUSH(4866) // eval order
+TEST_CASE("Impl cast from non-const value") {
+    int_holder h{ 8 };
```
onqtam, line 50: either this
```c++
    int_holder h;
    h.i = 8;
```
or a ctor for int_holder

Saalvage:
I knew I'd seen something like this before.. const r-values references have previously plagued doctest as well!
https://github.com/doctest/doctest/commit/ce13bc44b99c87e918b98bdadb17008d2dd54d55

Saalvage:
@onqtam #495 does NOT work with this. Your godbolt link compiles using clang, while the original issue is present in gcc. But it seems like it's broken on dev as well.. Works fine on master though!

Issue #655, opened 2022-05-27
Deadlock in MpiConsoleReporter when root rank fails assert


starintheuniverse:
## Description
Thanks for a great library and especially for an MPI extension! I am developing an MPI-based scientific library, so the MPI extension is critical for my usage of doctest.

For me all of the following MPI_TEST_CASE deadlock (with mpirun -np 2):

```
#include <doctest/extensions/doctest_mpi.h>
```

```
MPI_TEST_CASE("Root fails",1) {
  CHECK( 1 == 0 );
}
```
```
MPI_TEST_CASE("Root fails",2) {
  CHECK( 1 == 0 );
}
```
```
MPI_TEST_CASE("Root fails",2) {
  CHECK( test_rank != 0 );
}
```

This seems related to a note from the initial version (#418, [extensions.md](https://github.com/BerengerBerthoul/doctest/blob/339ec122bece7c45bd2892ea4701e0b6f491fb59/doc/markdown/extensions.md)):
> If the number of processes is not enought, prints the correct message, but then deadlocks (comes from MPI_Probe in MpiConsoleReporter)

Although #611 fixed deadlock for an insufficient process count (by eliminating a code path that fails a CHECK), there remains an underlying communication bug.

### Extra information
* doctest version: **v2.4.8** (dev at 16ed548bc9)
* Operating System: **Linux Mint 20 Cinnamon 4.6.7**
* Compiler+version: **gcc 9.3.0**
* MPI standard: **MPI-3.1**
* MPI implementation: **MPICH 3.3.2**

### Problem code

`doctest/extensions/mpi_reporter.h`

I think the deadlock comes from a technical dependency of MPI_Send in `log_assert()` upon the matching MPI_Recv in `test_case_end()`, which is not called until after `log_assert()`.

This is a 'may' dependency. [MPI Standard 3.1, section 3.4](https://www.mpi-forum.org/docs/mpi-3.1/mpi31-report/node57.htm#Node57):
> The standard mode send is _non-local_: successful completion of the send operation may depend on the occurrence of a matching receive.

[...]

PR #656, opened 2022-05-27
Use MPI_Isend in MpiConsoleReporter to avoid deadlock

starintheuniverse:
## Description
Implements the solution proposed in #655 to a communication bug in MpiConsoleReporter.

### DIFF
--- a/doctest/extensions/mpi_reporter.h
+++ b/doctest/extensions/mpi_reporter.h
@@ -213,7 +213,7 @@ public:
         }
       }
 
-      MPI_Waitall(int(requests.size()), requests.data(), MPI_STATUSES_IGNORE);
+      MPI_Waitall(static_cast<int>(requests.size()), requests.data(), MPI_STATUSES_IGNORE);
       m_failure_str_queue.clear();
     }
 
--- a/doctest/extensions/mpi_reporter.h
+++ b/doctest/extensions/mpi_reporter.h
@@ -66,6 +66,7 @@ private:
       return nullStream;
     }
   }
+  std::vector<std::pair<std::string, int>> m_failure_str_queue = {};
 public:
   MpiConsoleReporter(const ContextOptions& co)
     : ConsoleReporter(co,replace_by_null_if_not_rank_0(co.cout))
@@ -154,6 +155,21 @@ public:
       int rank;
       MPI_Comm_rank(MPI_COMM_WORLD, &rank);
 
+      std::vector<MPI_Request> requests;
+      requests.reserve(m_failure_str_queue.size());  // avoid realloc & copy of MPI_Request
+      for (const std::pair<std::string, int> &failure : m_failure_str_queue)
+      {
+        const std::string & failure_str = failure.first;
+        const int failure_line = failure.second;
+
+        int failure_msg_size = static_cast<int>(failure_str.size());
+
+        requests.push_back(MPI_REQUEST_NULL);
+        MPI_Isend(failure_str.c_str(), failure_msg_size, MPI_BYTE,
+                 0, failure_line, MPI_COMM_WORLD, &requests.back()); // Tag = file line
+      }
+
+
       // Compute the number of assert with fail among all procs
       int nb_fail_asserts_glob = 0;
       MPI_Reduce(&st.numAssertsFailedCurrentTest, &nb_fail_asserts_glob, 1, MPI_INT, MPI_SUM, 0, MPI_COMM_WORLD);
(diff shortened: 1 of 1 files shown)


onqtam:
@philbucher @BerengerBerthoul

philbucher:
Cool, thx! 👍 
I can review properly beginning of next week

doctest/extensions/mpi_reporter.h
```diff
@@ -154,6 +155,21 @@ struct MpiConsoleReporter : public ConsoleReporter {
...
+      requests.reserve(m_failure_str_queue.size());  // avoid realloc & copy of MPI_Request
+      for (const std::pair<std::string, int> &failure : m_failure_str_queue)
+      {
+        const std::string & failure_str = failure.first;
+        const int failure_line = failure.second;
+
+        int failure_msg_size = static_cast<int>(failure_str.size());
+
+        requests.push_back(MPI_REQUEST_NULL);
+        MPI_Isend(failure_str.c_str(), failure_msg_size, MPI_BYTE,
```
philbucher, line 168: I think you need to change also the `MPI_Recv` to `MPI_Irecv`

doctest/extensions/mpi_reporter.h
```diff
@@ -196,6 +212,9 @@ struct MpiConsoleReporter : public ConsoleReporter {
           s << "\n";
         }
       }
+
+      MPI_Waitall(int(requests.size()), requests.data(), MPI_STATUSES_IGNORE);
```
philbucher, line 216:
```suggestion
      MPI_Waitall(static_cast<int>(requests.size()), requests.data(), MPI_STATUSES_IGNORE);
```

is the cast even needed?

doctest/extensions/mpi_reporter.h
```diff
@@ -154,6 +155,21 @@ struct MpiConsoleReporter : public ConsoleReporter {
...
       MPI_Comm_rank(MPI_COMM_WORLD, &rank);
 
+      std::vector<MPI_Request> requests;
+      requests.reserve(m_failure_str_queue.size());  // avoid realloc & copy of MPI_Request
+      for (const std::pair<std::string, int> &failure : m_failure_str_queue)
+      {
+        const std::string & failure_str = failure.first;
+        const int failure_line = failure.second;
+
+        int failure_msg_size = static_cast<int>(failure_str.size());
```
philbucher, line 165:
```suggestion
        const int failure_msg_size = static_cast<int>(failure_str.size());
```

doctest/extensions/mpi_reporter.h
```diff
@@ -66,6 +66,7 @@ struct MpiConsoleReporter : public ConsoleReporter {
       return nullStream;
     }
   }
+  std::vector<std::pair<std::string, int>> m_failure_str_queue = {};
```
philbucher, line 69:
```suggestion
  std::vector<std::pair<std::string, int>> m_failure_str_queue;
```

I think this is not needed (?)

philbucher (commented):
Thanks for the fix!
Overall the proposed solution looks good to me, you are right that we should not rely on the implementation details of some MPI implementations
`MPI_Send` is usually blocking until a matching receive is issued, I guess this was not noticed so far due to small message sizes, which were buffered

I made minor comments, mostly style, but also about the now not matching `MPI_Irecv`

BTW I checked `MPI_Bsend` (which is a buffered send), but found it to be inconvenient to use since one needs to attach a buffer manually

doctest/extensions/mpi_reporter.h
BerengerBerthoul: yes you don't need it, it will be converted anyway

BerengerBerthoul:
Thanks @starintheuniverse for the PR and @philbucher for the review. It looks good to me

doctest/extensions/mpi_reporter.h
starintheuniverse: True, but I think it is more consistent with the member declarations seen in other classes in the file.

starintheuniverse:
Thanks @philbucher for the review and also @BerengerBerthoul for taking a look!

starintheuniverse:
I just noticed a separate preexisting bug in `MpiConsoleReporter` that I'd like to fix before this gets merged: Program will hang if `WARN` is used inside an `MPI_TEST_CASE`.

It seems that `st.numAssertsFailedCurrentTest` does not count `WARN` asserts, so does not equal the number of messages logged and sent. There will be a mismatch between number of messages sent and received. An easy fix is to `MPI_Reduce` over the actual number of messages rather than the number of failed asserts. This is particularly easy now that `MPI_Isend` is being used in the same function as the `MPI_Reduce`.

philbucher:
no more comments from my side, please go ahead :)
Thanks again!

Saalvage:
Thanks for your contribution :)

PR #657, opened 2022-06-01
Avoid static init problem in insufficient_procs() (MPI)

starintheuniverse:
## Description
Reverses an incidental change that came along with #623. That PR gives a proper initial value for `world_size_before_init`. Incidentally, it also adds an evaluation of this variable possibly prior to its initialization. (The function `insufficient_procs()` is called before entering `main()`). For me it caused all MPI_TEST_CASE to be marked as skipped, even when the actual process count was sufficient. The present PR fixes this problem.

### DIFF
--- a/doctest/extensions/doctest_mpi.h
+++ b/doctest/extensions/doctest_mpi.h
@@ -93,6 +93,7 @@ namespace doctest {
 extern std::unordered_map<int,mpi_sub_comm> sub_comms_by_size;
 extern int nb_test_cases_skipped_insufficient_procs;
 extern int world_size_before_init;
+int mpi_comm_world_size();
 
 int mpi_init_thread(int argc, char *argv[], int required_thread_support);
 void mpi_finalize();
@@ -113,7 +114,7 @@ void execute_mpi_test_case(F func) {
 
 inline bool
 insufficient_procs(int test_nb_procs) {
-  bool insufficient = test_nb_procs>world_size_before_init;
+  bool insufficient = test_nb_procs>mpi_comm_world_size();
   if (insufficient) {
     ++nb_test_cases_skipped_insufficient_procs;
   }
--- a/doctest/extensions/doctest_mpi.h
+++ b/doctest/extensions/doctest_mpi.h
@@ -114,7 +114,8 @@ void execute_mpi_test_case(F func) {
 
 inline bool
 insufficient_procs(int test_nb_procs) {
-  bool insufficient = test_nb_procs>mpi_comm_world_size();
+  static const int world_size = mpi_comm_world_size();
+  bool insufficient = test_nb_procs>world_size;
   if (insufficient) {
     ++nb_test_cases_skipped_insufficient_procs;
   }


Saalvage:
@philbucher Could you take a look when you have the time?

doctest/extensions/doctest_mpi.h
```diff
@@ -113,7 +114,8 @@ void execute_mpi_test_case(F func) {
 
 inline bool
 insufficient_procs(int test_nb_procs) {
-  bool insufficient = test_nb_procs>world_size_before_init;
+  static const int world_size = mpi_comm_world_size();
```
philbucher, line 117: now `mpi_comm_world_size` is called twice, here and in line 44
Can you move the call there,? Maybe then the `static` is not even needed

philbucher (commented):
minor comment otherwise no objections, thanks for the fix!

BerengerBerthoul:
Hi,
Thanks for noticing the bug! I tested your PR on my side and I see no regression. So I am OK to merge it.

doctest/extensions/doctest_mpi.h
starintheuniverse: Thanks for the comment. I am not totally clear on what you are suggesting. But if the number of occurrences of `mpi_comm_world_size()` is the issue, I would point out it is also called from `MpiConsoleReporter::test_run_end()`. Maybe an even better solution would be to move the static variable into `mpi_comm_world_size()` itself, just before returning. What do you think?

Saalvage:
@philbucher Sorry to bother you again, but do you think you could give a last lookover to this and the other PR? I'd really like to merge them before 2.4.9 and I'd really like to release that update in the near future!

doctest/extensions/doctest_mpi.h
philbucher: I mean here:
https://github.com/doctest/doctest/blob/4d6781ee36eb31a4e17c687e2ab5c9b5ce4c222e/doctest/extensions/doctest_mpi.h#L43-L44

could you use this variable instead?
```suggestion
  static const int world_size = world_size_before_init ; // or without static ???
```

My motivation against using static is to avoid anything related to the [Static Initialization Order Fiasco](https://en.cppreference.com/w/cpp/language/siof)
I don't think it would be a big problem in the current implementation, but I thought I would bring it up

If you think it is only working with the current implementation, then please go ahead, I don't want to block anything :)

philbucher:
Sorry for the delay, I posted one more comment. I am not sure if relevant, please go ahead if necessary, it was only about a detail anyway
Thanks!

doctest/extensions/doctest_mpi.h
starintheuniverse: Ah, thanks for the clarification. Well, I think that using the global variable `world_size_before_init` as in the suggested snippet is actually what causes static initialization order fiasco in the original code.

The reason is that usage of `world_size_before_init` would occur in separate translation units from initialization. (It surprised me too.) This happens because initialization occurs in the TU where `DOCTEST_CONFIG_IMPLEMENT` is defined before `#include`-ing the header (e.g. main.cpp); whereas `insufficient_procs()` is invoked in TUs where that identifier is not defined before `#include`-ing (e.g. wherever there are `MPI_TEST_CASE`s).

The present PR is meant to avoid static initialization order fiasco, based on an example in the FAQ [Construct On First Use idiom](https://isocpp.org/wiki/faq/ctors#construct-on-first-use-v2). Maybe the caution is anticipating a _deinitialization fiasco_ that sometimes comes with static local variables, but it will not happen for a mere `int`.

Thanks a lot for your diligence keeping an eye out for these tricky problems. :)

Saalvage:
Thanks for your contribution! :)

PR #667, opened 2022-06-24
Suppress hash() clang integer sanitizer issues

bnason-nf:
The hash() functions added in the latest v2.4.9 release trigger clang integer sanitizer issues. Since the integer behavior in these functions is intentional, this change suppresses the sanitizer issues.


Saalvage:
I'm very sorry for the incredibly late response to this very straightforward PR. I've decided to add various CI sanitizers in order to prevent similar situations like this in the future!

Thank you very much for your contribution!

bnason-nf:
Thanks @Saalvage, no worries at all about the response time, there was nothing urgent. I appreciate all the work you and @onqtam have done on this library.

Issue #627, opened 2022-03-31
Feature request: streaming output via console reporter AND junit/xml report written to file


tknerr:
## Description

What I'm trying to achieve is to get immediate feedback / streaming output reported to the console, but at the same time a report in junit / xml format being written to a file.

Currently, the reporters we can choose from all write to stdout, but don't support writing to a file directly. In [reporters.md](https://github.com/doctest/doctest/blob/master/doc/markdown/reporters.md) I found that you can specify multiple reporters, e.g. `--reporters=console,junit`, however this produces only garbled output as both write to `stdout` at the same time (I hoped the second reporter would write to `stderr` for example, so we could redirect that separately, but it doesn't).

So right now, if I want both streaming output to the console, but need a report file as well, I would need to run the tests twice (once with `--reporters=console`, and then again with `--reporters=junit > report.xml`). Ideally we could have that in a single run.

Is this currently indeed not possible, or am I missing something obvious here?

onqtam:
It's indeed not possible yet. Maybe in the future, but not ETA can be promised

tknerr:
On second look, it's not really that much garbled. It's just that when using `--reporters=console,junit` the xml declaration for the junit report comes first, then the console output, and then the XML content of the junit report, e.g.:

```
$ out/bin/FooTest -d -r=console,junit
<?xml version="1.0" encoding="UTF-8"?>
[doctest] doctest version is "2.4.0"
[doctest] run with "--help" for options
===============================================================================
/home/foo/sample/src/Foo.cpp:23:
TEST CASE:  Sampe Test

===============================================================================
[doctest] test cases:      1 |      1 passed |      0 failed |      0 skipped
[doctest] assertions:      1 |      1 passed |      0 failed |
[doctest] Status: FAILURE!
<testsuites>
  <testsuite name="out/bin/FooTest" errors="0" failures="0" tests="1" time="0.000151" timestamp="2022-04-01T16:49:19Z" doctest_version="2.4.0">
    <testcase classname="/home/foo/sample/src/Foo.cpp" name="Sample Test" time="0" status="run"/>
  </testsuite>
</testsuites>
```

It's consumable from stdout still, just a bit trickier to parse / split the console output from junit xml output...

tknerr:
I tested both with 2.4.0 and 2.4.8, but behaves the same way in both cases, i.e. the XML declaration always comes first, then console output, then junit XML output.

[...]

PR #669, opened 2022-07-07
allow specifying separate outputs for reporters

srnwk:
## Description
Extends the command-line arguments to allow specifying a filter of reporters
to which the output filename applies: `--out=junit[junit-report.xml],xml[report.xml],console[log.txt]`
The old syntax `--out=filename` still works, being treated like `--out=*[filename]`.
This allows sending for example JUnit output to a file, and still having the console reporter output to the console: `-r=junit,console -o=junit[report.xml]` (since no output file has a matcher that matches the `console` reporter, it uses its default, stdout).

(Should there be an explicit way to specify stdout as well? Possibly using `-` as the filename, like many *nix tools do?)

TODO:  
- [ ] update documentation
- [ ] API for setting these from `main`
- [ ] tests, coverage

## GitHub Issues
See #627

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -3110,6 +3110,7 @@ DOCTEST_MAKE_STD_HEADERS_CLEAN_FROM_WARNINGS_ON_WALL_BEGIN
 #include <math.h>
 #endif // __BORLANDC__
 #include <new>
+#include <memory>
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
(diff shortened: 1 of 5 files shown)


nickdademo:
Very nice! Looking forward to this one ;)

beyse:
Also looking forward to it

vitaliy-ostapchuk93:
any updates on this? would be really great to use 💯

mitchgrout:

beyse:
image

denravonska:
What were the actual objection to these changes?

mitchgrout:
> What were the actual objection to these changes?

My guess is it was never actually reviewed. `doctest` went unmaintained for quite a while, so this probably landed during this period (https://github.com/doctest/doctest/issues/627#issuecomment-1178912345). It was already 3 years stale by the time I took over as the current maintainer. I'm keen to get an implementation in for v2.6 since IMO it's a bare minimum feature for a library that supports multiple output formats.

PR #680, opened 2022-08-06
Optimized code structure

jokervTv:
Dear, sir

Fixed linker error by restructuring the code when tried to use cmake to integrate doctest with mpi.

In origin file:

```cpp
#ifdef DOCTEST_CONFIG_IMPLEMENT
// declarations and definitions, like
int mpi_comm_world_size();
// skip
int mpi_comm_world_size() {
// skip
}
#else
// only declarations
int mpi_comm_world_size(); // no definition
//skip
#define MPI_CHECK // define some macro for mpi test
// ...
#endif
```

So we have to define `DOCTEST_CONFIG_IMPLEMENT` to get the function definitions and get *macro* without `DOCTEST_CONFIG_IMPLEMENT` defined and then link them.
I don't think this approach is good enough for some tests, for example if there is only one file here.
That's what I think is wrong.


mitchgrout:

PR #689, opened 2022-08-25
Use variables instead of hard-coded `mpirun` `-np`

KiruyaMomochi:
## Description

CMake's [FindMPI](https://cmake.org/cmake/help/latest/module/FindMPI.html) module sets `MPIEXEC_EXECUTABLE` and `MPIEXEC_NUMPROC_FLAG`. We use them so it works for different MPI implementations.


Saalvage:
Thank you for your contribution! (And sorry for the late response.)

PR #692, opened 2022-09-12
Fix CMake example in docs

DanielKriz:
Name of target is different from the name of executable. Therefore this example would fail in the generation phase.

## Description
This pull request fixes the example for CMake provided in the documentation.


Saalvage:
Thank you for your contribution!

PR #694, opened 2022-09-13
Fix typos in docs

tocic:


Saalvage:
Thanks for your contribution! (And sorry for the late-ish response)

PR #697, opened 2022-09-18
Restore WIN32_LEAN_AND_MEAN and NOMINMAX

falbrechtskirchinger:
doctest defines `WIN32_LEAN_AND_MEAN` and `NOMINMAX` without restoring them. As a result, a unit test designed to catch unparenthesized uses of `std::numeric_limits<T>::min()` did not find such instances.

This patch undefines `WIN32_LEAN_AND_MEAN` and `NOMINMAX` if they have been defined by doctest.


Saalvage:
Thank you for your contribution! And sorry for the late response!

PR #702, opened 2022-10-03
Fix `TEST_CASE_CLASS`

### LINKED ISSUE #608: TEST_CASE_CLASS is broken in v2.4.8
## Description
`TEST_CASE_CLASS ` works with version 2.4.7, not with v2.4.8. It is broken by commit 95a78536.

### Steps to reproduce

`TEST_CASE_CLASS ` is not documented nor tested by the CI.  However if I take [this example ](https://github.com/doctest/doctest/issues/76#issuecomment-462040669) proposed in the description of the feature, and simplify it :

```
#include "doctest/doctest.h"

class my_type {
    int data = 5;

    TEST_CASE_CLASS("all_to_all_v 2") {
        my_type local;
        CHECK(local.data == 2);
    }
};
```
This does not compile. GCC error message:
 ```
doctest.h:1980:64: error: 'DOCTEST_ANON_VAR_4' was not declared in this scope; did you mean 'DOCTEST_ANON_VAR_0'?
 1980 |     global_prefix DOCTEST_GLOBAL_NO_WARNINGS(DOCTEST_ANONYMOUS(DOCTEST_ANON_VAR_), 
```

### Extra information

Note that a (better?) alternative to `TEST_CASE_CLASS` is mentioned [here](https://github.com/doctest/doctest/issues/76#issuecomment-808603422). Maybe we should try to implement the more flexible feature rather than fixing this one.

On a personal note, I was just trying to check if `TEST_CASE_CLASS` could solve my problem (a bug in `MPI_TEST_CASE` with several `SUBCASE`). But it will not (even when fixed). So I personally don't need the fix.

* doctest version: v2.4.8
* Operating System: Linux
* Compiler+version: GCC 11

### PR DESCRIPTION
**Pull request by Saalvage**
Fixes #608 

TODO:
- [ ] Add docs

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -2141,7 +2141,7 @@ int registerReporter(const char* name, int priority, bool isReporter) {
     DOCTEST_CLANG_SUPPRESS_WARNING_POP
 
 #define DOCTEST_REGISTER_FUNCTION_CLASS(global_prefix, f, decorators)                              \
-    global_prefix DOCTEST_GLOBAL_NO_WARNINGS_CLASS(DOCTEST_ANONYMOUS(DOCTEST_ANON_VAR_), /* NOLINT */ \
+    global_prefix DOCTEST_GLOBAL_NO_WARNINGS_CLASS(DOCTEST_ANONYMOUS(DOCTEST_ANON_VAR_) = /* NOLINT */ \
             doctest::detail::regTest(                                                              \
                     doctest::detail::TestCase(                                                     \
                             f, __FILE__, __LINE__,                                                 \
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -2191,7 +2191,7 @@ int registerReporter(const char* name, int priority, bool isReporter) {
                                                   decorators)
 #else // DOCTEST_TEST_CASE_CLASS
 #define DOCTEST_TEST_CASE_CLASS(...)                                                               \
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

**Comment by avoirtech** [comment]
Hi @Saalvage - I was trying out this branch (at b280c5c1c759d7b78cb4369554bed4433e473901) and it fixes the compile problem for me - thanks.

However, noticed that some of my TEST_CASE_CLASS cases were not being run. With the following two minimal examples in the same .h file, the first runs (failing as expected), the second does not run at all:

```c++
class NonTemplated {
    int foo;
public:
    TEST_CASE_CLASS("NonTemplatedRuns") {
        NonTemplated x;
        x.foo = 3;
        CHECK(x.foo == 10);
    }
};


template<typename T>
class Templated {
    T foo;
public:
    TEST_CASE_CLASS("TemplatedDoesNotRun") {
        Templated<int> x;
        x.foo = 3;
        CHECK(x.foo == 10);
    }
};
```

I'm not sure if this is related to the recent changes, and existing issue, or something I'm doing wrong, but thought I would check here first before raising an issue. Any thoughts?

Saalvage:
@avoirtech Your template type is not being instantiated anywhere. If you explicitly instantiate the template like so: `template class Templated<int>;` it works. However, it would run the same test for every instantiation present in your code.. I'm not sure if there's anything to alleviate this..

avoirtech:
Thanks @Saalvage, that worked for me, and I've raised an issue [here ](https://github.com/doctest/doctest/issues/718) to make others aware.

mitchgrout:
Closing as stale

PR #708, opened 2022-10-04
Clean clang-tidy with `DOCTEST_CONFIG_DISABLE`

### LINKED ISSUE #691: clang-tidy gives warnings of unused variables in test cases when compiled without doctest
## Description
clang-tidy gives warnings of unused variables in test cases when compiled without doctest.

Clang-tidy uses some clever methods to omit test cases from the binary when compiled with `DOCTEST_CONFIG_DISABLE`, but the test case code is still present (not #ifdefed out) so clang-tidy still sees it, and since the `CHECK`s are compiled out, it sees unused vars.

We run clang-tidy with warnings as errors, so this causes clang-tidy failures for us.

### Steps to reproduce
```c++
TEST_CASE("foo") {
  int i = 100;
  CHECK(i == 100);
}
```

When compiled with `DOCTEST_CONFIG_DISABLE`, the test case function declares an unused variable `i`, which clang-tidy complains about. I'm not sure if anything can be done about this; doctest can insert a `// NOLINTBEGIN` but it can't insert the final `// NOLINTEND` for the same reason it can't just `#ifdef` out the whole test -- there's no marker for the test end.

### Extra information

* doctest version: **v2.4.8**
* Operating System: **MacOS 12.5.1**
* Compiler+version: **Apple clang**
   ```
   Apple clang version 13.1.6 (clang-1316.0.21.2.5)
   Target: arm64-apple-darwin21.6.0
   Thread model: posix
   InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

Saalvage:
Fixes #691


mitchgrout:
Closing as stale

Issue #684, opened 2022-08-14
subcase-filter-levels infinite loop


ChristofferGreen:
## Description
Here are two simple subclasses:

`TEST_CASE("ABC") {
    SUBCASE("A") {}
    SUBCASE("B") {}
}`

And executing 'A' with --subcase-filter-levels=1 results in an infinite loop.

### Steps to reproduce

Define the above test cases and execute with arguments: "--test-case=ABC --subcase=A --subcase-filter-levels=1"

mitchgrout:
Non-reproducible on `dev` - given the age, probably fixed

Issue #597, opened 2022-01-22
Loop-generated `SUBCASE`s are not run


yeputons:
## Description

If the same `SUBCASE` line is used for multiple root subcases in the test, only the first one is run.

### Steps to reproduce

Run the following example:

```c++
#include "doctest.h"
#include <iostream>

TEST_CASE("generated subcases") {
    std::cout << "root" << std::endl;
    for (int i = 0; i < 3; i++) {
        std::cout << "outside of subcase" << std::endl;
        SUBCASE("") {
            std::cout << "inside subcase " << i << std::endl;
        }
    }
}
```
it prints

```
root
outside of subcase
inside subcase 0
outside of subcase
outside of subcase
root
outside of subcase
outside of subcase
outside of subcase
```

Note that there is extra run of `TEST_CASE` which does not enter any `SUBCASE`s. Moreover, `i = 1` and `i = 2` are not executed at all.
 
I'd expected this example to be equivalent to

```c++
#include "doctest.h"
#include <iostream>

TEST_CASE("generated subcases") {
    std::cout << "root" << std::endl;
    std::cout << "outside of subcase" << std::endl;
    SUBCASE("") {
        std::cout << "inside subcase " << 0 << std::endl;
    }
    std::cout << "outside of subcase" << std::endl;
    SUBCASE("") {
        std::cout << "inside subcase " << 1 << std::endl;
    }
    std::cout << "outside of subcase" << std::endl;
    SUBCASE("") {
        std::cout << "inside subcase " << 2 << std::endl;
    }
}
```

which prints

```
root
outside of subcase
inside subcase 0
outside of subcase
outside of subcase
root
outside of subcase
outside of subcase
inside subcase 1
outside of subcase
root
outside of subcase
outside of subcase
outside of subcase
inside subcase 2
```

### Extra information
It seems to me that `SUBCASE` actually remember its `__FILE__` and `__LINE__`, and presumably uses this information to check whether or not a subcase has been executed. However, it's possible to do without this information, purely control-based approach.

Possible usecases are very similar to parameterized tests. For example, I use it to test behavior of smart pointer implementations in school assignments, like this:

```c++
TEST_CASE("unique_ptr reset") {
    for (int orig_non_empty = 0; orig_non_empty < 2; orig_non_empty++) {
        INFO("orig_non_empty is " << orig_non_empty);

        SUBCASE("") {
            unique_ptr<int> p(orig_non_empty == 1 ? new int(10) : nullptr);
[...]

PR #725, opened 2022-11-24
Fix SUBCASE infinite loop

AlexanderHederstaf:
* Insert and check same hash for fullyTraversedSubcases

## Description

Change the saved hash to match the hash later checked in the Subcase constructor.
This fixes an infinite loop that I encountered with at least two different nested SUBCASEs inside a loop.

```
TEST_CASE("Infinite")
{
  for (int idx = 0; idx < 2; ++idx)
  {
      SUBCASE("A")
      {
          SUBCASE("SUB A")
          {
              std::cout << "sub a" << std::endl;
          }
      }
      SUBCASE("B")
      {
          SUBCASE("SUB B")
          {
              std::cout << "sub b" << std::endl;
          }
      }
  }
}
```

### DIFF
--- a/doctest/parts/doctest.cpp
+++ b/doctest/parts/doctest.cpp
@@ -1148,12 +1148,12 @@ namespace detail {
 
             if (!g_cs->reachedLeaf) {
                 // Leaf.
-                g_cs->fullyTraversedSubcases.insert(hash(g_cs->subcaseStack));
+                g_cs->fullyTraversedSubcases.insert(hash(hash(g_cs->subcaseStack, g_cs->currentSubcaseDepth), hash(m_signature)));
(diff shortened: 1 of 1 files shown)


AlexanderHederstaf:
I encountered this issue when bumping from 2.4.6 to 2.4.9. The impacted test did not have all its asserts trigger previously, related to https://github.com/doctest/doctest/issues/597. After the change in this PR all assertions are triggered and the test does not get stuck in an infinite loop.

AlexanderHederstaf:
Might be related to https://github.com/doctest/doctest/issues/684.

mitchgrout:
Stale, but I'm particularly interested in getting this fixed. Are you still interested? If not (or I don't get a response in a few days), I'll close and take over this change.

PR #734, opened 2023-01-12
Fix #733

### LINKED ISSUE #733: Help needed with the builds
Currently, one of the reasons version 2.4.10 is being delayed is the failing builds in the `dev` branch

Here's a recent run of GitHub Actions on another branch from a few days ago:
https://github.com/doctest/doctest/actions/runs/3855883199

Most of the builds fail at the install step - need help with that. This is the build config:
https://github.com/doctest/doctest/blob/dev/.github/workflows/main.yml

The ones that fail at the build step due to C++-ish issues will be fixed after that.

flagarde:
This should fix the problem :  `The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 40976EAF437D05B5 NO_PUBKEY 3B4FE6ACC0B21F32`

## GitHub Issues
#733

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -67,7 +67,7 @@ jobs:
   sanitizers:
     if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
     runs-on: ubuntu-22.04
-    
+
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


onqtam:
Thank you and sorry for the huge delay!

flagarde:
@onqtam Yes, I don't know how to fix this, it seems the depo doesn't have some package too ? For this reason I created my own docker https://github.com/flagarde/ci, maybe you could create your too or use them (but they are very minimal in term of installed package) https://hub.docker.com/u/flagarde

Issue #722, opened 2022-10-26
doctest.h -- error: call of overloaded ‘log10(unsigned int)’ is ambiguous


nbriggs:
## Description
Using doctest 2.4.9, as packaged with ccache, compiling with gcc on solaris (see versions below) results in:

```
In file included from /export/home/briggs/ccache-4.7.1/unittest/main.cpp:27:
/export/home/briggs/ccache-4.7.1/src/third_party/doctest.h: In member function ‘virtual void doctest::{anonymous}::ConsoleReporter::test_run_end(const doctest::TestRunStats&)’:
/export/home/briggs/ccache-4.7.1/src/third_party/doctest.h:6191:48: error: call of overloaded ‘log10(unsigned int)’ is ambiguous
 6191 |             auto totwidth = int(std::ceil(log10((std::max(p.numTestCasesPassingFilters, static_cast<unsigned>(p.numAsserts))) + 1)));
      |                                           ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/gcc/11/lib/gcc/sparcv9-sun-solaris2.11/11.2.0/include-fixed/math.h:24,
                 from /usr/gcc/11/include/c++/11.2.0/bits/std_abs.h:40,
                 from /usr/gcc/11/include/c++/11.2.0/cstdlib:77,
                 from /usr/gcc/11/include/c++/11.2.0/ext/string_conversions.h:41,
                 from /usr/gcc/11/include/c++/11.2.0/bits/basic_string.h:6607,
                 from /usr/gcc/11/include/c++/11.2.0/string:55,
                 from /export/home/briggs/ccache-4.7.1/src/Stat.hpp:29,
                 from /export/home/briggs/ccache-4.7.1/unittest/../src/Util.hpp:21,
                 from /export/home/briggs/ccache-4.7.1/unittest/main.cpp:19:
/usr/gcc/11/lib/gcc/sparcv9-sun-solaris2.11/11.2.0/include-fixed/iso/math_iso.h:209:28: note: candidate: ‘long double std::log10(long double)’
  209 |         inline long double log10(long double __X) { return __log10l(__X); }
      |                            ^~~~~
[...]

PR #747, opened 2023-03-10
Addresses  #722 doctest.h -- error: call of overloaded ‘log10(unsigned int)…

navinp0304:
#722

Typecast first argument with static_cast\<double\>

## Description

This is because in /usr/include/c++/12/cmath you have this line
```
  template<typename _Tp>
    inline _GLIBCXX_CONSTEXPR
    typename __gnu_cxx::__enable_if<__is_integer<_Tp>::__value,
                                    double>::__type
    log10(_Tp __x)
    { return __builtin_log10(__x); }
```
Probably in solaris you don't have that line. I don't have a solaris system to check. But commenting out the above lines gives a error message similar to yours.

An easy fix in doctest would be to change the first arg of log10 to double by static_cast since log10 expect double and the above snipped intrinsically does that.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,6 +53,10 @@ else()
     target_include_directories(${PROJECT_NAME} INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/>)
 endif()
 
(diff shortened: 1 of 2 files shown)


navinp0304:
Closing this to create  a new one.

PR #749, opened 2023-03-10
Address issue Help needed with the builds 733 and 722

navinp0304:
Address issue Help needed with the builds #733 and #722

## Description

Add gdwarf-4 compile flag for valgrind checks and static_cast for log10


navinp0304:
Missed only for clang.

PR #750, opened 2023-03-10
Clang valgrind error Fix 733 and 722

navinp0304:
Fix valgrind errors in clang and issue #733 and #722


navinp0304:
Missed changes

PR #768, opened 2023-04-11
Suppress Clang 16 warnings. Closes #766

### LINKED ISSUE #766: Tests don't compile on Clang 16 due to memory safety warnings with `-Werror`
## Description

I try to run CMake for this library, which I assume just builds the examples.

### Steps to reproduce

```sh
mkdir build
cd build
cmake .. -DCMAKE_CXX_COMPILER=clang++-16
make -k
```
This gave me a bunch of errors like this one:
```sh
[  0%] Building CXX object examples/all_features/CMakeFiles/all_features.dir/main.cpp.o
In file included from /home/holyblackcat/Sandbox/2/doctest-2.4.11/examples/all_features/main.cpp:2:
In file included from /home/holyblackcat/Sandbox/2/doctest-2.4.11/scripts/development_only/doctest/doctest.h:1:
/home/holyblackcat/Sandbox/2/doctest-2.4.11/doctest/parts/doctest_fwd.h:608:47: error: unsafe buffer access [-Werror,-Wunsafe-buffer-usage]
    bool isOnStack() const noexcept { return (buf[last] & 128) == 0; }
```

### Extra information

* Operating System: **Ubuntu 22.04**
* Compiler+version: **Clang 16 from https://apt.llvm.org/**

HolyBlackCat:
## Description

**What:** Silence Clang 16's `-Wunsafe-buffer-usage`.
**Why:** That + `-Werror` causes build errors in the examples.

## GitHub Issues

https://github.com/doctest/doctest/issues/766

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -212,6 +212,10 @@ jobs:
             compiler: clang
             version: "15"
 
+          - os: ubuntu-latest
+            compiler: clang
+            version: "16"
+
           - os: macOS-10.15
             compiler: xcode
             version: "10.3"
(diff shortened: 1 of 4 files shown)


onqtam:
if the warning suppression is not recognized by most clang versions it can be made conditional with ifdefs only for clang 16 and if it happens only in 1-2 places it can also be made locally there. Also seems that clang 16 is harder to add to the build matrix

HolyBlackCat:
@onqtam I think I've fixed it.

But some unrelated CI jobs are stuck as 'queued' me. :thinking:

mitchgrout:
Closing as implemented in #953

Issue #711, opened 2022-10-08
Tests fail under alpha


matoro:
## Description

I'm attempting to validate this package on alpha for Gentoo, but I'm getting (inconsistent) failures in the test suite related to the timeout decorator.  Usually somewhere between 5 and 15 of the test cases will fail, the failures all look like this:

```
35/155 Testing: test_cases_and_suites.cpp_xml
35/155 Test: test_cases_and_suites.cpp_xml
Command: "/usr/bin/cmake" "-DCOMMAND=/var/tmp/portage/dev-cpp/doctest-2.4.9/work/doctest-2.4.9_build/examples/all_features/all_features --no-skipped-summary --no-version -sf=*test_cases_and_suites.cpp --reporters=xml --dt-no-path-filenames=1 --dt-no-line-numbers=1 --dt-no-exitcode=1 --dt-gnu-file-line=0 --dt-no-time-in-output=1" "-DTEST_OUTPUT_FILE=/var/tmp/portage/dev-cpp/doctest-2.4.9/work/doctest-2.4.9/examples/all_features/test_output/test_cases_and_suites.cpp_xml.txt" "-DTEST_TEMP_FILE=/var/tmp/portage/dev-cpp/doctest-2.4.9/work/doctest-2.4.9_build/examples/all_features/CMakeFiles/temp_test_output_test_cases_and_suites.cpp_xml.txt" "-DTEST_MODE=COMPARE" "-P" "/var/tmp/portage/dev-cpp/doctest-2.4.9/work/doctest-2.4.9/scripts/cmake/exec_test.cmake"
Directory: /var/tmp/portage/dev-cpp/doctest-2.4.9/work/doctest-2.4.9_build/examples/all_features
"test_cases_and_suites.cpp_xml" start time: Oct 03 04:09 UTC
Output:
----------------------------------------------------------
-- Found Git: /usr/bin/git (found version "2.37.3") 
diff --git a/var/tmp/portage/dev-cpp/doctest-2.4.9/work/doctest-2.4.9/examples/all_features/test_output/test_cases_and_suites.cpp_xml.txt b/var/tmp/portage/dev-cpp/doctest-2.4.9/work/doctest-2.4.9_build/examples/all_features/CMakeFiles/temp_test_output_test_cases_and_suites.cpp_xml.txt
index 6fc245a..e9d1e79 100644
[...]

PR #770, opened 2023-04-15
Add sleep of one clock-resolution to test which should trigger timeout

matoro:
Fixes https://github.com/doctest/doctest/issues/711 under HPPA, and partially under alpha.


examples/all_features/test_cases_and_suites.cpp
```diff
@@ -46,6 +52,11 @@ TEST_CASE_FIXTURE(SomeFixture, "fixtured test - not part of a test suite") {
 
 TEST_CASE("normal test in a test suite from a decorator" * doctest::test_suite("ts1") *
           doctest::timeout(0.000001)) {
+#ifndef _WIN32
+    struct timespec res{};
+    clock_getres(CLOCK_MONOTONIC, &res);
+    std::this_thread::sleep_for(std::chrono::nanoseconds(res.tv_nsec));
+#endif
```
mitchgrout, line 59: Why not just `this_thread::sleep_for( /*1e-6 equivalent */)`?

examples/all_features/CMakeLists.txt
```diff
@@ -124,6 +130,12 @@ function(add_test_all_features test_name flags)
     add_executable(${test_name} ${files_with_output})
     target_compile_definitions(${test_name} PRIVATE ${flags})
     target_link_libraries(${test_name} doctest ${CMAKE_THREAD_LIBS_INIT})
+    if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+        target_link_options(${test_name} PRIVATE "-rtlib=compiler-rt")
+        if(LINUX)
+            target_link_libraries(${test_name} "-lgcc_s")
+        endif()
+    endif()
```
mitchgrout, line 138: Reason for adding these?

mitchgrout (commented):
Just clearing out old PRs, you still interested in submitting this change?

examples/all_features/CMakeLists.txt
matoro: Honestly I do not remember, clang does not support alpha or hppa so it cannot be that.  Probably OK to remove.

examples/all_features/test_cases_and_suites.cpp
matoro: Because in order to guarantee it isn't rounded to zero, it needs at least one resolution cycle, which on hppa is 4000000 nanoseconds.  Not sure what you mean where `1e-6` comes from.

mitchgrout: `1e-6` coming from the value provided to the `timeout` decorator, but good point about the need for it to be at least one cycle. 

It's a little awkward that this test doesn't have any handling for Win32 though; presumably your case is only running for Linux, so it doesn't affect you, but it would be nice if the test was: 

```cpp
SCENARIO(... * timeout(quanta)) {
  SleepAtLeast(quanta);
  /* Boom! */
}
```

...for all platforms.

matoro: > `1e-6` coming from the value provided to the `timeout` decorator, but good point about the need for it to be at least one cycle.
> 
> It's a little awkward that this test doesn't have any handling for Win32 though; presumably your case is only running for Linux, so it doesn't affect you, but it would be nice if the test was:
> 
> ```c++
> SCENARIO(... * timeout(quanta)) {
>   SleepAtLeast(quanta);
>   /* Boom! */
> }
> ```
> 
> ...for all platforms.

Do you mean this should have windows-specific code?  From my testing above, unix-specific code is necessary to determine the quanta at all.

mitchgrout: Ah no, mostly just lift the impl. out into some helper, in which we can deal with the platform-specific stuff. Just de-clutters the test-case and makes the intent clearer. If we _can_ also have the equivalent WIN32 code that would be nice, but it's not strictly required.

mitchgrout:
Closing for now as issue _should_ be resolved in v2.5.0. Please re-open if it continues to be an issue

Issue #687, opened 2022-08-21
Submit to Bazel Central Registry


zaucy:
I saw some minimal bazel support was added which is great! Would the maintainers here be interested in submitting to the [BCR](https://github.com/bazelbuild/bazel-central-registry)? It would involve adding a `MODULE.bazel` file to the repository with some details and adding some tests that can be executed with `bazel test`.

Saalvage:
Sorry for the late response! If it allows doctest to get into the hands of more people then I'm all for it! Would you be able to craft up a PR for this?

zaucy:
I meant to work on this a long time ago. I created a PR for it now https://github.com/doctest/doctest/pull/797 😅

mitchgrout:
Pending submission to BCR, will leave open until then

mitchgrout:
https://github.com/doctest/doctest/actions/runs/23644627797

PR #797, opened 2023-08-28
added bzlmod support

zaucy:
## Description

This adds support for the latest bazel MODULE system (bzlmod) and refactors the older minimal bazel support. Consuming `doctest` in a bazel project becomes as simple as adding this to a users `MODULE.bazel` file:

```python
bazel_dep(name = "doctest", version = "2.4.11")
```

Additionally I added GitHub actions to help with bazel integration. I could only get the 1 example to work correctly, but I think that example would be the most common use case when using bazel.

## GitHub Issues

prepares for #687, but does not close it until actually submitted to the BCR


scripts/update_stuff.py
```diff
@@ -42,3 +42,17 @@ def getVersionTuple(v):
...
+bzlmod_contents = ""
+for line in fileinput.input(["../MODULE.bazel"]):
+    if line.startswith("    version = "):
+        bzlmod_contents += "    version = \"" + version + "\",\n"
+    else:
+        bzlmod_contents += line
+
+bzlmod = open("../MODULE.bazel", "w")
+bzlmod.write(bzlmod_contents)
+bzlmod.close()
```
zaucy, line 58: I didn't test this, but it's almost an exact copy of the meson.build above!

zaucy:
The codecov report seems to be inaccurate. It's complaining about a file that I haven't touched 😅

mitchgrout:
Just clearing out old PRs, are you still interested in submitting this change?

zaucy:
yes im still interested. I can update this PR further since its been awhile since I've posted this.

examples/exe_with_static_libs/BUILD.bazel
```diff
@@ -0,0 +1,48 @@
...
+[expand_template(
+    name = src.replace(".", "_"),
+    out = "_" + src,
+    # doctest is at the root and is generally consumed with angle bracket
+    # includes, but with bazel we cannot do includes = ["."] as that is not
+    # allowed. To workaround this we replace the includes to use quote includes
+    # just for bazels sake.
+    # SEE: https://github.com/bazelbuild/bazel/issues/18913#issuecomment-1688357985
+    substitutions = {
+        "<doctest/doctest.h>": "\"doctest/doctest.h\"",
```
zaucy, line 20: I couldn't figure out a way to get `#include <doctest/doctest.h>` to work with bazel, so all bazel users will have to use `#include "doctest/doctest.h"` instead.

BUILD.bazel
```diff
@@ -0,0 +1,7 @@
+load("@rules_cc//cc:defs.bzl", "cc_library")
+
+cc_library(
+    name = "doctest",
+    textual_hdrs = glob(["doctest/**/*.h"]),
```
mitchgrout: Should exclude `parts/` for the time being

BUILD.bazel
mitchgrout, line 1: Could also consider having a non-monolithic library, see CMake's `doctest_with_main`. Will be doing some work around there to make this a viable target for consumers

MODULE.bazel
```diff
@@ -0,0 +1,7 @@
+module(
+    name = "doctest",
+    version = "2.4.12",
+)
+
+bazel_dep(name = "rules_cc", version = "0.2.16")
+bazel_dep(name = "bazel_skylib", version = "1.9.0")
```
mitchgrout: Just for my understanding, what are these for?

.gitignore
```diff
@@ -73,3 +72,8 @@ CTestCostData.txt
 LastTest.log
 LastTestsFailed.log
 temp
+
+# Bazel
+/bazel-*
+/examples/bazel-*
```
mitchgrout: `bazel-*` would cover both

.github/workflows/bazel.yml
mitchgrout, line 1: Make sure this aligns with all the current CI/CD configs (e.g. `push` should be restricted to branches, jobs should be auto-cancellable).

Don't forget final newline on file.

.github/workflows/release.yml
mitchgrout, line 1: This doesn't sound particularly `bazel`-specific. Probably should just be `.github/workflows/release.yml`

.bazelignore
mitchgrout, line 1: Do we actually need this?

.bazelrc
mitchgrout, line 1: Just checking, this is basically just saying, "add these options to these build targets"?

MODULE.bazel.lock
mitchgrout, line 1: Preferably we avoid lockfiles unless there's an absolute need for them

examples/MODULE.bazel
```diff
@@ -0,0 +1,9 @@
+module(name = "doctest_examples")
+
+bazel_dep(name = "rules_cc", version = "0.2.16")
+bazel_dep(name = "bazel_skylib", version = "1.9.0")
```
mitchgrout, line 4: Duplicated from top-level `BUILD.bazel`?

examples/MODULE.bazel
mitchgrout, line 1: I'm not 100% on actually running the `examples` across different build-systems. Unless we have a real good reason to, probably just skip it.

examples/MODULE.bazel.lock
mitchgrout, line 1: Ditto regarding lockfiles

examples/WORKSPACE.bazel
mitchgrout, line 1: I suspect this isn't meant to be here, project should only have one `WORKSPACE` file afaik

scripts/update_stuff.py
mitchgrout: Preferably test it, just bump `version.txt` to `2.5.0` or something and see if it works.

mitchgrout (changes requested):

examples/MODULE.bazel
zaucy: the second workspace/module is to test doctest as a consumer which serves two purposes:
1. catches edge case error when consuming `@doctest` as a user
2. gives an example on how you might consume `@doctest` with bazel as a user

but im also happy to remove the secondaryh workspace/module if you'd prefer

.bazelrc
zaucy: Not quite, its add these flags to `bazel` whenever you run `bazel`. It's a list of subcommands and a optional config name. `:ci` is the `ci` config, which I used in the github actions (used with `--config=ci`) just to help debug any issues and give all errors instead of just the first found.

This file doesn't affect downstream users

MODULE.bazel
zaucy: `bazel_skylib` isn't needed in the root module. I removed it.

`rules_cc` is required when using c/c++ rules such as `cc_library`, `cc_test`, etc.

`bazel_skylib` is still required for the examples to rewrite includes: https://github.com/doctest/doctest/pull/797#discussion_r2658359818

.github/workflows/release.yml
zaucy: That's fair. I only called it that due to bazel/bcr recommending a stable archive be available. I'll rename it.

.bazelignore
zaucy: Only if we want `bazel build ...` (build all) to work without issues when having the `examples` module.

Technically only the `examples` line is needed. The rest is just for hiding extra folders from bazel.

examples/MODULE.bazel
zaucy: same reason as https://github.com/doctest/doctest/pull/797#discussion_r2658476809

examples/WORKSPACE.bazel
zaucy: same reason as https://github.com/doctest/doctest/pull/797#discussion_r2658476809

scripts/update_stuff.py
zaucy: seems to work. I did notice that if there's a newline in `version.txt` it breaks, but that's all the updates.

.github/workflows/release.yml
mitchgrout, line 1: So just to confirm, we need to:

1. Produce a tag (e.g. v2.5.0)
2. Create and publish a release on that tag

...which will trigger this, which creates and uploads a `.tar.gz` to said release. And, the bazel architecture takes care of the rest?

examples/exe_with_static_libs/BUILD.bazel
mitchgrout: Will leave it as-is for now; I'd like to do a little re-organization (mostly to resolve the inconsistency with project vs global-install inclusion), so we can shuffle this into a `single-header/` directory; presumably having the include _not_ be `["."]` solves the problem?

.github/workflows/release.yml
zaucy: yep!

PR #804, opened 2023-09-07
Only call subcase_end currentSubcaseDepth times

### LINKED ISSUE #803: Combining subcases and threads can put doctest into an infinite loop
TEST_CASE("crash") {
        SUBCASE("sub") { CHECK(true); }
        std::thread([] { REQUIRE(false); });
    };

This test case causes doctest to try and print out its subcase stack 2^64-1 times. The same thing happens with any exception, segfault etc in the sub thread, not just REQUIRE. The problem is that reportFault calls subcase_end according to the stack depth, not the current stack location, so currentSubcaseLevel gets decremented below 0. Then the logTestStart attempts to print that many names from the subcase stack, with sad results.

patstew:
Fix #803, presumably this has been broken since #598.

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -4885,10 +4885,10 @@ namespace {
 
         DOCTEST_ITERATE_THROUGH_REPORTERS(test_case_exception, {message.c_str(), true});
 
-        while (g_cs->subcaseStack.size()) {
-            g_cs->subcaseStack.pop_back();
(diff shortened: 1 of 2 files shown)


mitchgrout:
Just clearing out old PRs, are you still interested in submitting this change?

patstew:
Yes, but there's a duplicate bug and PR now #915 #919 that's essentially the same so whichever.

mitchgrout:
Closed in favour of #919

PR #810, opened 2023-09-20
easy cmake configuration in build

astrowild99:
## Description
A simpler configuration for the `CMakeLists` described in the documentation, as the import using libraries as described in the current doc is not the simplest, more beginner friendly way to get the project started


mitchgrout:
Closed by #866

PR #817, opened 2023-09-30
Expose include directory as system path via CMake

severinstrobl:
## Description
Treading the include directory as a system include directory prevents clang-tidy from emitting errors from the doctest headers, avoiding issues when using various doctest macros.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,7 +50,7 @@ if(MAIN_PROJECT)
 
     add_custom_target(assemble_single_header ALL DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/doctest/doctest.h)
 else()
-    target_include_directories(${PROJECT_NAME} INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/>)
+    target_include_directories(${PROJECT_NAME} SYSTEM INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/>)
 endif()
 
 if(CMAKE_BUILD_TYPE STREQUAL "Debug")


mitchgrout:
Bumping this, still interested in making this change? 

If so, would also be good to consider alternate build-systems. For example, `meson` has [`include_directories(is_system: bool)`](https://mesonbuild.com/Reference-manual_functions.html#include_directories_is_system) which would be beneficial to those users (myself included)

BUILD.bazel
```diff
@@ -6,5 +6,6 @@ cc_library(
         ["doctest/**/*.h"],
         exclude = ["doctest/parts/**/*"],
     ),
+    includes = ["doctest"],
```
severinstrobl, line 9: Not sure if this is the best approach for Bazel...

meson.build
```diff
@@ -1,6 +1,6 @@
 project('doctest', ['cpp'], version: '2.4.12')
 
-doctest_dep = declare_dependency(include_directories: include_directories('doctest'))
+doctest_dep = declare_dependency(include_directories: include_directories('doctest', is_system: true))
```
severinstrobl, line 3: ... or for Meson.

mitchgrout: I reckon it's fine; they also use it in `mesonbuild/wrapdb` for `gtest`: 
https://github.com/mesonbuild/wrapdb/blob/528def7a7ad993771f7b46934c71e191050424a5/subprojects/packagefiles/gtest/googletest/meson.build#L13

BUILD.bazel
mitchgrout: I'm no bazel user, so can't say for sure; @zaucy any input?

zaucy: I believe this would make the include `<doctest.h>` when the official include path is `<doctest/doctest.h>`. I didn't want bazel to be too different than other build systems.

Unfortunately `includes = ["."]` is not allowed so we're stuck with `"doctest/doctest.h"`. I would prefer to be as close as possible to what is considered the official/standard include path.

severinstrobl: Then I'd say we keep the include path as-as for Bazel, having a change in the way _doctest_ is included is not worth the potential benefit.
@mitchgrout: WDYT?

mitchgrout: Changing include from `doctest/doctest.h` to `doctest.h` is a no-go, so we'll hold off on that for now. Is there no other way to force `doctest` to be included via `-isystem`? 

If not, we can always just make a note of it in `docs/build-systems.md` somewhere. Getting 2/3 done is worth it regardless.

zaucy: This is the workaround to make the examples work with the system include paths: https://github.com/doctest/doctest/blob/dev/examples/exe_with_static_libs/BUILD.bazel#L11-L23

Since `includes = ["."]` is not allowed there are only workarounds. There are a couple ways to get it as a system include:
1. Move doctest to its own dedicated folder such as `include` so we can do `includes = ["include"]`, but that's very distruptive for people using this upstream. I'd imagine we wouldn't want to do that. ❌ 
2. Copying the files with `bazel_skylib` to an include directory that bazel is aware of and doing `includes = ["the_generated_include_folder"]`. I actually did try this in my original PR, but it didn't quite work. ⚠️ 

Related issue https://github.com/bazelbuild/bazel/issues/18913#issuecomment-1688357985

I think having the include as `"doctest/doctest.h"` for bazel and mentioning it in the docs is the right play.

Issue #786, opened 2023-06-21
Using CHECK with bitfield fails to compile


usefulcat:
Using CHECK to test the value of a bitfield doesn't compile (Linux, gcc 12, c++20, doctest 2.4.11):

```
struct S {
    uint32_t a : 31;
    uint32_t b : 1;
};
TEST_CASE("bitfield") {
    S s;
    s.a = 0;
    s.b = 1;
    CHECK(s.a == 0); // error here
}
```
lib_1_src1.cpp:18:13: error: cannot bind bitfield ‘s.S::a’ to ‘unsigned int&’
   18 |     CHECK(s.a == 0);

Note that this use case works fine with Catch(2).

Edit:

Looks like the same problem also affects packed structs:
```
struct __attribute__((packed)) S {
    int i;
};
TEST_CASE("packed_struct") {
    S s;
    s.i = 0;
    CHECK(s.i == 0); // error: cannot bind packed field ‘s.S::i’ to ‘int&’
    //CHECK(int(s.i) == 0); // 'useless cast'
}
```
For bitfields, the error can be avoided by casting the member to its underlying type (e.g. CHECK(uint32_t(s.a) == 0)). But that doesn't work for packed structs, since it triggers a warning about a useless cast.

For me, not supporting bitfields is not too bad, as I don't have a lot of them and there is a possible workaround, but I do have a lot of packed structs.

hdoc:
Can confirm that @navinp0304's changes make it possible to use the `CHECK()` macro for bitfields in our application.

PR #825, opened 2023-11-08

navinp0304:
Use rvalue reference instead of universal reference so that bitfields bind to const lvalue


navinp0304:
missed win32

PR #826, opened 2023-11-08

navinp0304:
Using rvalue references instead of universal references
Removing win32 related changes


navinp0304:
missed pragma pack

Issue #630, opened 2022-04-08
issue introduced in 2.4.7


onqtam:
How to reproduce:
https://godbolt.org/z/5Y87aMW8W

fails with gcc but no problem with clang - doctest version 2.4.7 introduced the problem and it's because of this commit:
https://github.com/doctest/doctest/commit/20dbad670e1a6ebf52d3aa12d55c7784070a115d

Seems to be related to this member and/or the lack of a move constructor for the `stringstream_proxy` class
https://github.com/adishavit/argh/blob/master/argh.h#L67

Logging this to not forget about it - I haven't dug any deeper though.

Original report: https://github.com/adishavit/argh/issues/72

Tradias:
Thank you for writing a reproducer, I was trying to write one as I am also facing this issue using doctest+Boost.Outcome. A fix would be welcome, currently stuck on doctest 2.4.6 because of it.

Saalvage:
@Tradias Could you please confirm that the PR solves this completely?

Tradias:
@Saalvage yes it solves it. Thank you for the fix!

Saalvage:
Fixed in dev!

PR #827, opened 2023-11-08

navinp0304:
Change universal reference to rvalue reference by adding const

### DIFF
--- a/doctest/parts/doctest_fwd.h
+++ b/doctest/parts/doctest_fwd.h
@@ -1575,8 +1575,8 @@ DOCTEST_CLANG_SUPPRESS_WARNING_POP
         // https://github.com/catchorg/Catch2/issues/870
         // https://github.com/catchorg/Catch2/issues/565
         template <typename L>
-        Expression_lhs<L> operator<<(L&& operand) {
-            return Expression_lhs<L>(static_cast<L&&>(operand), m_at);
+        Expression_lhs<const L&&> operator<<(const L&& operand) { //bitfields bind to universal ref but not const rvalue ref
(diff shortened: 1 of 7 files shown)


navinp0304:
@onqtam can you please merge this ?

onqtam:
This part of the code is the most tricky of all... Just did a git blame and it seems that this commit was the last modification for them: ce13bc44b99c87e918b98bdadb17008d2dd54d55

Your proposed changes are not a direct revert of that commit - there are some differences.

I'm no longer in a position to evaluate such changes - I haven't touched C++ in 2-3 years in any meaningful way. I've moved on completely.

This project is looking for maintainers again - as noted here: https://github.com/doctest/doctest/issues/554#issuecomment-1740568920

The amount of work is very little - I'd argue that someone capable should be able to spend no more than 1-2 hours a month to keep up with PRs & issues - there aren't that many of them.

I can't merge this PR at this time - I can't evaluate it properly and am not sure if there are sufficient tests to catch any regressions.

navinp0304:
@Saalvage can you please review and merge this ?

onqtam:
@navinp0304 if you manage to reproduce the problems from before commit ce13bc44b99c87e918b98bdadb17008d2dd54d55 (which it was aimed at addressing) and then confirm that they're no longer present after your current patch and just say so - I'll merge this.

navinp0304:
https://github.com/doctest/doctest/issues/630 is fixed by the current merge. I tried running https://godbolt.org/z/5Y87aMW8W and it passed.

onqtam:
Sorry for taking so long - I wanted to look at the build failures and deal with those first, but I can't get myself to do even that - so I'm merging this as it is.

PR #830, opened 2023-11-21
Allow constructing doctest::String from anything with .data() and .size() members

nabijaczleweli:
## Description
Most obvious of these are std::string and std::string_view, and this is naturally useful for stringification implementations which otherwise need to extract the size explicitly.


nabijaczleweli:
never-mind, too annoying to do correctly

PR #835, opened 2023-12-09
Add Bazel module file to support Bzlmod

Vertexwahn:
Add Bazel module file to support Bzlmod


mitchgrout:
Related to #797, may close in favour-of depending on whichever becomes more active.

mitchgrout:
Closed by #797

PR #838, opened 2024-01-08
Bump cmake version to fix deprecation warning

0xE1E10:
Simple version bump to fix this warning:
```
CMake Deprecation Warning at build/_deps/doctest-src/CMakeLists.txt:1 (cmake_minimum_required):Compatibility with CMake < 3.5 will be removed from a future version of
CMake.

Update the VERSION argument <min> value or use a ...<max> suffix to tell
CMake that the project does not need compatibility with older versions.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.5)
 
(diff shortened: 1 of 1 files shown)


onqtam:
this https://github.com/doctest/doctest/pull/811 got merged in the dev branch a few days ago

PR #858, opened 2024-06-19
[feature] Add option to redirect output (`--out`) to `cout<<`/`cerr<<`

helmesjo:
## Description
Default output stream is `std::cout`. This PR still defaults to `std::cout` but makes it possible to chose between `std::cout` & `std::cerr` through the command line.

Default behavior remains unchanged.


mitchgrout:
Just wanting some extra context here. The current default is `std::cout` as you said, with the option allowing us to select a filename to send to.

1. Why do you need to be able to select `std::cout` when it's the default?
2. Why do you want to be able to emit to `std::cerr`?
3. Why could you not utilize shell-redirection to achieve this? e.g. `./test 1>&2` to send `cout` to `cerr`

helmesjo:
It's been a minute, but if I were to guess I think it could've been related to controlling output between multiple processes (eg. when run on CI). I honestly can't recall the exact motivation for this PR though. 😇

mitchgrout:
Yeah, use-case could be to do something like send JUnit or something to `stdout`, although it feels particularly niche. For now, I suggest we close (can always re-open later if there's a reinvigorated desire for it!). There's also #627 which feels similarly motivated.

PR #837, opened 2023-12-19
CMake: install doctest_with_main as a target if built

tchernobog:
## Description

This change allows installing and exporting the static library target doctest::doctest_with_main if the project is configured with DOCTEST_WITH_MAIN_IN_STATIC_LIB.

This is required for usage in third-party projects that install doctest at the package-manager level.

PR #862, opened 2024-07-29
DoctestWithMain Installation

pinwhell:
## Description
Exporting doctest_with_main as a library, this was added for the reason that once installing the library i didn't found a way to link against doctest_with_main seems it wasn't exported.


mitchgrout:
Seems related to #837

Issue #698, opened 2022-09-22
Update documentation to support modern CMake `FetchContent` method over ExternalProject.


Areustle:
## Description

Hello, thank you for creating or contributing to such a useful and vital tool.

The [current build_system documentation](https://github.com/doctest/doctest/blob/86892fc480f80fb57d9a3926cb506c0e974489d8/doc/markdown/build-systems.md) describes using the CMake `ExternalProject` mechanism for acquiring the doctest header file and using it in a CMake project. `ExternalProject` is great for use with **Non-CMake** externals, but doctest is a modern CMake package with target support, so users can instead use the more efficient and elegant `FetchContent` mechanism.

FetchContent supports CMake targets. It also moves the download step to project _configure_ time rather build time like ExternalProject.

I suggest updating the documentation to prefer FetchContent.

For example:

> - You can also use the following CMake snippet to automatically fetch the entire **doctest** repository from github and configure it as an available target:
> 
> ```cmake
> include(FetchContent)
> FetchContent_Declare(
>   doctest
>   GIT_REPOSITORY https://github.com/doctest/doctest
>   GIT_TAG master # or any SHA, Branch or tag like v2.4.9
> )
> FetchContent_MakeAvailable(doctest)
>```
> The target doctest::doctest is now exposed at the global scope
> 
> And later you'll be able to use the doctest include directory like this:
> 
> ```cmake
> # add it globally
> include_directories(doctest::doctest)
> 
> # or per target
> target_include_directories(my_target PUBLIC doctest::doctest)
> ```
> 
> Access the doctest headers with a scoped `#include` directive. For example in `tests/main.cpp`
> ```c++
> #define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
> #include "doctest/doctest.h"
> 
> int factorial(int number) { return number <= 1 ? number : factorial(number - 1) * number; }
> 
> TEST_CASE("testing the factorial function") {
>     CHECK(factorial(1) == 1);
>     CHECK(factorial(2) == 2);
>     CHECK(factorial(3) == 6);
>     CHECK(factorial(10) == 3628800);
> }
> ```

If this is something you're interested in supporting I can open a PR with the above changes.

### Extra information

* doctest version: **v42.42.42**
* Operating System: **Joe's discount operating system**
* Compiler+version: **Hidden Dragon v1.2.3**

stephenlevitt:
Thanks for these instructions! I was missing the scoped include directive.

[...]

PR #866, opened 2024-08-17
Document a simpler CMake integration via FetchContent (#823)

### LINKED ISSUE #823: Suggestion: document a simpler CMake integration via FetchContent
Currently in [build-systems.md](https://github.com/doctest/doctest/blob/master/doc/markdown/build-systems.md#cmake) the suggested CMake integrations are either via manual file download or `ExternalProject`.
The first is, well, manual and the second is heavyweight, cumbersome and downloads at build time (yikes!).

There is a much simpler way to integrate doctest using [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html).

Something like this:

```cmake
include(FetchContent)

FetchContent_Declare(doctest
    URL  https://raw.githubusercontent.com/doctest/doctest/v2.4.11/doctest/doctest.h
    DOWNLOAD_NO_EXTRACT TRUE
    DOWNLOAD_EXTRACT_TIMESTAMP TRUE #for CMake 3.24 and higher
)
FetchContent_MakeAvailable(doctest)

set(DOCTEST_INCLUDE_DIR ${doctest_SOURCE_DIR})
```

This downloads just the header and does it during configure, not build, time.

If you think this is reasonable to document in addition or instead of `ExternalProject` method happy to submit a PR.

gershnik:
## Description
As per #823, this documents simpler CMake integration via `FetchContent`

## GitHub Issues
Closes #823

### DIFF
--- a/doc/markdown/build-systems.md
+++ b/doc/markdown/build-systems.md
@@ -34,19 +34,18 @@ FetchContent_Declare(doctest
 FetchContent_MakeAvailable(doctest)
 
 set(DOCTEST_INCLUDE_DIR ${doctest_SOURCE_DIR})
-```
 
-And later you'll be able to use the **doctest** include directory like this:
+...
 
-```cmake
 # add it globally
 include_directories(${DOCTEST_INCLUDE_DIR})
 
 # or per target
 target_include_directories(my_target PUBLIC ${DOCTEST_INCLUDE_DIR})
+
 ```
 
-- Finally, if you want CTest integration, you can fetch and use **doctest** as a CMake sub-project:
+- If you want CTest integration, you can fetch and use **doctest** as a CMake sub-project:
 
 ```cmake
 include(FetchContent)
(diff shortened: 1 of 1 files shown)


cdeln:
Thanks for the PR.
Can you please undo the removal of the current documentation of `External_ProjectAdd`?
Ideally there should be only additions and no removals, put the new docs under another subsection.
This will make it easier to review.

gershnik:
I certainly can but it seems a wrong thing to do in this context. The second `FetchContent` approach added is an exact replacement of `External_ProjectAdd` (this parallels the changes proposed in #821). If `External_ProjectAdd` is left in the doc a non CMake expert reading it would likely be confused about which way to use. 
Second, I don't believe there is a single situation where `External_ProjectAdd` would be preferable to `FetchContent`. Well, perhaps this is too strong. There might be a crazy situation for someone, somewhere but a person in such situation would already know all about `FetchContent` vs. `External_ProjectAdd` and how to map them to each other.
What do you think?

cdeln:
To begin with, I apologize for not seeing the duplicate PR. I just started work on reviewing old issues for this project and started out in the wrong order. Let's take the best out of both of these PRs!

I don't use any of the features myself so I can't confirm what you say, but it does indeed look simpler to use `FetchContent` than `External_ProjectAdd`, and it seems like several people agree on this point. Could you double check (you probably have already) the discussion in #698 and #821 to see if there is any improvements we can do to this PR?

I still suggest we keep the old documentation. We can remove it in a separate PR, it's just that pure additions followed by pure deletions are much easier to review than a change request. Also, as a general rule of thumb, just because one does not use a feature does not mean no one else does (I think this is a mindset important in open source development). Adding new info to the docs is non-controversial, but the deletion _might_ trigger a discussion, so it's safer to do this way (especially when this project has low dev capacity, is very stable and is closing in on maintenance mode only).

To be concrete, I suggest we do as follows
1) Rearrange to make this an addition only contribution
2) Put the new docs before the old docs for `External_ProjectAdd` with it's own subsection. You can put `External_ProjectAdd` in it's own subsection too, with a disclaimer that this is not the recommended way of doing it, but it's known to work and there for legacy reasons.

gershnik:
Done

UltraBlackLinux:
I don't know if I'm doing anything wrong here, but I cannot get this to work. All I get are linker errors: ```.../src/test/test.cpp:7:(.text+0x52): undefined reference to `doctest::Context::Context(int, char const* const*)'```

And that's with the example from this PR:
```
FetchContent_Declare(
  doctest
  URL https://raw.githubusercontent.com/doctest/doctest/v2.4.12/doctest/doctest.h
  DOWNLOAD_NO_EXTRACT TRUE
)
FetchContent_MakeAvailable(doctest)
include_directories(${doctest_SOURCE_DIR})
```
Any idea what could be wrong here?

Edit: Necrobumping is never it, eh? I won't even get into it, but apparently I just don't know how to work with C++. Cleaning up my mess of a codebase made the problem vanish, so that's a win I guess lol

mitchgrout:
> I don't know if I'm doing anything wrong here, but I cannot get this to work. All I get are linker errors: `` .../src/test/test.cpp:7:(.text+0x52): undefined reference to `doctest::Context::Context(int, char const* const*)' ``
> 
> And that's with the example from this PR:
> 
> ```
> FetchContent_Declare(
>   doctest
>   URL https://raw.githubusercontent.com/doctest/doctest/v2.4.12/doctest/doctest.h
>   DOWNLOAD_NO_EXTRACT TRUE
> )
> FetchContent_MakeAvailable(doctest)
> include_directories(${doctest_SOURCE_DIR})
> ```
> 
> Any idea what could be wrong here?
> 
> Edit: Necrobumping is never it, eh? I won't even get into it, but apparently I just don't know how to work with C++. Cleaning up my mess of a codebase made the problem vanish, so that's a win I guess lol

That sounds like doctest hasn't been compiled under `DOCTEST_CONFIG_IMPLEMENT` or similar -- the described strategy here only makes the `.h` available, but it won't compile the code for you. Check out [doc/markdown/main.md](https://github.com/doctest/doctest/blob/v2.4.12/doc/markdown/main.md)

mitchgrout:
@gershnik while I'm here, change looks good, happy to merge into dev. Just want to check that you're still happy with this diff before I merge though.

gershnik:
@mitchgrout Sure, still looks good to me!

PR #868, opened 2024-08-23
fix: documentation of workaround for thread locals on ESP32 platforms

### LINKED ISSUE #861: doctest crashes on ESP32 at startup (with workaround)
## Description

TL;DR: When using doctest with ESP32, either
 * `#define DOCTEST_THREAD_LOCAL` to disable thread local storage,
or, alternatively
  * `#define DOCTEST_CONFIG_NO_MULTITHREADING`, which is safer and more efficient if you don't use assertions across multiple threads.

## Background
* I'm using doctest with PlatformIO for both native and ESP32 builds. The ESP32 build uses an Arduino-style main, with setup() and empty loop()
* ESP32 has an [IPC task](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/ipc.html) for each CPU which is created during startup.
* Before running any code that I can see in the debugger, the ESP32 crashes with this error:

```
Guru Meditation Error: Core  0 panic'ed (Unhandled debug exception). 
Debug exception reason: Stack canary watchpoint triggered (ipc0) 
Core  0 register dump:
PC      : 0x403799ea  PS      : 0x00060036  A0      : 0x8037a0e4  A1      : 0x3fcf0aa0  
A2      : 0x3fc8e2c4  A3      : 0xffffffff  A4      : 0x00060023  A5      : 0xb33fffff  
A6      : 0x00000000  A7      : 0x00060023  A8      : 0x80379add  A9      : 0x3fc8f320  
```

While the [ESP-IDF supports thread-local storage]((https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/thread-local-storage.html#c11-std) ), using the C++11 `thread_local`, they come at the cost of increasing the stack usage of *all* tasks in the system, regardless of whether they use the thread-local variable nor not.

In this case, the thread-locals used by doctest are enough to blow up the stack of the ipc tasks, which have a 4K stack by default.

One possible solution would be to increase the stack size of all tasks to accommodate the thread locals. If you are targeting IDF directly, this could be a workable solution. However, I'm using PlatformIO, and presently it's not possible to reconfigure ESP-IDF at that level, since the IDF is a pre-compiled, and hence a pre-configured library wherein stack sizes for the system tasks are fixed.

Disabling thread local storage, or disabling multithreading in doctest, using the `#define`s above are the workarounds I found.

This wasn't an easy find :-). I hope this is of use to anyone else using doctest on an architecture that allocates thread locals on the stack of each thread.

m-mcgowan:
## Description

Documentation of workaround to stop ESP32 platforms from crashing due to thread local storage.

## GitHub Issues

Closes #861.

### DIFF
--- a/doc/markdown/configuration.md
+++ b/doc/markdown/configuration.md
@@ -6,39 +6,40 @@ The identifiers should be defined before the inclusion of the framework header.
 
 Defining something ```globally``` means for every source file of the binary (executable / shared object).
 
-- [**```DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN```**](#doctest_config_implement_with_main)
-- [**```DOCTEST_CONFIG_IMPLEMENT```**](#doctest_config_implement)
-- [**```DOCTEST_CONFIG_DISABLE```**](#doctest_config_disable)
-- [**```DOCTEST_CONFIG_IMPLEMENTATION_IN_DLL```**](#doctest_config_implementation_in_dll)
-- [**```DOCTEST_CONFIG_NO_SHORT_MACRO_NAMES```**](#doctest_config_no_short_macro_names)
-- [**```DOCTEST_CONFIG_TREAT_CHAR_STAR_AS_STRING```**](#doctest_config_treat_char_star_as_string)
-- [**```DOCTEST_CONFIG_REQUIRE_STRINGIFICATION_FOR_ALL_USED_TYPES```**](#doctest_config_require_stringification_for_all_used_types)
-- [**```DOCTEST_CONFIG_DOUBLE_STRINGIFY```**](#doctest_config_double_stringify)
-- [**```DOCTEST_CONFIG_SUPER_FAST_ASSERTS```**](#doctest_config_super_fast_asserts)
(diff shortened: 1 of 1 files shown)


Jason2866:
@m-mcgowan To use actual toolchains together with Arduino 3.0.4 take a look here https://github.com/pioarduino/platform-espressif32

mitchgrout:
Sorry for the delay in review; are you still interested in merging this change?

Just a few notes:

- This is pretty specific; yes, it breaks on ESP32, but the problem is more that `thread_local` causes bloat. It would probably be clearer to indicate that, e.g. "On memory-constrained systems, `thread_local` may cause the program to utilise more RAM than expected. This can be remediated by defining `-DDOCTEST_THREAD_LOCAL=` or similar".
- It seems you've written your explanation under `DOCTEST_CONFIG_NO_MULTITHREADING` -- it probably makes more sense under `DOCTEST_THREAD_LOCAL` since again, that's the root cause of this
- The mention of `platformio.ini` while well-intentioned is just one of many ways to configure the project. I would rather avoid talking too heavily about build-system specifics in this file

mitchgrout:
Closing as stale

Issue #882, opened 2024-11-27
Strip prefixes from file names


UnePierre:
## Description

Having files ordered in a structure of sub-folders makes the project-relative paths meaningful.
I dislike repeating this information in every individual file name.

Absolute paths are non-portable across different test environments.
Thus, there is the option to avoid all file paths in the output: `-npf` / `-no-path-filenames`.
But this also strips the file names of the meaningful part of their paths.
Removing only a configurable part at the start of paths could preserve the meaningful remainder.

I'd like to propose an option to strip the longest matching in a list of prefixes from every file name.

### To be Discussed

1. Maybe, an additional option to convert path separators (from `\` to `/`) might be useful?  See: https://cmake.org/cmake/help/latest/command/cmake_path.html#to-cmake-path-list
2. How about case insensitive paths? Does this need further options?

onqtam:
Merged the PR but I won't comment on what you've proposed here to be discussed - I don't have the bandwidth (see #554 ). You decide if this issue should be closed or remain open.

UnePierre:
I'll close it then and wait for complains from `\`-platforms.
If need be, we can go back to `-npf` for the `example/`-tests.

onqtam:
Just noticed that you've changed the default from ```-npf``` to ```-sfp``` for the tests of doctest itself (which I saw but didn't think much about when I looked at the PR) and now the tests are broken on Windows when I tested it locally.

Here's the gist of the difference - seems exactly related to ```/``` vs ```\``` and there's also an additional separator (```:```) in there after the removed drive letter:

```
-examples/all_features/test_cases_and_suites.cpp(0): FATAL ERROR: 
+:\doctest\examples\all_features\test_cases_and_suites.cpp(0): FATAL ERROR: 
```

for now I'll fall back to ```-npf``` for the tests & CI of doctest in ```common.cmake``` and will push that to ```dev```. Unfortunately the builds on the CI are broken and nobody is willing to fix them - maybe I shouldn't have been so quick to merge :D

So this issue should remain open until this works easily on windows too.

cdeln:
What is the status of this issue?

onqtam:
implemented for unix but not for windows

PR #883, opened 2024-11-27
Feat: strip path prefixes

UnePierre:
## Description

This implements a good part of #882:

- Add the option `-spp` / `--strip-file-prefixes`.
- Document the option itself.
- Add and document a configurable DOCTEST_CONFIG_OPTIONS_FILE_PREFIX_SEPARATOR.
- Apply the meaningful path prefixes to the examples.


doc/markdown/commandline.md
```diff
@@ -58,6 +58,7 @@ All the options can also be set with code (defaults/overrides) if the user [**su
 | ```-ns``` &nbsp; ```--no-skip=<bool>``` | Don't skip test cases marked as skip with a decorator |
 | ```-gfl``` ```--gnu-file-line=<bool>``` | ```:n:``` vs ```(n):``` for line numbers in output (gnu mode is usually for linux tools/IDEs and is with the ```:``` separator) |
 | ```-npf``` ```--no-path-filenames=<bool>``` | Paths are removed from the output when a filename is printed - useful if you want the same output from the testing framework on different environments |
+| ```-spp``` ```--skip-path-prefixes=<bool>``` | Remove the longest matching one of these prefixes from any file path in the output - useful as ```-npf```, but can preserve some context through relative paths. Try: ```--spp=${CMAKE_SOURCE_DIR}/:${CMAKE_BINARY_DIR}/``` |
```
onqtam, line 61: this should not be a <bool> but a <string>

And perhaps should be changed from ```useful as -npf``` to ```similar to -npf``` and from ```but can preserve some context through relative paths``` to ```but can preserve some context by not removing the entire relative paths``` or something like that

UnePierre: I'll also add a hint at the `:` separator.

UnePierre:
Sorry, I wasn't quick enough in re-basing this pull request in order to add #881's new test's result.

onqtam:
oh, sorry, I was too quick with the merge. yeah, open a new PR and I'll merge that one too.

UnePierre:
> oh, sorry, I was too quick with the merge. yeah, open a new PR and I'll merge that one too.

My pleasure.

PR #892, opened 2025-02-13
Python 3 update to range_based_execution.py

sddale:
## Description
Updated this script for compatibility with Python 3.
As far as I can tell, there is still not a built in way to run doctest tests in parallel. As such, this script is still useful but the current version on `dev` is written for Python 2.

### DIFF
--- a/examples/range_based_execution.py
+++ b/examples/range_based_execution.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
 
 import sys
 import math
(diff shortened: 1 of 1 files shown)


mitchgrout:
Just bumping this. A few notes:

1. This is mostly an _example_ script of how you would achieve parallelism with `doctest`. See `doc/markdown/commandline.md`. I'm not opposed to promoting this to a first-class exported program, but it does increase maintenance burden. If you're also doing isolated tests (e.g. one .exe per suite), then your build system is already inherently running them in parallel, making this less useful.
2. `there is still not a built in way to run doctest tests in parallel` -- correct, as long as we're talking about a single suite. To run tests in parallel would assume that tests don't rely on any global state. This is _generally wrong_, since as soon as your code touches non-reentrant-safe code (e.g. `strtok`), everything's going to blow up.

mitchgrout:
Closing as stale

PR #902, opened 2025-05-01
ci: modernize build configurations for Linux

severinstrobl:
## Description
- switched to containerized builds for old compilers
- disabled ASAN/UBSAN/TSAN for Clang 11 & 12

## Removed configurations
- GCC 4.9
- Clang 3.5
- Clang 3.6
- Clang 3.7
- Clang 3.8

## Added configurations
- GCC 13
- GCC 14
- Clang 16
- Clang 17
- Clang 18

### DIFF
--- a/.github/workflows/build_and_test.py
+++ b/.github/workflows/build_and_test.py
@@ -59,8 +59,7 @@ if _os == "Windows":
 elif _os == "Linux":
     if _compiler == "clang":
         if version_tuple(_version) <= version_tuple("6.0") or (
-            version_tuple(_version) >= version_tuple("11")
-            and version_tuple(_version) < version_tuple("13")
+            version_tuple("11") <= version_tuple(_version) < version_tuple("13")
         ):
             flags = ""
     elif _compiler == "gcc":
@@ -75,9 +74,8 @@ if _os == "Windows":
     tsan_flags = ""
 elif _os == "Linux":
     if _compiler == "clang":
-        if version_tuple(_version) <= version_tuple("3.9") or version_tuple(
-            _version
-        ) == version_tuple("11"):
+        if (version_tuple(_version) <= version_tuple("3.9") or
+            version_tuple(_version) == version_tuple("11")):
             tsan_flags = ""
     elif _compiler == "gcc":
         if version_tuple(_version) <= version_tuple("6.0"):
--- a/.github/workflows/build_and_test.py
+++ b/.github/workflows/build_and_test.py
@@ -30,7 +30,7 @@ def log_and_call(command):
     return os.system(command)
 
 
-def run_test(build_type, test_mode, flags, test = True):
+def run_test(build_type, test_mode, flags, test=True):
     print("Running: " + "; ".join([build_type, test_mode, flags, str(test)]))
     if log_and_call("cmake -E remove_directory build"):
         exit(1)
@@ -39,8 +39,8 @@ def run_test(build_type, test_mode, flags, test = True):
         f"-B build "
         f"-D CMAKE_BUILD_TYPE={build_type} "
         f"-D DOCTEST_TEST_MODE={test_mode} "
-        + (flags and f'-D CMAKE_CXX_FLAGS="{flags}" ') +
-        f"-D CMAKE_CXX_COMPILER={used_cxx}"
+        + (flags and f'-D CMAKE_CXX_FLAGS="{flags}" ')
+        + f"-D CMAKE_CXX_COMPILER={used_cxx}"
     ):
         exit(2)
     if log_and_call("cmake --build build"):
@@ -58,7 +58,10 @@ if _os == "Windows":
     flags = ""
 elif _os == "Linux":
     if _compiler == "clang":
-        if version_tuple(_version) <= version_tuple("6.0"):
+        if version_tuple(_version) <= version_tuple("6.0") or (
+            version_tuple(_version) >= version_tuple("11")
+            and version_tuple(_version) < version_tuple("13")
+        ):
             flags = ""
     elif _compiler == "gcc":
         if version_tuple(_version) <= version_tuple("5.0"):
@@ -72,7 +75,9 @@ if _os == "Windows":
     tsan_flags = ""
 elif _os == "Linux":
     if _compiler == "clang":
-        if version_tuple(_version) <= version_tuple("3.9"):
+        if version_tuple(_version) <= version_tuple("3.9") or version_tuple(
+            _version
+        ) == version_tuple("11"):
             tsan_flags = ""
     elif _compiler == "gcc":
         if version_tuple(_version) <= version_tuple("6.0"):
@@ -92,7 +97,7 @@ for configuration in ["Debug", "Release"]:
             configuration,
             "COMPARE",
             "-fno-exceptions -D DOCTEST_CONFIG_NO_EXCEPTIONS_BUT_WITH_ALL_ASSERTS",
-            test = False,
+            test=False,
         )
         run_test(configuration, "COMPARE", "-fno-rtti")
     if _os == "Linux":
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,16 +3,16 @@ name: CI
 on:
   push:
     paths-ignore:
-      - 'doc/**'
-      - 'scripts/**'
-      - 'LICENSE.txt'
-      - 'README.md'
+      - "doc/**"
+      - "scripts/**"
+      - "LICENSE.txt"
+      - "README.md"
   pull_request:
     paths-ignore:
-      - 'doc/**'
-      - 'scripts/**'
-      - 'LICENSE.txt'
-      - 'README.md'
+      - "doc/**"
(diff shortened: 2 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -231,64 +245,103 @@ jobs:
...
+          export CTEST_OUTPUT_ON_FAILURE="${{ env.CTEST_OUTPUT_ON_FAILURE }}"
+          export CTEST_PARALLEL_LEVEL="${{ env.CTEST_PARALLEL_LEVEL }}"
+          export CMAKE_GENERATOR="${{ env.CMAKE_GENERATOR }}"
+          export ASAN_OPTIONS="${{ env.ASAN_OPTIONS }}"
+          export TSAN_OPTIONS="${{ env.TSAN_OPTIONS }}"
+          EOF
+
       - name: Install (Linux)
-        if: runner.os == 'Linux'
+        if: runner.os == 'Linux' && matrix.container == ''
```
severinstrobl, line 264: Switching all Linux build configurations over to the containerized approach would be trivial and reduce the complexity of the CI configuration if this is preferred by the maintainers.

examples/all_features/bitfield_packed_struct.cpp
```diff
@@ -7,6 +7,7 @@ DOCTEST_MAKE_STD_HEADERS_CLEAN_FROM_WARNINGS_ON_WALL_END
 
 DOCTEST_GCC_SUPPRESS_WARNING("-Wmissing-declarations")
 DOCTEST_CLANG_SUPPRESS_WARNING("-Wmissing-prototypes")
+DOCTEST_CLANG_SUPPRESS_WARNING("-Wpacked")
```
severinstrobl, line 10: Disabled a warning triggered by Clang 3.9, 4.0, and 5.0.

.github/workflows/build_and_test.py
```diff
@@ -58,7 +58,10 @@ def version_tuple(v):
     flags = ""
 elif _os == "Linux":
     if _compiler == "clang":
-        if version_tuple(_version) <= version_tuple("6.0"):
+        if version_tuple(_version) <= version_tuple("6.0") or (
+            version_tuple(_version) >= version_tuple("11")
```
mitchgrout, line 62: Calls to `version_tuple` should always consist of a version in `<major>.<minor>` format, since...

```py
>>> version_tuple("6") < version_tuple("6.0")
True
>>> version_tuple("6.0") < version_tuple("6.0")
False
```

Also applicable on lines 63 and 80.

You could also consider simplifying this particular expression to

```py
version_tuple("11.0") < version_tuple(_version) < version_tuple("13.0")
```

.github/workflows/build_and_test.py
```diff
@@ -72,7 +75,9 @@ def version_tuple(v):
     tsan_flags = ""
 elif _os == "Linux":
     if _compiler == "clang":
-        if version_tuple(_version) <= version_tuple("3.9"):
+        if version_tuple(_version) <= version_tuple("3.9") or version_tuple(
+            _version
+        ) == version_tuple("11"):
```
mitchgrout, line 80: Assuming this formatting is from `black` or similar? Otherwise, nicer to read as just

```py
if version_tuple(_version) <= version_tuple("3.9") or
   version_tuple(_version) == version_tuple("11.0")
```

.github/workflows/main.yml
```diff
@@ -104,114 +104,128 @@ jobs:
         compiler: ["cl", "clang", "clang-cl"]
 
         include:
-          - os: ubuntu-18.04
+          - container: ubuntu:18.04
             compiler: gcc
             version: "4.8"
+            os: ubuntu-24.04
```
mitchgrout, line 110: Mainly to confirm my understanding but also for anyone reading this PR, we can't use 18.04 directly due to Actions providing the warning...

> ⚠ The ubuntu-18.04 environment is deprecated

...so we're going to swap to the recommended version of 24.04, and then run Ubuntu _again_ as a container to get access to the compiler's we're after, right?

.github/workflows/main.yml
```diff
@@ -104,114 +104,128 @@ jobs:
...
 
-          - os: ubuntu-18.04
```
mitchgrout, line 111: As noted in your change you're dropping support for gcc 4.9; what's the reasoning? Lack of a supported container, unavailability via `apt`?

.github/workflows/main.yml
```diff
@@ -104,114 +104,128 @@ jobs:
...
-            compiler: clang
-            version: "3.5"
-
-          - os: ubuntu-18.04
-            compiler: clang
-            version: "3.6"
-
-          - os: ubuntu-18.04
-            compiler: clang
-            version: "3.7"
```
mitchgrout, line 157: Same as with gcc, could you quickly elaborate on why we're dropping these?

.github/workflows/main.yml
```diff
@@ -104,114 +104,128 @@ jobs:
...
+          - os: ubuntu-24.04
             compiler: clang
             version: "14"
 
-          - os: ubuntu-latest
+          - os: ubuntu-24.04
             compiler: clang
             version: "15"
 
+          # disabled due to pre-existing build failures
```
mitchgrout, line 216: Given that the CI is still not 100% green, I don't see much harm in enabling these and looking to resolve the failures before next release. What failures are we experiencing here?

mitchgrout (changes requested):

.github/workflows/build_and_test.py
severinstrobl: The entire approach of comparing versions via conversions to tuples is IMHO flawed and should be replaced by something [more reliable](https://packaging.pypa.io/en/latest/version.html). In this specific case, it is important to consider that Clang changed the versioning format from `<major>.<minor>` to just `<major>` starting with version 7. So here, the values of `_version` we want to handle are `11` and `12`. So, unless I missed something, `version_tuple("11")` and `version_tuple("13")` is what we want.
Good point about combining the two checks, though.

.github/workflows/build_and_test.py
severinstrobl: Indeed, I'll format this check manually.

.github/workflows/main.yml
severinstrobl: Indeed, there are no [GitHub-hosted runners](https://github.com/actions/runner-images?tab=readme-ov-file#available-images) with Ubuntu 18.04/20.04 any more. And rather than trying to install some ancient compiler packages in a more recent Ubuntu release, we can use a containerized version of Ubuntu 18.04/20.04 running in Docker with the host being a Ubuntu 24.04 GitHub-hosted runner. After some experimentation this seemed like the best option to maintain coverage for these older compilers without having to create a Frankenstein version of Ubuntu.

.github/workflows/main.yml
severinstrobl: Yes, GCC 4.9 is not available in the standard packages sources for Ubuntu 18.04. And according to the [C++ support page](https://gcc.gnu.org/projects/cxx-status.html#cxx14), GCC 4.9 does not fully support C++14, so I assume people restricted to C++11 and old compilers are on GCC 4.8, while anyone on C++14 is using at least GCC 5, which is still covered in the changed CI setup.

.github/workflows/main.yml
severinstrobl: Similar reasoning, no packages for these old versions in the main repo and assuming/hoping people are at least on 3.9 by now.

.github/workflows/main.yml
severinstrobl: A number of warnings are triggered related to unsafe memory accesses, see #766 for some details. As this will require either code changes or disabling of some warnings as proposed, for example, in #768, I would rather get the CI green across all platforms first. WDYT?

severinstrobl:
@mitchgrout: Sorry for the delayed response to your review, I was travelling and out of the country for the past week.

.github/workflows/main.yml
mitchgrout: I think it's probably clearer to start with the additional compilers, then work to resolve the issues, since we may find more problems than the aforementioned two issues.

.github/workflows/build_and_test.py
mitchgrout: Ah, I see my mistake; the `_version` comes from the `main.yml` configuration, so needs to be aligned to that. No change required in that case.
I agree that swapping to `packaging.version.Version` or similar is worthwhile; better to do in another change I think.

mitchgrout:
> @mitchgrout: Sorry for the delayed response to your review, I was travelling and out of the country for the past week.

No rush. I'm happy with the change as-is, although enabling those extra `clang` compilers I think would be worthwhile. Drop a ping when you're ready to merge

severinstrobl:
> No rush. I'm happy with the change as-is, although enabling those extra `clang` compilers I think would be worthwhile. Drop a ping when you're ready to merge

@mitchgrout: The additional Clang versions are also enabled, I'd say we're good to go here.

Issue #871, opened 2024-08-30
Fix broken CI


cdeln:
CI is broken, no idea how long it has been in this state, but it should be fixed before any PRs touching code can be merged.

mitchgrout:
For the `coverity` job, seems that we may need to update the repo config somewhat, **or** provide an access token. In either case, it seems to be something only the project owner / higher-level maintainers could achieve. At least for now, it seems we can run the job every other couple of hours, but that's not _that_ great

onqtam:
@mitchgrout just made you an owner!

severinstrobl:
Guess we can consider this one done now 😄

PR #907, opened 2025-05-14
ci: modernize build configurations for macOS

severinstrobl:
### Removed configurations
- macOS 10/Xcode 10.3
- macOS 12/Xcode 14.0.1
- macOS latest/Xcode 13.2.1

### Added configurations
- macOS 13 (Intel):
  - Xcode 14.1
  - Xcode 14.3.1
  - Xcode 15.2
  - GCC 12 (w/o sanitizers)
- macOS 14 (Arm64)
  - Xcode 15.0.1
  - Xcode 15.4
  - Xcode 16.2
- macOS 15 (Arm64)
  - Xcode 16.3
  - GCC 12 (w/o sanitizers)

## GitHub Issues

Part of #871

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -225,21 +225,43 @@ jobs:
             compiler: clang
             version: "18"
 
-          - os: macOS-10.15
+          # macOS Intel
+          - os: macos-13
             compiler: xcode
-            version: "10.3"
+            version: "14.1" # Apple Clang 14.0.0
 
-          - os: macOS-latest
+          - os: macos-13
             compiler: xcode
-            version: "13.2.1"
+            version: "14.3.1" # Apple Clang 14.0.3
 
-          - os: macos-12
+          - os: macos-13
             compiler: xcode
-            version: "14.0.1"
+            version: "15.2" # Apple Clang 15.0.0
 
-          - os: macOS-latest
+          - os: macos-13
             compiler: gcc
-            version: "11"
+            version: "12"
+
+          # macOS arm64
+          - os: macos-14
+            compiler: xcode
+            version: "15.0.1" # Apple Clang 15.0.0
+
+          - os: macos-14
+            compiler: xcode
+            version: "15.4" # Apple Clang 15.0.0
+
+          - os: macos-14
+            compiler: xcode
+            version: "16.2" # Apple Clang 16.0.0
+
+          - os: macos-15
+            compiler: xcode
+            version: "16.3" # Apple Clang 17.0.0
+
+          - os: macos-15
+            compiler: gcc
+            version: "12"
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -225,21 +225,43 @@ jobs:
...
+            compiler: xcode
+            version: "15.4" # Apple Clang 15.0.0
+
+          - os: macos-14
+            compiler: xcode
+            version: "16.2" # Apple Clang 16.0.0
+
+          - os: macos-15
+            compiler: xcode
+            version: "16.3" # Apple Clang 17.0.0
```
severinstrobl, line 260: These two configurations fail with ASAN/UBSAN in the `disabled_but_evaluated` test case, not sure if this an issue with the code or the tooling.

.github/workflows/main.yml
```diff
@@ -225,21 +225,43 @@ jobs:
...
-          - os: macos-12
+          - os: macos-13
             compiler: xcode
-            version: "14.0.1"
+            version: "15.2" # Apple Clang 15.0.0
 
-          - os: macOS-latest
+          - os: macos-13
             compiler: gcc
-            version: "11"
```
mitchgrout, line 242: Just clarifying again, dropping due to lack of OS support / compiler availability?

.github/workflows/main.yml
mitchgrout: It's odd, it runs and passes without issue on other toolchains:

```text
        Start 148: disabled_but_evaluated
        .....
148/164 Test #148: disabled_but_evaluated .....................   Passed    0.03 sec
```

The error from this failure is pretty strange though:

```text
dyld[7671]: main executable has no entry point
CMake Error at /Users/runner/work/doctest/doctest/scripts/cmake/exec_test.cmake:69 (message):
  Running
  '/Users/runner/work/doctest/doctest/build/examples/all_features/disabled_but_evaluated
  --dt-no-path-filenames=1 --dt-no-line-numbers=1 --dt-no-exitcode=1
  --dt-gnu-file-line=0 --dt-no-time-in-output=1' ended with code 'Subprocess
  aborted'
```

It does make me suspect that this is a toolchain issue, but even so, the implementation of the files are even stranger; there is no `main` method, and we have globally set `DOCTEST_CONFIG_DISABLE` which should suppress doctest's `main` method too. 

I might try spinning up a VM in the next few days to see if I can reproduce it, but if we can't answer it we can always leave it red and try to fix at a later point

.github/workflows/main.yml
severinstrobl: Exactly, no runners available any more for these older versions of macOS. We still have one generation of Intel-based runners, everything beyond this is ARM64.

.github/workflows/main.yml
mitchgrout: Well, seems that this fixes it; for posterity, can we pop a comment alongside `main` summarizing why it's necessary (i.e. preventing a strange optimization).
I don't think I've ever seen the compiler/linker fully optimize away a `main` method before. Maybe worth setting up an issue to track this?

severinstrobl: Added a comment. I think it is somewhat specific to this case, where we don't have any test cases registered at all. You'd want to investigate under which circumstances the main gets optimized away and how to prevent this?

PR #909, opened 2025-05-16
fix: replace deprecated <ciso646> with <version> to enjoy C++20!

houchen-li:
## Description

"\<ciso646\>" is deprecated in C++17 and removed since C++20. Use "\<version\>" instead.

https://en.cppreference.com/w/cpp/header/ciso646


mitchgrout:
Already resolved in #901

Issue #906, opened 2025-05-12
Replace `version_tuple` in `.github/workflows/build_and_test.py`


mitchgrout:
As noted in #902 the current `version_tuple` pattern is insufficient for testing. Currently, the implementation is:

```py
def version_tuple(v):
    return tuple(map(int, (v.split("."))))
```

...which has the following problems:

```py
>>> version_tuple("1")
(1,)
>>> version_tuple("1.0")
(1, 0)
>>> version_tuple("1") == version_tuple("1.0")
False
```

Our target compilers do not all report a single coherent version, comparison by `tuple` is insufficient as it is defined to be lexicographical ordering; however, if we interpreted as semver then our checks would be at least a bit more coherent. One proposed option is `packaging.version.Version`, which instead has:

```py
>>> Version("1")
<Version('1')>
>>> Version("1.0")
<Version('1.0')>
>>> Version("1") == Version("1.0")
True
```

SlugsGate:
Hi! I am a newcomer here. Can I try to work on this issue?

mitchgrout:
> Hi! I am a newcomer here. Can I try to work on this issue?

You're more than welcome to. Have a quick read through https://github.com/doctest/doctest/blob/master/CONTRIBUTING.md#pull-requests for info about making a PR, and just @ me on the issue or similar when you're done.

SlugsGate:
Hmm. I run `python D:\Git-Contribs\doctest\doctest\.github\workflows\build_and_test.py Windows x64 gcc` without making any changes after cloning but i failed the `no_multithreading` test. I have attached the log file. This also occurs when I do make my changes.

[LastTest.log](https://github.com/user-attachments/files/20261376/LastTest.log)

mitchgrout:
Closed by #1066

PR #913, opened 2025-05-17
switched from tuple to Version in build_and_test.py

SlugsGate:
## Description

Replaced tuple with Version in `build_and_test.py`. The current `version_tuple` pattern was insufficient for testing and `Version` provided a much better alternative.

## GitHub Issues

As requested in #906

### DIFF
--- a/.github/workflows/build_and_test.py
+++ b/.github/workflows/build_and_test.py
@@ -1,6 +1,8 @@
 import os
 import sys
 
+from packaging.version import Version, InvalidVersion
+
 _os = sys.argv[1]
 assert _os in ["Linux", "macOS", "Windows"]
 
@@ -21,7 +23,7 @@ elif _compiler == "clang" or _compiler == "xcode":
 else:
     used_cxx = _compiler
 
-if _os == "Linux":
+if _os == "Linux" and _version:
     used_cxx += "-" + _version
 
 
@@ -30,7 +32,7 @@ def log_and_call(command):
     return os.system(command)
 
 
-def run_test(build_type, test_mode, flags, test=True):
+def run_test(build_type, test_mode, flags, test = True):
     print("Running: " + "; ".join([build_type, test_mode, flags, str(test)]))
     if log_and_call("cmake -E remove_directory build"):
         exit(1)
@@ -39,8 +41,8 @@ def run_test(build_type, test_mode, flags, test=True):
         f"-B build "
         f"-D CMAKE_BUILD_TYPE={build_type} "
         f"-D DOCTEST_TEST_MODE={test_mode} "
-        + (flags and f'-D CMAKE_CXX_FLAGS="{flags}" ')
-        + f"-D CMAKE_CXX_COMPILER={used_cxx}"
+        + (flags and f'-D CMAKE_CXX_FLAGS="{flags}" ') +
+        f"-D CMAKE_CXX_COMPILER={used_cxx}"
     ):
         exit(2)
     if log_and_call("cmake --build build"):
@@ -50,7 +52,10 @@ def run_test(build_type, test_mode, flags, test=True):
 
 
 def version_tuple(v):
-    return tuple(map(int, (v.split("."))))
+    try:
+        return Version(v)
+    except InvalidVersion:
+        raise ValueError(f"Invalid version string: {v}")
 
 
 flags = "-fsanitize=address,undefined -fno-omit-frame-pointer"
@@ -58,12 +63,10 @@ if _os == "Windows":
     flags = ""
 elif _os == "Linux":
     if _compiler == "clang":
-        if version_tuple(_version) <= version_tuple("6.0") or (
-            version_tuple("11") <= version_tuple(_version) < version_tuple("13")
-        ):
+        if _version and version_tuple(_version) <= version_tuple("6.0"):
             flags = ""
     elif _compiler == "gcc":
-        if version_tuple(_version) <= version_tuple("5.0"):
+        if _version and version_tuple(_version) <= version_tuple("5.0"):
             flags = ""
 
 if _os == "Linux" and _compiler == "gcc":
@@ -74,11 +77,10 @@ if _os == "Windows":
     tsan_flags = ""
 elif _os == "Linux":
     if _compiler == "clang":
-        if (version_tuple(_version) <= version_tuple("3.9") or
-            version_tuple(_version) == version_tuple("11")):
+        if _version and version_tuple(_version) <= version_tuple("3.9"):
             tsan_flags = ""
     elif _compiler == "gcc":
-        if version_tuple(_version) <= version_tuple("6.0"):
+        if _version and version_tuple(_version) <= version_tuple("6.0"):
             tsan_flags = ""
 
 if _os == "Linux" and _compiler == "gcc":
@@ -95,7 +97,7 @@ for configuration in ["Debug", "Release"]:
             configuration,
             "COMPARE",
             "-fno-exceptions -D DOCTEST_CONFIG_NO_EXCEPTIONS_BUT_WITH_ALL_ASSERTS",
-            test=False,
+            test = False,
         )
         run_test(configuration, "COMPARE", "-fno-rtti")
     if _os == "Linux":


.github/workflows/build_and_test.py
```diff
@@ -21,7 +23,7 @@
 else:
     used_cxx = _compiler
 
-if _os == "Linux":
+if _os == "Linux" and _version:
```
mitchgrout, line 26: 90% sure the CI will always set `_version`, so worthwhile trying instead:

```py
_version = Version(sys.argv[4])
# ...
if _os == "Linux":
  used_cxx += "-" + str(_version)
```

Also good to convert the `str` value to a `Version` as fast as possible, and only once

.github/workflows/build_and_test.py
```diff
@@ -50,20 +52,21 @@ def run_test(build_type, test_mode, flags, test=True):
 
 
 def version_tuple(v):
-    return tuple(map(int, (v.split("."))))
+    try:
+        return Version(v)
+    except InvalidVersion:
+        raise ValueError(f"Invalid version string: {v}")
```
mitchgrout, line 58: I'd prefer we get just rid of this method entirely, as it boils down to just `Version`. Also, don't catch `InvalidError`, since this is for the CI it's fine if we have detail errors showing up

.github/workflows/build_and_test.py
```diff
@@ -50,20 +52,21 @@ def run_test(build_type, test_mode, flags, test=True):
...
 
 flags = "-fsanitize=address,undefined -fno-omit-frame-pointer"
 if _os == "Windows":
     flags = ""
 elif _os == "Linux":
     if _compiler == "clang":
-        if version_tuple(_version) <= version_tuple("6.0") or (
-            version_tuple("11") <= version_tuple(_version) < version_tuple("13")
-        ):
+        if _version and version_tuple(_version) <= version_tuple("6.0"):
```
mitchgrout, line 66: Unintended change in behaviour; we recently updated this on `dev`, may need to rebase

mitchgrout (changes requested):
Looks like `packaging.version` isn't available on all our runners, so you have some extra failing checks. My guess is we'll need to slap in a `pip install packaging==25.0` or similar into the `.github/workflows/main.yml`

SlugsGate:
Hi! Please disregard the force push and other pull request. I was trying to figure out if I should open another pull request or not.

SlugsGate:
So it looks like we need to add another argument for compiler versions as some tests fail because of this. the MacOS tests fail because I do not know the proper syntax for its package managers so we will need to update those accordingly. Rest of the failing tests are failing because of errors in `.cpp` files.

.github/workflows/build_and_test.py
aaslam12: Reverted back the if statement.

.github/workflows/build_and_test.py
```diff
@@ -39,8 +43,8 @@ def run_test(build_type, test_mode, flags, test=True):
         f"-B build "
         f"-D CMAKE_BUILD_TYPE={build_type} "
         f"-D DOCTEST_TEST_MODE={test_mode} "
-        + (flags and f'-D CMAKE_CXX_FLAGS="{flags}" ')
-        + f"-D CMAKE_CXX_COMPILER={used_cxx}"
+        + (flags and f'-D CMAKE_CXX_FLAGS="{flags}" ') +
+        f"-D CMAKE_CXX_COMPILER={used_cxx}"
```
mitchgrout, line 47: Not sure why you're adjusting the layout here; revert

.github/workflows/build_and_test.py
```diff
@@ -49,21 +53,17 @@ def run_test(build_type, test_mode, flags, test=True):
...
-
-
 flags = "-fsanitize=address,undefined -fno-omit-frame-pointer"
 if _os == "Windows":
     flags = ""
 elif _os == "Linux":
     if _compiler == "clang":
-        if version_tuple(_version) <= version_tuple("6.0") or (
-            version_tuple("11") <= version_tuple(_version) < version_tuple("13")
+        if Version(str(_version)) <= Version("6.0") or (
```
mitchgrout, line 61: `_version` is already of type `Version`; this conversion is redundant

.github/workflows/main.yml
```diff
@@ -28,6 +28,9 @@ jobs:
       - name: Install
         run: sudo apt-get install -y ninja-build lcov
 
+      - name: Install Python Dependencies
+        run: pip install packaging
+
```
mitchgrout, line 33: What's the rationale for installing `packaging` for the `coverage` task? Seems that we don't use it, as we don't use `build_and_test.py` for this job

.github/workflows/main.yml
```diff
@@ -244,6 +247,20 @@ jobs:
...
+      - name: Install Python Dependencies (non-containerized)
+        if: matrix.container == ''
+        run: |
+          # Different package managers depending on OS
+          if [ "${{ runner.os }}" == "Linux" ]; then
+            pip3 install packaging
+          elif [ "${{ runner.os }}" == "macOS" ]; then
+            pip3 install packaging
+          else
+            pip install packaging  # Windows
```
mitchgrout, line 260: Preferably these would be pegged to either an exact version, or have a constraint like `>=25.0`. I'd recommend having something like `.github/workflows/requirements.txt` to avoid duplication

.github/workflows/main.yml
```diff
@@ -309,6 +329,16 @@ jobs:
...
+      # For x64 builds
+      - name: Verify Python Dependencies (x64, non-containerized)
+        if: matrix.container == ''
+        run: python3 -c "from packaging.version import Version; print('packaging module is installed')"
+
+      # For x64 containerized builds
+      - name: Verify Python Dependencies (x64, containerized)
+        if: runner.os == 'Linux' && matrix.container != ''
+        run: docker exec build-container bash -c "python3 -c \"from packaging.version import Version; print('packaging module is installed')\""
+
```
mitchgrout, line 341: I don't think we get much value from this; if it doesn't exist, the `build_and_test.py` run will just fail with a pretty clear error

.github/workflows/main.yml
```diff
@@ -381,7 +411,7 @@ jobs:
 
       - name: Generate
         run: cmake -B build -S . -G "${{ matrix.toolset == 'v143' && 'Visual Studio 17 2022' || 'Visual Studio 16 2019' }}" \
-            -A ${{ matrix.architecture }} -T ${{ matrix.toolset }}
+          -A ${{ matrix.architecture }} -T ${{ matrix.toolset }}
```
mitchgrout, line 414: Not sure why you're changing formatting here; revert

.github/workflows/build_and_test.py
```diff
@@ -21,16 +23,18 @@
 else:
     used_cxx = _compiler
 
+_version = Version(sys.argv[4])
+# ...
```
mitchgrout, line 27: No point in this comment being here

.github/workflows/build_and_test.py
```diff
@@ -1,6 +1,8 @@
 import os
 import sys
 
+from packaging.version import Version, InvalidVersion
```
mitchgrout, line 4: `InvalidVersion` is now unused

.github/workflows/build_and_test.py
aaslam12: Ah. I think my IDE is automatically reformatting without me realizing. I'll revert it back.

.github/workflows/main.yml
```diff
@@ -289,7 +303,10 @@ jobs:
             fi
 
             apt-get update
-            apt-get install -y cmake ninja-build valgrind libc6-dbg:i386 python3
+            apt-get install -y cmake ninja-build valgrind libc6-dbg:i386 python3 python3-pip
+
+            pip3 install packaging
```
mitchgrout, line 308: You're missing the `==25.0` pin here

.github/workflows/build_and_test.py
```diff
@@ -21,8 +23,10 @@
 else:
     used_cxx = _compiler
 
+_version = Version(sys.argv[4]) if len(sys.argv) >= 5 else ""
```
mitchgrout, line 26: A little unfortunate that we have `_version: Version | str` -- personally I'd prefer `None` but either should work, since the comparison with `Version` should fail if we accidentally try to compare.

mitchgrout (commented):
Think we're pretty much good to go (except one `==25.0` missing), we'll merge after that + CI finishes to make sure we haven't broken any unexpected jobs

.github/workflows/build_and_test.py
aaslam12: I have replaced it with `None`.

.github/workflows/main.yml
aaslam12: Thanks

aaslam12:
[ci (clang, ubuntu-24.04, ubuntu:20.04, 7)](https://github.com/doctest/doctest/actions/runs/15148440251/job/42589633110?pr=913#logs) fails because it is stuck in setting up the `tzdata` package for some reason when the others do not get stuck there.

aaslam12:
Some checks are still failing because of the error: `ModuleNotFoundError: No module named 'packaging'` even after I tried moving this further in the `main.yaml` steps. I noticed that this issue only gets fixed if I remove the `==25.0`. Do you know of any reason why this would happen? @mitchgrout

mitchgrout:
> > Think we're pretty much good to go (except one `==25.0` missing), we'll merge after that + CI finishes to make sure we haven't broken any unexpected jobs
> 
> [ci (clang, ubuntu-24.04, ubuntu:20.04, 7)](https://github.com/doctest/doctest/actions/runs/15148440251/job/42589633110?pr=913#logs) and a couple others fail because they are getting stuck in setting up the `tzdata` package for some reason when the others do not get stuck there.
> 
> The MacOS tests fail because I do not know the proper syntax for its package managers so we will need to update those accordingly. Rest of the failing tests are failing because of errors in `.cpp` files. There are also some tests failing because of chocolatey.

I'll have a more detailed look over the CI config tomorrow, but I don't think there's much harm in setting `TZ`; seems there's a few posts out there detailing that as a reasonable solution. My guess is `python3-pip` introduces it, and for those particular containers, `TZ` or similar simply isn't set.

For MacOS, the error implies you should be using `brew` for installing. Could try setting up a VM, or just push to CI until it starts working.

Regarding `==25.0` breaking, I think we may have just overlooked that the `pip install` needs to be run within the container -- with the current config, it would be installed in the host, no?

aaslam12:
Ignore review request. Sorry. Misclick.

.github/workflows/main.yml
```diff
@@ -320,7 +322,9 @@ jobs:
             cd /workspace
             source container-env.sh
             python3 .github/workflows/build_and_test.py ${{ runner.os }} x64 ${{ matrix.compiler }} ${{ matrix.version }}
-        run: docker exec build-container bash -c "$CMD"
+        run:
+          python3 -m pip install packaging==25.0
```
severinstrobl: This needs to go into the list of commands in `$CMD`, otherwise it will be executed on the host, not in the scope of the container. Also, this needs to be done only once, best in the `Install (Linux, containerized)` step.

.github/workflows/main.yml
```diff
@@ -330,7 +334,9 @@ jobs:
       # MacOS doesn't support x86 from Xcode 10 onwards.
       - name: Build & Test x86
         if: (runner.os == 'Linux' && matrix.container == '') || (runner.os == 'Windows' && matrix.compiler != 'clang-cl')
-        run: python3 .github/workflows/build_and_test.py ${{ runner.os }} x86 ${{ matrix.compiler }} ${{ matrix.version }}
+        run:
+          python3 -m pip install packaging==25.0
```
severinstrobl: IMHO it is cleaner to separate the installation of build dependencies and test execution, makes it easier to spot what failed when the pipeline breaks.

.github/workflows/main.yml
```diff
@@ -335,6 +338,12 @@ jobs:
         if: runner.os == 'Windows'
         uses: ilammy/msvc-dev-cmd@v1
 
+      # Add Python packaging installation for all runs (except containerized)
+      - name: Install Python Dependencies
+        run: |
+          python3 -m pip install packaging==25.0
```
mitchgrout, line 344: All the others are 21.3, but this one is 25.0 -- why is that?

.github/workflows/main.yml
```diff
@@ -335,6 +338,12 @@ jobs:
         if: runner.os == 'Windows'
         uses: ilammy/msvc-dev-cmd@v1
 
+      # Add Python packaging installation for all runs (except containerized)
```
mitchgrout, line 341: You say 'except containerized', but afaik this is going to run even for containers?

.github/workflows/main.yml
aaslam12: I looked at the logs for runners and some of the runners only had versions up to 21, windows has it up to 25. It's weird.

.github/workflows/main.yml
aaslam12: How do I make it run on non-containerized builds?

aaslam12:
> Can you please rebase on top of `doctest:dev` just so we can double-check that the macOS jobs are passing

Rebasing complete.

mitchgrout:
Sorry, let this go stale. If you're still interested, could you please rebase again; CI/CD should be green at this time.

I'd also still _really_ prefer that we have a `requirements.txt` or similar, since manually specifying the reqs. in the CI/CD config is going to be a maintenance burden.

mitchgrout:
Superseded by #1066

PR #924, opened 2025-07-01
Split toString(String) into && and const& overloads

RedBeard0531:
This reduces the cost of DOCTEST_CONFIG_DOUBLE_STRINGIFY to effectively nothing.

### DIFF
--- a/doctest/parts/doctest.cpp
+++ b/doctest/parts/doctest.cpp
@@ -871,7 +871,7 @@ String toString(const char* in) { return String("\"") + (in ? in : "{null string
 String toString(const std::string& in) { return in.c_str(); }
 #endif // VS 2019
 
-String toString(String in) { return in; }
+String toString(const String& in) { return in; }
 
 String toString(std::nullptr_t) { return "nullptr"; }
 
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -1108,6 +1108,8 @@ String toString(const DOCTEST_REF_WRAP(T) value) {
     return StringMaker<T>::convert(value);
 }
 
+inline String&& toString(String&& in) { return static_cast<String&&>(in); }
+
 #ifdef DOCTEST_CONFIG_TREAT_CHAR_STAR_AS_STRING
 DOCTEST_INTERFACE String toString(const char* in);
 #endif // DOCTEST_CONFIG_TREAT_CHAR_STAR_AS_STRING
@@ -1117,7 +1119,7 @@ DOCTEST_INTERFACE String toString(const char* in);
 DOCTEST_INTERFACE String toString(const std::string& in);
 #endif // VS 2019
 
-DOCTEST_INTERFACE String toString(String in);
+DOCTEST_INTERFACE String toString(const String& in);
 
 DOCTEST_INTERFACE String toString(std::nullptr_t);
 
@@ -3962,7 +3964,7 @@ String toString(const char* in) { return String("\"") + (in ? in : "{null string
 String toString(const std::string& in) { return in.c_str(); }
 #endif // VS 2019
 
-String toString(String in) { return in; }
+String toString(const String& in) { return in; }
 
 String toString(std::nullptr_t) { return "nullptr"; }
 
--- a/doctest/parts/doctest_fwd.h
+++ b/doctest/parts/doctest_fwd.h
@@ -1105,6 +1105,8 @@ String toString(const DOCTEST_REF_WRAP(T) value) {
     return StringMaker<T>::convert(value);
 }
 
+inline String&& toString(String&& in) { return static_cast<String&&>(in); }
+
 #ifdef DOCTEST_CONFIG_TREAT_CHAR_STAR_AS_STRING
 DOCTEST_INTERFACE String toString(const char* in);
 #endif // DOCTEST_CONFIG_TREAT_CHAR_STAR_AS_STRING
@@ -1114,7 +1116,7 @@ DOCTEST_INTERFACE String toString(const char* in);
 DOCTEST_INTERFACE String toString(const std::string& in);
 #endif // VS 2019
 
-DOCTEST_INTERFACE String toString(String in);
+DOCTEST_INTERFACE String toString(const String& in);
 
 DOCTEST_INTERFACE String toString(std::nullptr_t);
 


doctest/parts/doctest.cpp
```diff
@@ -871,7 +871,7 @@ String toString(const char* in) { return String("\"") + (in ? in : "{null string
 String toString(const std::string& in) { return in.c_str(); }
 #endif // VS 2019
 
-String toString(String in) { return in; }
+String toString(const String& in) { return in; }
```
RedBeard0531, line 874: This was doing two copies when passed an lvalue `String`. Now it just does one. That wasn't what I was trying to fix, but it is a nice side benefit.

RedBeard0531:
I'm pretty sure the failures in CI are unrelated to my change and are pre-existing issues.

mitchgrout:
> This reduces the cost of DOCTEST_CONFIG_DOUBLE_STRINGIFY to effectively nothing.

Could you give a quick explanation of your reasoning for this? Just saves having to dig through the code as much.

RedBeard0531:
Previously the double stringify logic would do toString(toString(foo)). This is nice if your custom toString returns a std::string because it will convert it to a doctest::String. But it added overhead to the built-in cases since they already return String. The problem was that the toString(String) took by value then retuned by copy (although I now see that this case is probably eligible for implicit move even if it isn't eligible for NRVO). Now there is  dedicated String&& overload that literally does nothing and returns an r-value reference to the original String object.

mitchgrout:
> Previously the double stringify logic would do toString(toString(foo)). This is nice if your custom toString returns a std::string because it will convert it to a doctest::String. But it added overhead to the built-in cases since they already return String. The problem was that the toString(String) took by value then retuned by copy (although I now see that this case is probably eligible for implicit move even if it isn't eligible for NRVO). Now there is dedicated String&& overload that literally does nothing and returns an r-value reference to the original String object.

I would expect (at least if your TU observes `DOCTEST_CONFIG_IMPLEMENT`) that (N)RVO would apply and elide the copy, but I can see a possible benefit to having the pr-value overload available for cases where we compile doctest in a separate unit. 

However, I would also think that if we were to inline the `String toString(String)` implementation in the header, we'd get the same effect, since the compiler could reason about the method a little better at the call site, and possibly elide it that way.

It'd be good to confirm via disassembly whether:

1. Your current fix actually does elide an unnecessary/copy move somewhere in the call chain, and
2. We can't achieve this by a direct inline of the `toString` method for this particular overload

RedBeard0531:
As you can [see](https://godbolt.org/z/6e6fo9TfP) the new version generates significantly better code both with and without optimizations. This largely comes down to there only being 1 `String` object being created rather than two.

The rightmost column covers the case where `DOCTEST_CONFIG_IMPLEMENT` is defined since you were curious, but I don't think that matters since it should only be in a single TU and the docs even [recommend](https://github.com/doctest/doctest/blob/1da23a3e8119ec5cce4f9388e91b065e20bf06f5/doc/markdown/tutorial.md#scaling-up) not putting any actual code in that TU other than the `#include`.

> (N)RVO would apply and elide the copy

Note that copy elision (the standardese term for (N)RVO) is [explicitly](https://eel.is/c++draft/class.copy.elision#1.1) disallowed for the case of returning a function parameter. It couldn't possibly work (at least for out-of-line function calls) because ABIs require that the uninitialized return slot be a different memory address than any by-value parameter. That said, a _[different](https://eel.is/c++draft/expr.prim.id.unqual#15)_ rule kicks in that allows `return arg;` to implicitly become a move rather than a copy . The wording, requirements, and location for that rule have changed many times since it was introduced in C++11 and the current version is much more relaxed than it used to be. I _think_ that implicit moving from arguments has always been allowed, but I'm not 100% sure. FWIW, here is the codegen for g++-15.1 at -O3 for the existing `toString(String)` showing that a move is performed:

```asm
doctest::toString(doctest::String):
        movdqu  xmm0, XMMWORD PTR [rsi]
        mov     rax, rdi
        movups  XMMWORD PTR [rdi], xmm0
        mov     rdx, QWORD PTR [rsi+16]
        mov     QWORD PTR [rdi+16], rdx
        mov     BYTE PTR [rsi], 0
        mov     BYTE PTR [rsi+23], 23
        ret
```

mitchgrout:
> As you can [see](https://godbolt.org/z/6e6fo9TfP) the new version generates significantly better code both with and without optimizations. This largely comes down to there only being 1 `String` object being created rather than two.
> 
> The rightmost column covers the case where `DOCTEST_CONFIG_IMPLEMENT` is defined since you were curious, but I don't think that matters since it should only be in a single TU and the docs even [recommend](https://github.com/doctest/doctest/blob/1da23a3e8119ec5cce4f9388e91b065e20bf06f5/doc/markdown/tutorial.md#scaling-up) not putting any actual code in that TU other than the `#include`.
> 
> > (N)RVO would apply and elide the copy
> 
> Note that copy elision (the standardese term for (N)RVO) is [explicitly](https://eel.is/c++draft/class.copy.elision#1.1) disallowed for the case of returning a function parameter. It couldn't possibly work (at least for out-of-line function calls) because ABIs require that the uninitialized return slot be a different memory address than any by-value parameter. That said, a _[different](https://eel.is/c++draft/expr.prim.id.unqual#15)_ rule kicks in that allows `return arg;` to implicitly become a move rather than a copy . The wording, requirements, and location for that rule have changed many times since it was introduced in C++11 and the current version is much more relaxed than it used to be. I _think_ that implicit moving from arguments has always been allowed, but I'm not 100% sure. FWIW, here is the codegen for g++-15.1 at -O3 for the existing `toString(String)` showing that a move is performed:
> 
> ```assembly
> doctest::toString(doctest::String):
>         movdqu  xmm0, XMMWORD PTR [rsi]
>         mov     rax, rdi
>         movups  XMMWORD PTR [rdi], xmm0
>         mov     rdx, QWORD PTR [rsi+16]
>         mov     QWORD PTR [rdi+16], rdx
>         mov     BYTE PTR [rsi], 0
>         mov     BYTE PTR [rsi+23], 23
>         ret
> ```

Interesting! I wasn't aware of that exception for parameters, but I suppose it makes sense. I'm satisfied with that, so I'll merge now

Issue #928, opened 2025-07-19
-Wnrvo warning for gcc 15.1.1


nylonhat:
```
doctest.h:4510:28: warning: not eliding copy on return in 'doctest::String doctest::{anonymous}::translateActiveException()' [-Wnrvo]
 4510 |             return ex.what();
```
### Steps to reproduce
Include `"doctest.h"` and compile using gcc 15.1.1 with `-Wnrvo`.

### Reason
Compiler thinks we might want nrvo on variable `res`, but in this case we don't. Explicitly returning a `String` will remove warning.

https://github.com/doctest/doctest/blob/1da23a3e8119ec5cce4f9388e91b065e20bf06f5/doctest/parts/doctest.cpp#L1407-L1433

mitchgrout:
Just updated CI/CD to include `gcc-15` (15.2.0), but `-Wnrvo` doesn't trigger.

https://github.com/doctest/doctest/actions/runs/20654069875/job/59303615235?pr=1011

It _does_ however trigger for `clang-21` (21.1.7).

Just confirming this was indeed on gcc 15.1?

mitchgrout:
Closed by #1012

PR #929, opened 2025-07-19
Fix for gcc -Wnrvo warning

nylonhat:
## Description
Fix to remove  -Wnrvo warning for gcc 15.
Explicitly return a `String` to help compiler know we don't want nrvo. 
Does not functionally change anything.

## GitHub Issues
#928

### DIFF
--- a/doctest/parts/doctest.cpp
+++ b/doctest/parts/doctest.cpp
@@ -1410,7 +1410,7 @@ namespace {
         auto&  translators = getExceptionTranslators();
         for(auto& curr : translators)
             if(curr->translate(res))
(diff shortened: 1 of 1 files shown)


mitchgrout:
Makes more sense to just `-Wno-nrvo` it IMO. Also, `doctest.h` needs regeneration

mitchgrout:
Just bumping this change; also, good to rebase as CI/CD should be green now.

mitchgrout:
Closing in favour of #1012 - if you do prefer to fix this by actually employing NRVO feel free to re-open

PR #932, opened 2025-08-03
chore: Replace scripts/cmake/assemble_single_header.cmake with scripts/assemble.py

### LINKED ISSUE #914: Replace `assemble_single_header.cmake` with more fit-for-purpose script
We currently generate our single-header `doctest.h` file using [`scripts/cmake/assemble_single_header.cmake`](https://github.com/doctest/doctest/blob/bf90bf31782c6f0f671fed69cd7624b37a4189f5/scripts/cmake/assemble_single_header.cmake), which basically:

1. Loads `doctest_fwd.hpp` and `doctest.cpp`, then
2. Pastes them into an output file with some extra configuration set

This script is only needed for developers, since we check-in the generated `doctest.h` file; we could replace this instead with a more involved script, which will perform the following:

1. Discover (or be informed of) all relevant `doctest` sources and headers, then
2. Topologically sort them based on include dependencies, then
3. Strip `#include` statements which reference already-loaded `doctest` files, then
4. [Optionally] strip unnecessary comments, then
5. Paste files in-order into the output file

mitchgrout:
## Description

Adds `scripts/assemble.py` which provides the basis for breaking up the monolithic `doctest_fwd.h` and `doctest.cpp` files. No changes were made to `doctest_fwd.h` and `doctest.cpp` as part of this change. `doctest.h` changes slightly due to:

1. The autogenerated comment no longer referencing `cmake`, and
2. The `#include "doctest_fwd.h"` being stripped due to the intended behaviour of `assemble.py`

## GitHub Issues

Closes #914

### DIFF
--- a/scripts/assemble.py
+++ b/scripts/assemble.py
@@ -24,7 +24,9 @@ from textwrap import dedent
 from typing import Generator, NoReturn
 
 
-TEMPLATE = string.Template(dedent("""\
+TEMPLATE = string.Template(
+    dedent(
+        """\
   // ============================================================= lgtm [cpp/missing-header-guard]
   // == DO NOT MODIFY THIS FILE BY HAND - IT IS AUTO GENERATED! ==
   // =============================================================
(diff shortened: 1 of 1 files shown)


mitchgrout:
@severinstrobl mind having a quick look over this?

scripts/assemble.py
```diff
@@ -0,0 +1,134 @@
+#!/usr/bin/env python3
+
+"""
+Generate the monolithic doctest.h header file.
+
+Our original script, scripts/cmake/assemble_single_header.cmake,
```
severinstrobl: Nit: As the original script is deleted as part of this PR, I don't think it makes much sense to reference it here by name.

scripts/assemble.py
```diff
@@ -0,0 +1,134 @@
...
+def main(args: list[str]) -> NoReturn:
+  """Script entry-point."""
+
+  if len(args) != 1:
+    print("Usage: scripts/assemble.py", file=sys.stderr)
+    sys.exit(1)
+
+  script = Path(__file__).resolve()
+  root   = script.parent.parent
+  incdir = root / "doctest" / "parts"
```
severinstrobl:
```suggestion
  src_dir = root / "doctest" / "parts"
```

As we are also reading `.cpp` files.

scripts/assemble.py
```diff
@@ -0,0 +1,134 @@
...
+      2. Inlining header content, and
+      3. Scrubbing #include's to other doctest files
+    """
+
+    if file in visited:
+      return
+
+    visited.add(file)
+    with file.open("r") as fd:
+      content = fd.read()
```
severinstrobl: Maybe `content = file.read_text(encoding="utf-8")`?

scripts/assemble.py
```diff
@@ -0,0 +1,134 @@
...
+
+    for line in content.splitlines(keepends=False):
+      header = extract_header(line)
+      if header is not None:
+        if (incdir / header) in headers:
+          yield from process_file(incdir / header, visited=visited)
+        else:
+          yield line
```
severinstrobl: Can be combined into `if header is not None and (incdir / header) in headers:` to remove one `else`.

scripts/assemble.py
```diff
@@ -0,0 +1,134 @@
...
+      else:
+        yield line
+
+
+  # Would be a lot nicer with t-strings (Python 3.14)
+  # or jinja2, but it's best to avoid unnecessary dependencies,
+  # even for dev tooling
```
severinstrobl: How about something along the lines of:
```python
  template = string.Template(dedent("""\
  // ============================================================= lgtm [cpp/missing-header-guard]
  // == DO NOT MODIFY THIS FILE BY HAND - IT IS AUTO GENERATED! ==
  // =============================================================
  $headers
  #ifndef DOCTEST_SINGLE_HEADER
  #define DOCTEST_SINGLE_HEADER
  #endif // DOCTEST_SINGLE_HEADER
  $sources
  """))

  content = template.substitute(
    headers="\n".join(chain.from_iterable(process_file(file, visited=visited) for file in headers)),
    sources="\n".join(chain.from_iterable(process_file(file, visited=visited) for file in sources)),
  )

  with output.open("w") as sink:
    sink.write(content)
```

scripts/assemble.py
severinstrobl, line 1: You might want to run the entire file through `black` or `ruff`, [PEP8](https://peps.python.org/pep-0008/#indentation) strongly suggests 4 spaces for indentation.

scripts/assemble.py
mitchgrout: Was mainly thinking it would make locating the replacement script a bit easier, but I'm indifferent if we keep it in or not.

scripts/assemble.py
mitchgrout: Perhaps maintaing both `header_dir` and `source_dir` for clarity -- I think it could benefit us to move these files from `doctest/parts` to a standard layout, i.e. a `inc` and `src` directory.

scripts/assemble.py
mitchgrout: Just decided to rework the entire docstring to simplify it a bit

scripts/assemble.py
mitchgrout: That works; also, we can just use `output.write_text()` as well now

severinstrobl (approved):

mitchgrout:
Apparently `from __future__ import annotations` is a Python3.7 feature, and some of the containers are running Python3.6. The CI config seems a bit screwy, it seems to only consider changes on the most-recent commit when selecting jobs, so I didn't catch that.

I'll open up another PR tonight to try to fix that up.

PR #935, opened 2025-08-06
fix: Suppress -Wunsafe-buffer-usage for clang

### LINKED ISSUE #766: Tests don't compile on Clang 16 due to memory safety warnings with `-Werror`

mitchgrout:
## Description

Adds `-Wunknown-warning-option` and `-Wunsafe-buffer-usage` to the common-warnings list.
Since `-Wunsafe-buffer-usage` was added from clang-16 onwards, older versions of `clang` will (rightfully) warn if they see a suppression for it. 
`-Wunknown-warning-option` prevents this, but it is a bit of a blanket fix, since we won't get warnings for incorrect cases like `-Wtf`.
However, it seems that `gcc` is already configured this way, so maybe not the worst option?

Long-run, it would be good to reduce the number of warnings we're globally turning off, since it makes our claim of "compiles without warnings" a little less interesting.

## GitHub Issues

Closes #766

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -294,6 +294,12 @@ jobs:
           if [ "${{ matrix.compiler }}" = "gcc" ]; then
             sudo apt-get install -y g++-${{ matrix.version }} g++-${{ matrix.version }}-multilib
           elif [ "${{ matrix.compiler }}" = "clang" ]; then
+            # clang-17 does not automatically provide the i386 sanitizer runtimes,
+            # but libclang-rt-V-dev:i386 does not always exist. This is the best intermediate option
+            if [ "${{ matrix.version }}" = "17" ]; then
+              sudo apt-get install -y libclang-rt-${{ matrix.version }}-dev:i386
+            fi
+
             sudo apt-get install -y clang-${{ matrix.version }} libclang-${{ matrix.version }}-dev llvm-${{ matrix.version }}-dev g++-multilib
           fi
 
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -162,11 +162,13 @@
 #define DOCTEST_SUPPRESS_COMMON_WARNINGS_PUSH                                                      \
     DOCTEST_CLANG_SUPPRESS_WARNING_PUSH                                                            \
     DOCTEST_CLANG_SUPPRESS_WARNING("-Wunknown-pragmas")                                            \
+    DOCTEST_CLANG_SUPPRESS_WARNING("-Wunknown-warning-option")                                     \
     DOCTEST_CLANG_SUPPRESS_WARNING("-Wweak-vtables")                                               \
     DOCTEST_CLANG_SUPPRESS_WARNING("-Wpadded")                                                     \
     DOCTEST_CLANG_SUPPRESS_WARNING("-Wmissing-prototypes")                                         \
     DOCTEST_CLANG_SUPPRESS_WARNING("-Wc++98-compat")                                               \
     DOCTEST_CLANG_SUPPRESS_WARNING("-Wc++98-compat-pedantic")                                      \
+    DOCTEST_CLANG_SUPPRESS_WARNING("-Wunsafe-buffer-usage")                                        \
                                                                                                    \
     DOCTEST_GCC_SUPPRESS_WARNING_PUSH                                                              \
     DOCTEST_GCC_SUPPRESS_WARNING("-Wunknown-pragmas")                                              \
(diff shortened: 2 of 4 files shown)


mitchgrout:
@severinstrobl another if you have the time -- there was also #768 but that's 2+ years old.

As an aside, I'm surprsied that people are depending on the upstream to suppress warnings for them, when they could just `-Isystem` doctest, which AFAIK on all major compilers will suppress any warnings generated by the header.

severinstrobl:
> As an aside, I'm surprsied that people are depending on the upstream to suppress warnings for them, when they could just `-Isystem` doctest, which AFAIK on all major compilers will suppress any warnings generated by the header.

Well, there is a problem with using _doctest_ via  CMake, see my somewhat ancient PR #817 :grin:

That aside, I fully agree: disabling as many compiler warnings as _doctest_ currently does, "compiles without warnings" is a bit of a stretch.

mitchgrout:
Ah, looks like there's some extra examples/tests that need the lint-suppression applied as well, ech. I'll fix those up tonight, but hopefully in this change we can get `windows-*, clang-cl` green again

mitchgrout:
> > As an aside, I'm surprsied that people are depending on the upstream to suppress warnings for them, when they could just `-Isystem` doctest, which AFAIK on all major compilers will suppress any warnings generated by the header.
> 
> Well, there is a problem with using _doctest_ via CMake, see my somewhat ancient PR #817 😁
> 
> That aside, I fully agree: disabling as many compiler warnings as _doctest_ currently does, "compiles without warnings" is a bit of a stretch.

I think it could be good to resurrect this; my current (albeit slow) plan was to get CI back to passing, deal with build system stuff, then put out a 2.5.0 with those, then start looking at some extra features for a 2.6

severinstrobl:
> Ah, looks like there's some extra examples/tests that need the lint-suppression applied as well, ech. I'll fix those up tonight, but hopefully in this change we can get `windows-*, clang-cl` green again

I think the CI for Windows is horribly outdated, the referenced runner images probably don't even exist any more. I'll have a look and see how bad it is, maybe I can fix something up.

severinstrobl:
@mitchgrout: Would you mind rebasing this one on _dev_ so we can see if this fixes the remaining failures we see in the CI pipeline?

mitchgrout:
> @mitchgrout: Would you mind rebasing this one on _dev_ so we can see if this fixes the remaining failures we see in the CI pipeline?

Just snuck in a fix on `dev` for the Python3.6 problem, plus updated `common.cmake` on this branch to resolve the unsafe-buffer-usage warnings in examples. Hopefully we get Windows + clang + gcc passing now...

mitchgrout:
> @mitchgrout: Would you mind rebasing this one on _dev_ so we can see if this fixes the remaining failures we see in the CI pipeline?

@severinstrobl looks like we've resolved pretty much everything, except somehow this change broken `ubuntu-24.04 - clang-17`. I'll poke around in-container to see why, but past that, all that's left the coverage job. **Edit:** ~~Seems it was just a flaky run, which doesn't inspire confidence in GitHub's systems tbh. All that's broken is `coverage` now.~~ My bad -- the x64 job runs fine, but the x86 ubuntu-24.04 / clang-17 fails due to missing deps for some sanitizers. That's out-of-scope for this PR I think, but we should aim to resolve that soon.

As noted on the source issue, I think that we may need Viktor to do some project config, since from what I recall, we either need to set a secret token, _or_ change some setting in the project to allow tokenless coverage.

Alternatively, we could just drop `coverage` in favour of running `gcovr` in-pipeline. I never really understood why everyone decided to just do project analysis on external services when doing it locally is extremely straighforward. Maybe because you get a nice summary comment or something?

mitchgrout:
And with that, our pipeline (sans coverage) is passing! The changes to the CI are a bit jank but it gets things working again. If you have the chance @severinstrobl it would be good to have a look over that, but otherwise maybe we merge to dev soon-ish (~24 hours from now?)

From there, we can look at tidying up a few more bugs, then maybe release this all as a 2.4.13 /  2.5.0.

severinstrobl:
Getting the _Codecov_ action to work again would be my preferred option. Cobbling something together ourselves based on _gcov_/_lcov_ would certainly work, but I don't think we need to reinvent the wheel here. Having the automatic summary comments is handy, as is the web-based interface for the results of the coverage job.

PR #937, opened 2025-08-06
ci(macOS): use macOS 14 with XCode 15.4 for GCC

severinstrobl:
## Description

XCode on macOS 15 was updated, the default version is now Xcode 16 which is incompatible with GCC, so we need to switch to macOS 14 with still uses Xcode 15.

## GitHub Issues

Part of #871


mitchgrout:
LGTM

PR #938, opened 2025-08-10
chore: Remove type-hinting from scripts/assemble.py for Python3.6 compat

mitchgrout:
## Description

Removes type-hints from `scripts/assemble.py`, as `from __future__ import annotations` support was only added from Python3.7 onwards. We could re-add these in the future by bumping the Python version on the CI, but I'd like to stabilize the CI first before doing much hacking in there.

## GitHub Issues

#914

### DIFF
--- a/scripts/assemble.py
+++ b/scripts/assemble.py
@@ -13,15 +13,12 @@ the final doctest.h header.
 # ///
 
 
-from __future__ import annotations
-
 import re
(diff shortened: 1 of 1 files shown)


mitchgrout:
Tested locally via `python:3.6.15` docker-image and found no problem, should work on CI now 🤞

PR #940, opened 2025-08-12
ci: Remove coverage job

mitchgrout:
## Description

Removes the `coverage` job. In order to get this working, we'd need to configure the repository itself, which can only be done by owners. AFAIK, this is only Viktor who is not particularly active. Removing this, plus #935, brings the CI back to 100% passing.

We could look at using `gcovr` directly, although I'm not entirely sure how good of an experience that would be.  

## GitHub Issues

Part of #871


mitchgrout:
FYI @severinstrobl

mitchgrout:
@Saalvage not sure how active you are on this repo, but as an owner would you be able/willing to provision and configure a codecov token? If not, we may just yank the job entirely.

(ref. https://docs.codecov.com/docs/quick-start#step-2-get-the-repository-upload-token)

mitchgrout:
Unfortunate, but without a token AFAIK we can't run codecov. I'll merge this for now, but hopefully we can do _something_ for coverage at some stage. At least with this, we can enforce that MRs MUST pass pipelines, rather than having to just ignore a handful of failing jobs.

Issue #941, opened 2025-08-19
Simplify doctest_fwd.h and doctest.cpp


mitchgrout:
As per #600, one of the ideal goals is to chunk up `doctest_fwd.h` (and `doctest.cpp` by extension) into more manageable chunks. This is now technically possible via #932, which can scan for `.h` files in `parts/` and assemble the final single-header output.

This issue is primarily to track this task.

mitchgrout:
@Repiteo almost done with this. I'm still tossing up how to deal with macros though. I was considering just popping them into `doctest/macros.h` similar to what Catch2 does, but it _also_ makes some amount of sense to attach them to the translation unit/modules they are associated with.

For example, should `TEST_SUITE` originate from:

- `doctest/macros.h`
- `doctest/macros/test_suite.h`
- `doctest/test_suite.h`
- `doctest/test_suite/macros.h`
- Somewhere else?

They're currently loosely grouped by config, e.g. `DISABLE`, `NO_EXCEPTIONS`, `ASSERTS_RETURN_VALUES`, but some macros like `INFO`, or `TEST_CASE`, don't really care about those. Hence, maybe separating them would be a good move.

PR #944, opened 2025-08-20
chore: Extract warning-suppression macros to doctest/parts/warnings.h

mitchgrout:
## Description

Extracts warning-suppression macros `doctest_fwd.h` into `parts/warnings.h`

## GitHub Issues

Relates to #941

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -106,7 +106,6 @@
 #ifndef DOCTEST_ICC
 #define DOCTEST_ICC 0
 #endif // DOCTEST_ICC
-
 // =================================================================================================
 // == COMPILER WARNINGS HELPERS ====================================================================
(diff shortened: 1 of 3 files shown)


mitchgrout:
@severinstrobl I've starte chunking up `doctest_fwd.h` -- you interested in looking over the non-trivial bits?

In this change, just needed to ensure our `PUSH` / `POP` calls remain in `doctest_fwd.h`. For MSVC, pushing/popping in separate files seems to at least be a warning, if not a hard-error; but it also just makes sense.

I suspect we'll have to re-organize a few more files, or possibly split them into different logical units. But, in general I'll be aiming to keep `doctest.h` roughly the same.

severinstrobl (approved):

Issue #875, opened 2024-10-29
Plattform detection is broken for MacOS


nolange:
The platform detection is completely broken, apparently copied from Catch2 which [fixed it 7 years ago](https://github.com/catchorg/Catch2/commit/e3e6453229705ad9ac122c3a165538f4616d1571).
Dont know all side-effects, but one is certainly that isDebuggerActive uses the Linux implementation, which results in surprises if one changed that implementation to print warnings for unexpected errors:

https://github.com/nolange/doctest/actions/runs/11581408434/job/32242103784

cdeln:
Can you provide an MRE for this please?

nolange:
> Can you provide an MRE for this please?

Is this simple enough: https://github.com/nolange/doctest/commit/78862f322e7f4dd3d2e61112fac14cb699723706

I don't have a MAC, look at the CI runners from that commit.

Funny fact, you know the `DOCTEST_PLATFORM_MAC` is not working because the MAC Code errors out when you fix detection:

```
In file included from /Users/runner/work/doctest/doctest/scripts/development_only/doctest/doctest.h:5:
/Users/runner/work/doctest/doctest/doctest/parts/doctest.cpp:1497:60: error: zero as null pointer constant [-Werror,-Wzero-as-null-pointer-constant]
        if(sysctl(mib, DOCTEST_COUNTOF(mib), &info, &size, 0, 0) != 0) {
                                                           ^
                                                           nullptr
```

PR #947, opened 2025-09-07
Improve linux is debugger active 2

nolange:
## Description

Replace linux debugger detection by using POSIX io functions.
The C++ IO  functions don't help at all, at best they don't cause trouble.

This reduces dependencies and improves robustness.

Thanks to bug #875 in doctest, the MAC ci pipelines will fail.


mitchgrout:
Thanks for the pull-request. As this is your first time making changes on this repo, please have a read over our [contribution guidelines](https://github.com/doctest/doctest/blob/master/CONTRIBUTING.md#pull-requests). Some small housekeeping bits before I look over your changes fully:

- Make sure you have checked-in the generated `doctest.h` file (I'll eventually get around to making a CI job to enforce this)
- Ensure your pull-request body and title adequately describes your changes. If you are making significant changes (such as with this change), please also describe _why_ these changes are desirable if there's no source-issue tracking it
- Preferably prefix your title with a [conventional](https://www.conventionalcommits.org/en/v1.0.0-beta.2/#summary) verb like `fix`, `feat`, `chore`, etc
- Ensure you have referenced any and all relevant issues in your pull-request

I'll aim to have a proper look over these this evening, mid-week at latest.

mitchgrout:
I'm going to need a bit of convincing before accepting this change. The gist of it is, "use C standard I/O instead of std::ifstream", correct? 

> The C++ IO functions don't help at all, at best they don't cause trouble.

Gonna need some context here. 

> This reduces dependencies and improves robustness.

Similarly. Depending on the C standard library vs depending on the C++ standard library makes no difference to me. I see no reduction in dependencies, nor any real change in robustness.

nolange:
> I'm going to need a bit of convincing before accepting this change. The gist of it is, "use C standard I/O instead of std::ifstream", correct?

Direct access without stuff like buffering, seeking or character encoding. the "virtual" files are a bit special in that regard.

> Similarly. Depending on the C standard library vs depending on the C++ standard library makes no difference to me. I see no reduction in dependencies, nor any real change in robustness.

the point is that there is a minimal set of operations, and the the C++ abstractions can add stuff on top that:

- already requires a workaround for libc++ in your code with that `ErrnoGuard` which is explained in the [code taken from catch2](https://github.com/catchorg/Catch2/blob/v2.13.1/include/internal/catch_debugger.cpp#L79-L102)
- cause allocations
- cause buffering (and in the worst case seeking to 0 which will reset the proc file contents and potentially size)
- cause exceptions
 
 I dont see a single upside using `std::ifstream` but the added abstraction can affect robustness as the standard allows freedoms that can be problematic for `not-normal-files`. Realistically C++ libraries for Linux would be tested for that.
 
And if you want to use the break-if-debugging function in your code then the C++ abstraction will do alot more instead of breaking into your code ASAP. Potential exceptions further limit where you can use it (eg. Destructors are noexcept)

mitchgrout:
Sorry, I'm still not particularly sold on this. Yes, I understand that with C++ comes abstraction cost. The `errno` stuff is gross, but the consequence of an [implementation errata](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=79820). I'm sure if the same was found for a C function, we wouldn't be proposing writing ASM.

For the time being, I'm closing this. If there's enough want for it, or some demonstrable issue that arises from it, I'm happy to reconsider.

PR #948, opened 2025-09-07
redo OS detection, fix for MacOS

nolange:
## Description

On Mac the `DOCTEST_PLATFORM_LINUX` is defined and not `DOCTEST_PLATFORM_MAC`.

Fix this, and make the now active code compile.

### DIFF
--- a/doctest/parts/doctest.cpp
+++ b/doctest/parts/doctest.cpp
@@ -1473,15 +1473,16 @@ namespace detail {
         }
         return false;
     }
-#elif defined(DOCTEST_PLATFORM_MAC)
+#elif (defined(DOCTEST_PLATFORM_MAC) || defined(DOCTEST_PLATFORM_IPHONE)) &&                       \
+        defined(__apple_build_version__)
     // The following function is taken directly from the following technical note:
     // https://developer.apple.com/library/archive/qa/qa1361/_index.html
     // Returns true if the current process is being debugged (either
     // running under the debugger or has a debugger attached post facto).
     bool isDebuggerActive() {
-        int        mib[4];
-        kinfo_proc info;
-        size_t     size;
+        int               mib[4];
+        struct kinfo_proc info;
+        std::size_t       size;
         // Initialize the flags so that, if sysctl fails for some bizarre
         // reason, we get a predictable result.
         info.kp_proc.p_flag = 0;
@@ -1493,8 +1494,11 @@ namespace detail {
         mib[3] = getpid();
         // Call sysctl.
         size = sizeof(info);
-        if(sysctl(mib, DOCTEST_COUNTOF(mib), &info, &size, 0, 0) != 0) {
-            std::cerr << "\nCall to sysctl failed - unable to determine if debugger is active **\n";
+        if(sysctl(mib, sizeof(mib) / sizeof(*mib), &info, &size, nullptr, 0) != 0) {
+#ifndef DOCTEST_CONFIG_NO_INCLUDE_IOSTREAM
+            std::cerr << "\nCall to sysctl failed - unable to determine if debugger is "
+                         "active **\n";
+#endif
             return false;
         }
         // We're being debugged if the P_TRACED flag is set.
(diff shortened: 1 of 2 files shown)


doctest/parts/doctest.cpp
```diff
@@ -1473,15 +1473,16 @@ namespace detail {
         }
         return false;
     }
-#elif defined(DOCTEST_PLATFORM_MAC)
+#elif (defined(DOCTEST_PLATFORM_MAC) || defined(DOCTEST_PLATFORM_IPHONE)) &&                       \
+        defined(__apple_build_version__)
```
mitchgrout, line 1477: Why do we need this check?

doctest/parts/doctest.cpp
```diff
@@ -1473,15 +1473,16 @@ namespace detail {
...
     // https://developer.apple.com/library/archive/qa/qa1361/_index.html
     // Returns true if the current process is being debugged (either
     // running under the debugger or has a debugger attached post facto).
     bool isDebuggerActive() {
-        int        mib[4];
-        kinfo_proc info;
-        size_t     size;
+        int               mib[4];
+        struct kinfo_proc info;
+        std::size_t       size;
```
mitchgrout, line 1485: AFAIK shouldn't require explicit qualification to `std`

doctest/parts/doctest.cpp
```diff
@@ -1493,8 +1494,11 @@ namespace detail {
         mib[3] = getpid();
         // Call sysctl.
         size = sizeof(info);
-        if(sysctl(mib, DOCTEST_COUNTOF(mib), &info, &size, 0, 0) != 0) {
-            std::cerr << "\nCall to sysctl failed - unable to determine if debugger is active **\n";
+        if(sysctl(mib, sizeof(mib) / sizeof(*mib), &info, &size, nullptr, 0) != 0) {
```
mitchgrout, line 1497: Not sure why you're swapping from `DOCTEST_COUNTOF` in arg 2

doctest/parts/doctest.cpp
```diff
@@ -2661,11 +2665,15 @@ namespace {
...
                 std::tm timeInfo;
-#ifdef DOCTEST_PLATFORM_WINDOWS
+#if defined(_MSC_VER) || defined(__MINGW32__)
                 gmtime_s(&timeInfo, &rawtime);
-#else // DOCTEST_PLATFORM_WINDOWS
+#elif defined(__ORBIS__) || defined(__PROSPERO__)
+                gmtime_s(&rawtime, &timeInfo);
+#elif defined(__IAR_SYSTEMS_ICC__)
+                timeInfo = *std::gmtime(&rawtime);
+#else
```
mitchgrout, line 2674: Preferable to comment what the fallback cases on your `#else` are going to end up being

doctest/parts/doctest_fwd.h
```diff
@@ -239,18 +239,30 @@ DOCTEST_MSVC_SUPPRESS_WARNING(4623) // default constructor was implicitly define
...
+#elif defined(WIN32) || defined(__WIN32__) || defined(_WIN32) || defined(_MSC_VER) ||              \
+        defined(__MINGW32__)
+#define DOCTEST_PLATFORM_WINDOWS
+
+#if defined(WINAPI_FAMILY) && (WINAPI_FAMILY == WINAPI_FAMILY_APP)
+#define DOCTEST_PLATFORM_WINDOWS_UWP
+#endif
+#elif defined(__ORBIS__) || defined(__PROSPERO__)
+#define DOCTEST_PLATFORM_PLAYSTATION
+#endif
```
mitchgrout, line 265: I see no point in including support for these, esp. without test support for regression

mitchgrout (changes requested):
Some small houskeeping bits:
- Preferably have a commit per logical change made -- you've at least:
    1. Refactored platform support, and
    2. Fixed platform detection for Mac, and
    3. Fixed broken code for Mac, and
    4. Added support for extra platforms
- If there's no linked GitLab issue, either N/A or just remove the section
- Need to regenerated `doctest.h`

doctest/parts/doctest.cpp
nolange: Taken verbatim from Catch2, and per C++ Standard including `<cstddef>` (etc) is meant to add `std::size_t`, and whether `size_t` is defined too is up to the implementation.

doctest/parts/doctest.cpp
nolange: Copied verbatim from Catch2. I'd argue this would ease up maintenance, given that alot doctest code seems to originate from there

doctest/parts/doctest_fwd.h
nolange: Again, to be easy comparable to Catch2.
And to be annoying: this PR fixes your support of `DOCTEST_PLATFORM_MAC` which apparently never worked, whereas `__ORBIS__` support mightve even be tested on Catch2.

doctest/parts/doctest.cpp
nolange: 1. cause your old stuff did not work
2. I dont even own an MAC
3. I copied stuff from Catch2 which does work (I believe I tried removing the check and then it did not)

Digging throug Catch2, apparently the code only compiles with `AppleClang` and that what the check is for:
[Catch2 code](https://github.com/catchorg/Catch2/blob/3bcd0a4e74b2bb5f6ec80ad464775442af84343c/src/catch2/internal/catch_debugger.cpp#L21)

mitchgrout:
Just bumping this; CI/CD should be green now, and will require a little rebase work due to https://github.com/doctest/doctest/issues/941.

mitchgrout:
Superseded by #1057

PR #949, opened 2025-09-07
chore: Reorder isDebuggerActive calls to be performed as late as possible

nolange:
## Description

The check for a attaching debugger is executed often, even though the result
could be ignored later.
Reorder the checks so the check is not run if not needed


mitchgrout:
Seems fine, but looks like some test output needs to be updated. I'll have a look over again and merge once you're done with that

mitchgrout:
Just bumping this; CI/CD should be green now, and will require a little rebase work due to #941

PR #953, opened 2025-09-24
CMake: update minimim required to 3.14

nim65s:
Hi,

With CMake v4, we get an annoying warning about the `cmake_minimum_required(VERSION 3.5)` being too old.

Bumping this to 3.10 would be fine, but I noticed that other things were waiting for more that that, like the `ARCH_INDEPENDENT` setting.

For context, CMake v3.14 was released in 2019. Debian oldoldstabe has CMake v3.18, and ubuntu 20.04 (which is EOL) has CMake v3.16. So I guess 3.14 should be fine for 99% users.

While here I upgraded this number in other places, because those `cmake_minimum_required(VERSION 3.0)` (or even 2.8) must but updated too before they start breaking things everywhere. We can lower those to 3.10 if you prefer.

And last, I did sneak in some documentation updates on the same topic. That commit can easily be removed from this PR and put on another one if you prefer.


nim65s:
Ok, the job of the third commit was already mostly handled in #866, thanks :)

nim65s:
Hum, CI displayed a fun issue with clang adding `/usr/local/include` by default then failing because that path is not safe 🤯 
(only on macos/xcode)

mitchgrout:
This changes a few too many things in one fells swoop for my liking. We've already bumped to v3.5 (#811) which appears to be sufficient for now. If there is a firm deadline for when CMake breaks compat. for v3.5, then we can consider increasing it further.

I'll close this in a week or so unless there's any extra info

G4m4:
Hi! Doctest user here. I'm trying with this message to tip the scales in favor of merging this PR (we would do it locally anyway). Considering that doctest is a test framework, I would definitely appreciate that it works out of the box without any friction.

This used to be a warning but starting with Cmake 4.0 it is [an error](https://cmake.org/cmake/help/latest/release/4.0.html#deprecated-and-removed-features).
This is especially painful on MacOS where brew relies on recent versions of CMake ([4.2.1 at the time of writing this](https://formulae.brew.sh/formula/cmake)).

mitchgrout:
Given it's a hard-error on real systems, we can bump to a new minimum. 

This MR is unfortunately out-of-date, and with no response in 2 weeks I assume the author may not be able to update it for us. Are you happy to whip up a new MR with the required changes @G4m4?

G4m4:
Thanks for the quick reply and happy to! Here it is https://github.com/doctest/doctest/pull/1023

nim65s:
I'm still around if needed :)

mitchgrout:
> I'm still around if needed :)

If you're happy to get this rebased we can go with this. As a heads-up, you can remove this diff:

https://github.com/doctest/doctest/pull/953/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR26-R28

...since it should already be resolved by #963

nim65s:
That is right, a rebase was enough, thanks :)

CMakeLists.txt
```diff
@@ -28,8 +24,11 @@ option(DOCTEST_INTERNAL_COVERAGE "Set for running coverage" OFF)
 add_library(${PROJECT_NAME} INTERFACE)
 add_library(${PROJECT_NAME}::${PROJECT_NAME} ALIAS ${PROJECT_NAME})
 
-if(NOT CMAKE_VERSION VERSION_LESS 3.8)
-    target_compile_features(${PROJECT_NAME} INTERFACE cxx_std_11)
+target_compile_features(${PROJECT_NAME} INTERFACE cxx_std_11)
+
+if(APPLE)
+    # /usr/local/include is added by default then rejected as unsafe with xcode
+    target_compile_options(${PROJECT_NAME} INTERFACE -Wno-poison-system-directories)
```
mitchgrout: Remove

nim65s: thanks, done

mitchgrout (approved):

PR #958, opened 2025-10-24
Fix dll interface export for the filldata::fill symbol

grandemk:
## Description

I have a link error when I use the DOCTEST_CONFIG_IMPLEMENTATION_IN_DLL way of using doctest.

```
undefined symbol: public: static void __cdecl doctest::detail::filldata<void const *>::fill(class std::basic_ostream<char, struct std::char_traits<char>> *, void const *)
```

The template is fully specialized here, so the code will reside in the dll and the user of the code, `REQUIRE(pointer != nullptr)` will need to see the definition and know that the static function is defined in the dll.

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -1192,12 +1192,12 @@ DOCTEST_MSVC_SUPPRESS_WARNING_POP
 
     template <>
     struct filldata<const void*> {
-        static void fill(std::ostream* stream, const void* in);
+        DOCTEST_INTERFACE static void fill(std::ostream* stream, const void* in);
     };
 
     template <>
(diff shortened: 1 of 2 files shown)


mitchgrout:
Thanks for the change. A few requests:

1. Please make sure `doctest.h` is updated and checked-in as per `CONTRIBUTING.md`
2. Preferably use a more descriptive MR title/commit message

Mac failures appear to be for different reasons, so don't worry about those

grandemk:
Hi, thanks for the feedback,
I 've updated the MR / commit messages and the doctest.h has been commited

mitchgrout:
LGTM

PR #959, opened 2025-10-27
Split compilation jobs according to the os

Alex-PLACET:
## Description
To continue the work on the CI, to simplify the maintenance, here I split the compilation jobs according to the OS

### DIFF
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -1,4 +1,4 @@
-name: CI
+name: CI Linux
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


mitchgrout:
Tossing up whether it's worth breaking it down by both platform + compiler -- for `ci-linux.yml` in particular, we have some grotty conditionals to grab different packages, depending on whether we're intending on running `gcc` or `clang`.

Breaking them into separate configs would result in a bit of duplication w.r.t grabbing `cmake` -- thoughts?

mitchgrout:
Closed in favour of #1064

PR #960, opened 2025-10-27
Enforce POSIX newlines & UTF-8 encoding

Repiteo:
## Description
Explicitly specifies newlines as POSIX and encoding as UTF-8 in more locations, particularly in generation scripts. I develop on a Windows machine, and I kept getting false-positive changes for `doctest.h` because of newlines being converted to `CRLF`.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,1 @@
-* text=auto
+* text=auto eol=lf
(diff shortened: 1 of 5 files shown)


Repiteo:
Looks like `write_text()` only got `newline` as an option in 3.10, so I swapped it out for the `with open()` syntax

mitchgrout:
LGTM

PR #962, opened 2025-10-28
chore: Extract feature detection to `doctest/parts/feature_detection.h`

Repiteo:
## Description
Continues the split of `doctest_fwd.h` into chunks with the "feature detection" section being split into its own file. Added `#include <doctest/parts/compiler.h>` to the top of the file to ensure the header alone doesn't churn out false-positive errors; this had no impact on builds and still resulted in an identical `doctest/doctest.h` output

## GitHub Issues
Helps address #941


mitchgrout:
Thanks for the change; unfortunately, I'm about halfway through reworking `doctest_fwd` already! I just haven't pushed changes yet in case I need to rework commits. It's been a little while since I've been away for the last month or so, but hopefully I'll be finished with it soon.

Repiteo:
Oh, that's great news! Looking forward to seeing it when it's ready (or a draft if you want to showcase it sooner)

PR #964, opened 2025-10-28
ci: Bump `checkout` to v5 & `codeql-action` to v4

Repiteo:
## Description

Both of these actions use very outdated versions on the repo, both of which produced warnings for me. No changes needed to be made elsewhere as a result of this bump


mitchgrout:
LGTM

PR #965, opened 2025-11-14
fix executable_dll_and_plugin test on cygwin

corngood:
## Description

I've been bringing up cygwin support in nixpkgs, and found that this particular test failed due to the incorrect shared lib path.  With this change, the whole test suite passes.


mitchgrout:
Thanks for the PR. Just a few house-keeping bits first:

* Ensure your PR is targeting `dev`, not `master` (see [CONTRIBUTING.md](https://github.com/doctest/doctest/blob/master/CONTRIBUTING.md#pull-requests))
* Likely just due to your being based off `master` at the moment, but ensure your CI is passing

@ or request review from me once we're ready!

corngood:
@mitchgrout There's one CI failure, but it doesn't look like it could be related.  What do you think?

mitchgrout:
My guess is they've once again updated the Mac images on us. I'll look to fix that soonish. For now, change LGTM.

PR #970, opened 2025-12-16
chore: Remove scripts/development_only/doctest.h

mitchgrout:
## Description

Removes the `scripts/development_only/doctest.h` header, and adjusts the CMake config to use the genuine `doctest.h`

Seemingly the intent of this was to have better source-reporting, but after #914 is implemented, we can just directly compile each source-part, making this redundant.

## GitHub Issues

#941

### DIFF
--- a/scripts/hello_world.cpp
+++ b/scripts/hello_world.cpp
@@ -1,5 +1,5 @@
 #define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
-#include "doctest.h"
+#include "doctest/doctest.h"
 
 int fact(int n) {
     return n <= 1 ? n : fact(n - 1) * n;
--- a/scripts/how_stuff_works/testing_crash_scenarios.cpp
+++ b/scripts/how_stuff_works/testing_crash_scenarios.cpp
@@ -1,5 +1,5 @@
 #define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
-#include "doctest.h"
+#include "doctest/doctest.h"
 
 DOCTEST_MAKE_STD_HEADERS_CLEAN_FROM_WARNINGS_ON_WALL_BEGIN
 #include <iostream>
--- a/scripts/playground/main.cpp
+++ b/scripts/playground/main.cpp
@@ -1,5 +1,5 @@
 #define DOCTEST_CONFIG_IMPLEMENT
-#include "doctest.h"
+#include "doctest/doctest.h"
 
 int main(int argc, char** argv) {
     doctest::Context context;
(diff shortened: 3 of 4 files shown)


mitchgrout:
I'll make sure CI/CD is green before merging, looks like some weird issue with the Windows runner that might just be a transient issue. Also managing to crash clang on Ubuntu which is fun; I've had that happen a few times locally but haven't bothered to dig into _why_ it happens yet

mitchgrout:
> I'll make sure CI/CD is green before merging, looks like some weird issue with the Windows runner that might just be a transient issue. Also managing to crash clang on Ubuntu which is fun; I've had that happen a few times locally but haven't bothered to dig into _why_ it happens yet

Well, that's fucking annoying. The `assemble_single_header` target is getting run while the compiler's trying to lex the source file, and it's triggering failures as a result. Unsure why this hasn't broke the CI/CD before.

**EDIT:** Obvious in retrospect, but all tests/examples were loading the `development_only` copy of `doctest.h`, which is just a constant file. So, no issues there. The genuine `doctest/doctest.h` file _was_ being assembled, but in parallel with all other sources being compiled, so there was no strong dependency between actually making the file, and tests beginning to be compiled. As far as I can see, there isn't an easy way to enforce this either; all attempts thus far have failed. However, with #971 we can catch this at the PR level pretty easily, so the simplest option for now is to just scrap the whole `custom_target` shenanigans.

A new fun problem is, swapping to `doctest/doctest.h`:

1. Breaks `clang-tidy` in real fun ways, and
2. Drops our code-coverage by 11%

...so that's something to be resolved also.

**EDIT:** `clang-tidy` breakage is likely because the file being analysed is now _just_ `doctest.h`, whereas before it was `doctest.h => doctest.cpp`. So, we're seeing a bunch of breakages related to things like, "symbol defined in header" which wouldn't trigger before, since `clang-tidy` would contextually be in a `.cpp` file.

Repiteo:
I don't consider the code coverage or clang-tidy issues as blockers tbh. This exposed a hole in how the file was evaluated, but I would think addressing that is better suited for followup PRs which actually handle the separate files. Failing that, there *could* be a dedicated clang-tidy pass PR which this could integrate, but code coverage will probably have to bite the bullet no matter what.

mitchgrout:
> I don't consider the code coverage or clang-tidy issues as blockers tbh. This exposed a hole in how the file was evaluated, but I would think addressing that is better suited for followup PRs which actually handle the separate files. Failing that, there _could_ be a dedicated clang-tidy pass PR which this could integrate, but code coverage will probably have to bite the bullet no matter what.

Yeah, separate PR for clang-tidy I reckon. It'll probably just be suppressions

PR #971, opened 2025-12-16
ci: Add task to check for stale doctest.h files

mitchgrout:
## Description

Adds `.github/workflows/code-generation.yml` which provides a `check-generated-header` job. The intent of this job is to check whether the `doctest.h` file would change if re-generated. This aims to catch pull-requests which have directly modified the `doctest.h` header without actually updating the constituent `parts/*` files.

## GitHub Issues

N/A

### DIFF
--- a/.github/workflows/code-generation.yml
+++ b/.github/workflows/code-generation.yml
@@ -1,0 +1,35 @@
+name: Code Generation
+
+on:
+  push:
+    branches: [dev, master]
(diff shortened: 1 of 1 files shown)


mitchgrout:
For a little extra verification, intentionally pushed a change to `doctest.h` itself, and observed a failed `check-generated-header` job:
image

PR #972, opened 2025-12-16
ci: Simplify conditional job configuration to support internal PRs

mitchgrout:
## Description

Resolving an issue that occurs when developing directly on `doctest/doctest`. When pushing a branch, we trigger the `on:push` rule, which is not restricted to any branches. When a pull-request is created, it also creates jobs but all of these are skipped. This results in a weird summary on pull-requests:

image

To resolve this, I updated the CI/CD config to use `on:push:branches` to restrict job creation to only `dev` and `master`, and removed the `if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository` snippet entirely. This _should_ have the effect of running jobs when...

1. We push to `dev` / `master`, i.e. after a pull-request is accepted
2. We create or update a pull-request, even from an internal branch

Behaviour was confirmed by checking that no actions were created upon initial push of this branch, but were created after the pull-request was created. This is in contrast to #971 where the actions were created upon push, then _again_ on pull-request creation (even though they were all skipped).

## GitHub Issues

N/A

### DIFF
--- a/.github/workflows/ci-mingw-w64.yml
+++ b/.github/workflows/ci-mingw-w64.yml
@@ -2,11 +2,13 @@ name: CI MinGW-w64
 
 on:
   push:
+    branches: [dev, master]
     paths-ignore:
       - "doc/**"
       - "scripts/**"
       - "LICENSE.txt"
       - "README.md"
+
   pull_request:
     paths-ignore:
       - "doc/**"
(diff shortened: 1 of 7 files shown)


mitchgrout:
Also confirming that the expected number of jobs is generated, based off counting the possible configs in .github/workflows/*.yml and comparing against expected runs on latest `dev` commit.

PR #973, opened 2025-12-17
chore: Prepare project for separation of monolithic .cpp code

mitchgrout:
## Description

Moves existing `parts/*.h` files into `parts/public/` and `doctest.cpp` into `parts/private/`. The reasoning here is that currently, `doctest.cpp` has quite a lot of inline helpers (for example, `Atomic<T>`) which need to be usable across various future source files. The best way to achieve this is to create private headers, e.g. `doctest/parts/private/atomic.h`. Without a proper public/private distinction, we may end up pulling these into the _public_ part of the `doctest.h` header.

Also moves some global information (namely include-guards, includes, and warning suppressions) from `doctest.cpp` into `scripts/assemble.py`. This is more of a temporary thing; it allows for code to be more easily lifted out into auxilliary `.cpp` files without having to worry about also bringing error suppression with it. As these `.cpp` files aren't _really_ intended to be compiled in isolation (they're in a "parts" directory after all), this is fine, but it would be good in the future to eliminate this by making the `.cpp` files into genuine source files.

## GitHub Issues

#941

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -3089,6 +3089,7 @@ DOCTEST_SUPPRESS_COMMON_WARNINGS_POP
 #ifndef DOCTEST_SINGLE_HEADER
 #define DOCTEST_SINGLE_HEADER
 #endif // DOCTEST_SINGLE_HEADER
+
 #if defined(DOCTEST_CONFIG_IMPLEMENT) || !defined(DOCTEST_SINGLE_HEADER)
 
 #ifndef DOCTEST_SINGLE_HEADER
@@ -3096,10 +3097,8 @@ DOCTEST_SUPPRESS_COMMON_WARNINGS_POP
 #endif // DOCTEST_SINGLE_HEADER
 
 DOCTEST_CLANG_SUPPRESS_WARNING_WITH_PUSH("-Wunused-macros")
-
 #ifndef DOCTEST_LIBRARY_IMPLEMENTATION
 #define DOCTEST_LIBRARY_IMPLEMENTATION
-
 DOCTEST_CLANG_SUPPRESS_WARNING_POP
 
 DOCTEST_SUPPRESS_COMMON_WARNINGS_PUSH
@@ -3887,7 +3886,7 @@ const char* skipPathFromFilename(const char* file) {
         {
             const auto prefix_start = pos;
             pos = std::min(prefixes.find(separator, prefix_start), prefixes.size());
-            
+
             const auto prefix_size = pos - prefix_start;
             if(prefix_size > longest_match)
             {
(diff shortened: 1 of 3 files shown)


mitchgrout:
Depends on !970 to get rid of the development `doctest.h` header

Repiteo:
This shouldn't need #970 as a dependency, as we can just pass the altered `doctest.cpp` path to the development header
```diff
diff --git a/scripts/development_only/doctest/doctest.h b/scripts/development_only/doctest/doctest.h
index 9fbd512..f409c6b 100644
--- a/scripts/development_only/doctest/doctest.h
+++ b/scripts/development_only/doctest/doctest.h
@@ -2,4 +2,4 @@
 #ifndef DOCTEST_SINGLE_HEADER
 #define DOCTEST_SINGLE_HEADER
 #endif
-#include "doctest.cpp"
+#include "private/doctest.cpp"

```
Obviously removing the development header should still happen, but this PR isn't blocked by that

mitchgrout:
> This shouldn't need #970 as a dependency, as we can just pass the altered `doctest.cpp` path to the development header
> 
> ```diff
> diff --git a/scripts/development_only/doctest/doctest.h b/scripts/development_only/doctest/doctest.h
> index 9fbd512..f409c6b 100644
> --- a/scripts/development_only/doctest/doctest.h
> +++ b/scripts/development_only/doctest/doctest.h
> @@ -2,4 +2,4 @@
>  #ifndef DOCTEST_SINGLE_HEADER
>  #define DOCTEST_SINGLE_HEADER
>  #endif
> -#include "doctest.cpp"
> +#include "private/doctest.cpp"
> ```
> 
> Obviously removing the development header should still happen, but this PR isn't blocked by that

Yeah, it'd work, but with global warning suppression being moved into `scripts/assemble.py` it would result in most jobs failing due to warnings being triggered. Mostly just sequencing it this way to keep the CI happy.

Repiteo:
Shouldn't the global warning suppressions be put in its own private header instead? Repeat headers are already ignored during the assembling process, so it'd allow those source files to still have proper warning support. Though, if this is a temporary measure anyway, it's probably not that big of a deal

mitchgrout:
> Shouldn't the global warning suppressions be put in its own private header instead? Repeat headers are already ignored during the assembling process, so it'd allow those source files to still have proper warning support. Though, if this is a temporary measure anyway, it's probably not that big of a deal

`clang` in particular is a bit pedantic about warning-suppression. If a warning is suppressed in one file, it must also be _unsuppressed_ in the same file. I was originally planning on having a global file for all sources to include, but ran into this problem.

My current plan is:

1. Hoist warning-suppression into `assemble.py`, so that all files still 'see' the same suppressions
2. Once everything's broken down into parts, start moving suppressions into individual files, thus removing them from `assemble.py`
3. Long-run, try to eliminate as many warnings by actually addressing issues in source, rather than suppressing

Repiteo:
Ahh, that's fair. Could maybe still work with header guard shenanigans, but it makes much more sense to only dig into that *after* the public/private split

PR #982, opened 2025-12-21
chore: Extract string-related code to doctest/parts/string.h

mitchgrout:
## Description

Extracts `doctest::String` and related stringification code to `doctest/parts/*/string.{h,cpp}`.

This also includes `stringifyBinaryExpr` which is practically a stringification method, but is mostly only used by `Result` and `Expression`. The `tlss{Push,Pop}` functions are also declared in `string.h`, but are not _yet_ implemented in `string.cpp` due to a reliance on `ContextState`, which in turn requires `String` to be defined. This will be resolved in a later change.

## GitHub Issues

#941


doctest/parts/public/to_string.h
```diff
@@ -0,0 +1,233 @@
...
+    template <typename T>
+    struct has_insertion_operator<T, decltype(operator<<(declval<std::ostream&>(), declval<const T&>()), void())> : types::true_type { };
+
+    template <typename T>
+    struct should_stringify_as_underlying_type {
+        static DOCTEST_CONSTEXPR bool value = detail::types::is_enum<T>::value && !doctest::detail::has_insertion_operator<T>::value;
+    };
+
+    DOCTEST_INTERFACE std::ostream* tlssPush();
+    DOCTEST_INTERFACE String tlssPop();
```
mitchgrout: This is a bit of an annoying one, `tlssPush / tlssPop` are used pretty much only by stringification (except `MessageBuilder` which is stringification in its own right), _but_ the implementation relies on `ContextState *g_cs` which in turn requires `ContextOptions` which requires... `doctest::String`.

So, I've opted to have the forward-declaration present here, but the implementation remaining in `doctest.cpp`, up until I get to dealing with the `Context*` family of classes.

mitchgrout:
I organized this as `string.h` and `to_string.h` simply because that's the order in which I got to them, but they also make sense as a single header, i.e. both in `string.h`. Any preferences?  ( @cdeln @Repiteo )

doctest/parts/private/string.cpp
```diff
@@ -0,0 +1,226 @@
...
+        setLast();
+    }
+
+    String::~String() {
+        if(!isOnStack())
+            delete[] data.ptr;
+    } // NOLINT(clang-analyzer-cplusplus.NewDeleteLeaks)
+
+    DOCTEST_CLANG_SUPPRESS_WARNING_WITH_PUSH("-Wshorten-64-to-32")
+    DOCTEST_GCC_SUPPRESS_WARNING_WITH_PUSH("-Wconversion")
```
mitchgrout: Ah, should remove this; this was real early on, when I didn't have global error suppression going.

doctest/parts/private/string.cpp
```diff
@@ -0,0 +1,226 @@
...
+namespace {
+    // case insensitive strcmp
+    static int stricmp(const char* a, const char* b) {
+        for(;; a++, b++) {
+            const int d = tolower(*a) - tolower(*b);
+            if(d != 0 || !*a)
+                return d;
+        }
+    }
+} // namespace
```
Repiteo: This'll fix the `clang-tidy` error; an anonymous namespace isn't necessary for a static function
```suggestion
    // case insensitive strcmp
    static int stricmp(const char* a, const char* b) {
        for(;; a++, b++) {
            const int d = tolower(*a) - tolower(*b);
            if(d != 0 || !*a)
                return d;
        }
    }
```

doctest/doctest.h
```diff
@@ -634,120 +634,356 @@ namespace detail {
 
 } // namespace detail
 } // namespace doctest
+#include "doctest/parts/private/prelude.h"
```
Repiteo: Not strictly a blocker, but I wonder why this header wasn't trimmed in the generation process

mitchgrout:
> Implementation looks good! Not sure why the checks against `windows.h` are only now failing, but it can be suppressed in `prelude.h` like this:
> 
> ```c++
> DOCTEST_CLANG_SUPPRESS_WARNING_WITH_PUSH("-Wnonportable-system-include-path")
> #include <windows.h>
> DOCTEST_CLANG_SUPPRESS_WARNING_POP
> ```

Yeah, I suspected it was a tooling bump on the `windows-2025` CI image, but there's nothing in their `CHANGELOG` to indicate that. Plus, all jobs passed on `dev` after merging your `#undef` tweaks. My guess is they're _doing_ a tooling bump, and we've just kinda hit it halfway through?

**EDIT:** Nope, it's just https://github.com/doctest/doctest/pull/982#discussion_r2638091103 - `prelude.h` was loaded as part of the header, so it didn't see the warning-suppressions for source-files, which includes `-Wnonportable-system-include-path`

doctest/doctest.h
mitchgrout: 🤦‍♂️ yeah, that's in `public/string.h`, that'd be why. I must've mistook it for `string.cpp` while double-checking everything hah.

PR #988, opened 2025-12-22
chore: Extract assertion management to `doctest/parts/assert/`

mitchgrout:
## Description

Extracts most assertion code into the following files:

- `doctest/parts/assert/type.h`: Provides the main `assertType` enumeration
- `doctest/parts/assert/data.h`: Provides the common `AssertData` structure
- `doctest/parts/assert/comparator.h`: Provides utility comparators used by later files
- `doctest/parts/assert/result.h`: Provides the `ResultBuilder` and `Result` structures
- `doctest/parts/assert/expression.h`: Provides the `ExpressionDecomposer` structure
- `doctest/parts/assert/handler.h`: Provides the core `unary_assert` and `binary_assert` handlers

Some parts of the `ResultBuilder` implementation remain in `doctest.cpp`, since they currently depend on macros like `ITERATE_THROUGH_REPORTERS` which are currently inaccessible. This will be corrected when we get to extracting the reporters module.

The include for `doctest/parts/assert/handler.h` in `doctest_fwd.h` has to be placed in a slightly awkward spot, since:

1. It requires `checkIfShouldThrow` to be defined, which
2. Requires `assertType` to be defined

This will be tidied up a bit next when exceptions are extracted.

## GitHub Issues

#941

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,6 +68,7 @@ if(${DOCTEST_WITH_MAIN_IN_STATIC_LIB})
         ${doctest_parts_folder}/private/doctest.cpp
         ${doctest_parts_folder}/private/assert/data.cpp
         ${doctest_parts_folder}/private/assert/expression.cpp
+        ${doctest_parts_folder}/private/assert/handler.cpp
         ${doctest_parts_folder}/private/assert/result.cpp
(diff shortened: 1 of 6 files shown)


mitchgrout:
cc @Repiteo - one of the larger changes, interested in your thoughts on the current file organization

PR #1001, opened 2025-12-24
chore: Extract macros to doctest/parts/macros.h

mitchgrout:
## Description

Extracts all test macros to `doctest/parts/macros.h`

## GitHub Issues

#941


mitchgrout:
cc @Repiteo - for now, this might be the easiest since it has the least diff with `doctest.h`. We could consider the best strategy for chunking it up into something a bit more manageable in the future. This'll at least let us close out #941 soon-ish.

Repiteo (approved):
Exciting that this almost completely removes the need for `doctest_fwd.h`! There's absolutely room for this to be further segmented across other files, but I don't consider that a blocker for the reasons you've already stated.

Issue #1000, opened 2025-12-24
Ensure library compiles from parts


mitchgrout:
As mentioned in a few different PRs associated with #941, it would be good if we can actually compile the library from the `parts/` folder. We will still primarily distribute the library in assembled form.

mitchgrout:
We've got a viable candidate with `doctest_with_main`, closing for now

mitchgrout:
Re-opening as #1020 revealed some issues with the `doctest_with_main` target

PR #1002, opened 2025-12-24
feat: Have segments function as standalone files

Repiteo:
## Description
An initial pass at ensuring the individual part segments can function as standalone header & source files. There's a few ways that this aims to implement the separate logic:
- Include header include guards in every `doctest/parts/` header.
- Add new warning suppression wrappers for public & private files, each expanding on the common suppressions with those found in `doctest_fwd.h` and `assembly.py` respectively. These too are added to any file producing warnings.

The bulk of work has been done on the `public` section in an attempt to minimize header includes from the start. It's likely that `private` can expand on this, removing a need for the dedicated `prelude.h` if we can rely on a small handful of public headers to only include the standard library as needed.

## GitHub Issues
- Implements (partially) #1000
- Depends on #941


mitchgrout:
Personally prefer standard header inclusion styles (e.g. `#if !defined(...)`) since you don't have to do any special casing or stripping during single header assembly, but we can always add extra logic to it as required

Repiteo:
I think there'll always be some degree of extra logic, as we have to account for stripping local includes. But header guards *would* account for distinguishing files, so I can absolutely try that!

Repiteo:
Integrated header guards via defines instead of pragma & removed the expanded assembly logic to best showcase their implementation

mitchgrout:
A few small things; I'm testing this using the `doctest_with_main` target that we have available. There are a few uses of `g_cs`, `Timer`, `XmlWriter` etc. which either don't see:

* `using detail::<name>`
* `using namespace detail`

...in the translation-unit (or any included header), so will currently fail. We can either fix them here, or I can patch them later on. I'm gonna update `build_and_test.py` to actually build this target so we can ensure we don't break it moving forward.

Repiteo:
This was more of an icebreaker PR than an attempt to get into the weeds of private dependencies. Once this is merged, it'll be far easier to isolate those source files for specific headers only. It's likely we could even drop `prelude.h` entirely, as the majority of library includes would only need to be associated with their relevant source files

I'd say this can be merged as-is, with other fixes to come shortly after

PR #1005, opened 2025-12-26
chore: Run clang-format across segmented files

Repiteo:
## Description
Applies the current clang-format settings across all current segmented files. Posting as a draft, as this is more of a proof-of-concept. While the ability to actually apply these changes will be very beneficial, there's two major considerations that need to be accounted for first:
- Updating `.clang-format`. A lot of the options specified are rather unconventional, and weren't actually utilized by the repository in the overwhelming majority of cases. A pass applying clang-format would ideally go hand-in-hand with one that decides on what rules we should apply
- Missing headers/metadata. The lack of headers and/or explicitly specified metadata are likely causing formatting quirks. The headers will be accounted for via #1000, but metadata would need to be part of the same pass that updates the clang-format rules

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -55,9 +55,8 @@
 #define DOCTEST_TOSTR(x) DOCTEST_TOSTR_IMPL(x)
 
 #define DOCTEST_VERSION_STR                                                                        \
-    DOCTEST_TOSTR(DOCTEST_VERSION_MAJOR) "."                                                       \
-    DOCTEST_TOSTR(DOCTEST_VERSION_MINOR) "."                                                       \
-    DOCTEST_TOSTR(DOCTEST_VERSION_PATCH)
+    DOCTEST_TOSTR(DOCTEST_VERSION_MAJOR)                                                           \
+    "." DOCTEST_TOSTR(DOCTEST_VERSION_MINOR) "." DOCTEST_TOSTR(DOCTEST_VERSION_PATCH)
 
 #define DOCTEST_VERSION                                                                            \
     (DOCTEST_VERSION_MAJOR * 10000 + DOCTEST_VERSION_MINOR * 100 + DOCTEST_VERSION_PATCH)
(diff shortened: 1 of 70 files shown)


mitchgrout:
Regarding your first point, should we just say fuck it and swap to a built-in style like `LLVM`? The current .clang-format file is all over the place, and doesn't consider more recently added options. Probably easiest to pick a current style that most-closely matches what we have.

For your second point, I'm whipping up an MR that adds in all required `#include`'s to all `.h` and `.cpp` files (companion to #1002), which should help.

We should also consider all of the `clang-format: off` comments that are currently in the codebase. Most of them could just be removed, since they don't really do fuck all, but some (e.g. `context.cpp`) are nice to have.

**Edit:** Just looking at the current built-in styles for `clang-format-21`, they're all quite different to what we have (although, even our current style doesn't match .clang-format very well). We could continue with using our own style, but we just need to be mindful of keeping `clang-format` (1.) part of the CI/CD checks, and (2.) up-to-date. My personal preference if going this route is to actually specify _every legal configuration option_ (or at least comment why we're not setting it), since it makes it easier to know what needs to be considered when bumping `clang-format` to a new version.

Repiteo:
> My personal preference if going this route is to actually specify _every legal configuration option_ (or at least comment why we're not setting it), since it makes it easier to know what needs to be considered when bumping `clang-format` to a new version.

We do something similar in the [Godot repo](https://github.com/godotengine/godot/blob/master/.clang-format), where LLVM is our "baseline" to keep the rules we change distinguished. We could emulate that implementation here, albeit with different altered rules

mitchgrout:
> My personal preference if going this route is to actually specify _every legal configuration option_ (or at least comment why we're not setting it), since it makes it easier to know what needs to be considered when bumping `clang-format` to a new version.
> 
> We do something similar in the [Godot repo](https://github.com/godotengine/godot/blob/master/.clang-format), where LLVM is our "baseline" to keep the rules we change distinguished. We could emulate that implementation here, albeit with different altered rules

I'm happy to go with this. Do you want to whip up the change, or shall I?

I'll also be setting up a CI/CD job shortly (to merge after this) just to keep style in sync, just in case you're also working on that.

Repiteo:
I'll go ahead and set that up!

Repiteo:
Superseded by #1015

PR #1009, opened 2025-12-29
chore: Finalize extraction and remove doctest.cpp

### LINKED ISSUE #941: Simplify doctest_fwd.h and doctest.cpp

mitchgrout:
## Description

Removes all remaining implementation code from `doctest/parts/doctest.cpp`, and removes the file. 

## GitHub Issues

Closes #941

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,7 +2,7 @@
 Make sure the PR is against the dev branch and not master which contains
 the latest stable release. Also make sure to have read CONTRIBUTING.md.
 Please do not submit pull requests changing the single-include `doctest.h`
-file, it is generated from the 2 files in the doctest/parts/ folder.
+file, it is generated from the files in the doctest/parts/ folder.
 -->
 
 
@@ -15,7 +15,7 @@ was that you have a large set of tests and found that this speeds them up.
 -->
 
 ## GitHub Issues
-<!-- 
+<!--
 If this PR was motivated by some existing issues, reference them here.
 
 If it is a simple bug-fix, please also add a line like 'Closes #123'
--- a/.github/workflows/code-generation.yml
+++ b/.github/workflows/code-generation.yml
@@ -32,8 +32,7 @@ jobs:
         run: |
           if ! git diff --exit-code doctest/doctest.h; then
             echo "::error file=doctest/doctest.h title=doctest.h out-of-date::doctest.h differs when regenerated"
-            echo "Ensure you have only changed files in the doctest/parts/* directory"
-            echo "and either build via CMake, or run the scripts/assemble.py script directly"
-            echo "to assemble the single-header include form of doctest"
+            echo "Ensure you have only changed files in the doctest/parts/* directory,"
+            echo "then run scripts/assemble.py to assemble the single-header file for doctest"
             exit 1
           fi
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,9 +68,13 @@ if(${DOCTEST_WITH_MAIN_IN_STATIC_LIB})
         ${doctest_parts_folder}/private/assert/data.cpp
         ${doctest_parts_folder}/private/assert/expression.cpp
         ${doctest_parts_folder}/private/assert/handler.cpp
+        ${doctest_parts_folder}/private/assert/message.cpp
         ${doctest_parts_folder}/private/assert/result.cpp
         ${doctest_parts_folder}/private/assert/type.cpp
         ${doctest_parts_folder}/private/color.cpp
+        ${doctest_parts_folder}/private/context.cpp
+        ${doctest_parts_folder}/private/context/options.cpp
+        ${doctest_parts_folder}/private/context_scope.cpp
         ${doctest_parts_folder}/private/debugger.cpp
         ${doctest_parts_folder}/private/exception_translator.cpp
         ${doctest_parts_folder}/private/exceptions.cpp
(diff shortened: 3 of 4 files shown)


mitchgrout:
@Repiteo just at the finish line now. `doctest_fwd.h` has to stick around for a little bit (maybe forever?) to ensure we actually get all our public headers in. One thing I'm not 100% on how we should deal with is this chunk from [doctest_fwd.h](https://github.com/doctest/doctest/blob/941-finalize/doctest/parts/doctest_fwd.h#L92-L97):

```cpp

// this is here to clear the 'current test suite' for the current translation unit - at the top
DOCTEST_TEST_SUITE_END();

#endif // DOCTEST_CONFIG_DISABLE
```

Just thinking about #1000, it'd be good to ensure that wherever it goes, it still gets included by anyone trying to actually use the library. We could maybe put it in `test_suite.h` which is included by `test_case.h` which in turn is included by `macros.h`, which should be OK?

**Edit:** Given the current separation, it'd probably have to be `test_suite/something.h` and included _after_ `macros.h`. I'd rather have all symbols declared by the time we start declaring macros, so moving `test_suite.h` _after_ `macros.h` feels wrong. Alternatively, we could just inline the expansion of the macro; when I first saw it, I assumed something was missing, since an `_END()` on its own just looks... weird

Repiteo:
`doctest_fwd.h` only needs to stick around for however long it takes for the segmented files to properly handle header dependencies. If anything, I could rework #1002 to remove that file once this PR is merged. That's because not every segmented file will need to rely on all the public headers; only the ones relevant for that specific implementation are needed

doctest/parts/public/test_suite.h
```diff
@@ -35,6 +35,17 @@ DOCTEST_INTERFACE int setTestSuite(const TestSuite& ts);
...
+
+#ifndef DOCTEST_CONFIG_DISABLE
+
+// this is here to clear the 'current test suite' for the current translation unit - at the top
+DOCTEST_GLOBAL_NO_WARNINGS( /* NOLINT(cert-err58-cpp) */
+    DOCTEST_ANONYMOUS(DOCTEST_ANON_VAR_),
+    doctest::detail::setTestSuite(doctest::detail::TestSuite() * "")
+);
+
+#endif // DOCTEST_CONFIG_DISABLE
```
Repiteo: No need for the `DOCTEST_CONFIG_DISABLE` wrapper, it's already applied to the file itself
```suggestion
// this is here to clear the 'current test suite' for the current translation unit - at the top
DOCTEST_GLOBAL_NO_WARNINGS( /* NOLINT(cert-err58-cpp) */
    DOCTEST_ANONYMOUS(DOCTEST_ANON_VAR_),
    doctest::detail::setTestSuite(doctest::detail::TestSuite() * "")
);
```

doctest/parts/private/reporters.cpp
```diff
@@ -0,0 +1,4 @@
+#include "doctest/parts/private/reporters/xml.h"
+#include "doctest/parts/private/reporters/junit.h"
+#include "doctest/parts/private/reporters/console.h"
+#include "doctest/parts/private/reporters/debug_output_window.h"
```
Repiteo, line 4: File could be removed if the generator is adjusted to scan for private headers as well as sources, but not something this PR needs to worry about

Repiteo (approved):
The end is in sight

doctest/doctest.h
```diff
@@ -1736,6 +1736,17 @@ DOCTEST_INTERFACE int setTestSuite(const TestSuite& ts);
...
 namespace doctest_detail_test_suite_ns {
 DOCTEST_INTERFACE doctest::detail::TestSuite& getCurrentTestSuite();
+
+#ifndef DOCTEST_CONFIG_DISABLE
+
+// this is here to clear the 'current test suite' for the current translation unit - at the top
+DOCTEST_GLOBAL_NO_WARNINGS( /* NOLINT(cert-err58-cpp) */
+    DOCTEST_ANONYMOUS(DOCTEST_ANON_VAR_),
+    doctest::detail::setTestSuite(doctest::detail::TestSuite() * "")
+);
```
Repiteo: Fixes the `-Wpedantic` warnings
```suggestion
)
```

doctest/parts/private/reporters.cpp
mitchgrout: My plan is to create a proper `.cpp` file per each one of these; they're all defined inline at the moment, so that gives a natural solution to this problem

mitchgrout:
> `doctest_fwd.h` only needs to stick around for however long it takes for the segmented files to properly handle header dependencies. If anything, I could rework #1002 to remove that file once this PR is merged. That's because not every segmented file will need to rely on all the public headers; only the ones relevant for that specific implementation are needed

The main benefit of having a dedicated file like this is that it ensures that the generated `doctest.h` remains at least sort-of coherent. At the moment, `assemble.py` will process files in alphabetical order (plus topological, where required). So `assert/comparator.h` which show up in the generated file before `version.h`, which could be a bit weird.

We could consider just moving the `parts/` into a standard `inc/` and `src/` layout, in which case `doctest_fwd.h` could just be `inc/doctest.h`, which makes for a nice consistent experience (e.g. including single-header `doctest.h` is (mostly) the same as including demonolithed-library `inc/doctest.h`).

One small thing, it's _possible_ that consumers could be depending on `doctest_fwd.h` to exist, although nothing in the documentation says it's correct to include that file. Just pays to be mindful when removing it.

Repiteo:
> The main benefit of having a dedicated file like this is that it ensures that the generated `doctest.h` remains at least sort-of coherent. At the moment, `assemble.py` will process files in alphabetical order (plus topological, where required). So `assert/comparator.h` which show up in the generated file before `version.h`, which could be a bit weird.

That's one benefit of having dedicated headers on the segmented files though, as it would largely resolve dependency order for those specialized files. Though, if keeping a 100% strict order is desired, there's also the option of simply moving those ordered files to the assembler itself; that'd make the glob portion more of a sanity-test to ensure everything is properly included.

> We could consider just moving the `parts/` into a standard `inc/` and `src/` layout, in which case `doctest_fwd.h` could just be `inc/doctest.h`, which makes for a nice consistent experience (e.g. including single-header `doctest.h` is (mostly) the same as including demonolithed-library `inc/doctest.h`).

I could see that working, yeah. Wouldn't rush into that straight away though, as we probably don't want to overhaul the new folder structure so soon.

> One small thing, it's _possible_ that consumers could be depending on `doctest_fwd.h` to exist, although nothing in the documentation says it's correct to include that file. Just pays to be mindful when removing it.

While that *might* be true, we shouldn't be beholden to undocumented behavior. With that said, whatever direction *is* taken **should** have proper documentation after everything settles down.

PR #1010, opened 2026-01-01
ci: Build `doctest_with_main` in CI

mitchgrout:
## Description

Ensures `doctest_with_main` is compiled along with tests, and fixes issues preventing it from compiling. 

## GitHub Issues

#1000

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -46,11 +46,13 @@ jobs:
       - name: LCOV
         run: |
           mkdir coverage
-          lcov -c -d build/ -o coverage/lcov.info --include "*doctest/parts*"
+          lcov -c -d build/ -o coverage/lcov.info --include "*/doctest/doctest.h"
 
       - name: Codecov
         uses: codecov/codecov-action@v5
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           files: ./coverage/lcov.info
+          disable_search: true
+          plugins: noop
           fail_ci_if_error: true
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -256,6 +256,7 @@
     DOCTEST_CLANG_SUPPRESS_WARNING("-Wmissing-braces")                                             \
     DOCTEST_CLANG_SUPPRESS_WARNING("-Wmissing-field-initializers")                                 \
     DOCTEST_CLANG_SUPPRESS_WARNING("-Wunused-member-function")                                     \
+    DOCTEST_CLANG_SUPPRESS_WARNING("-Wunused-function")                                            \
     DOCTEST_CLANG_SUPPRESS_WARNING("-Wnonportable-system-include-path")                            \
                                                                                                    \
     DOCTEST_GCC_SUPPRESS_WARNING("-Wconversion")                                                   \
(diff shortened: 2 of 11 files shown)


.github/workflows/coverage.yml
```diff
@@ -46,11 +46,13 @@ jobs:
...
-          lcov -c -d build/ -o coverage/lcov.info --include "*doctest/parts*"
+          lcov -c -d build/ -o coverage/lcov.info --include "*/doctest/doctest.h"
 
       - name: Codecov
         uses: codecov/codecov-action@v5
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           files: ./coverage/lcov.info
+          disable_search: true
+          plugins: noop
```
mitchgrout, line 57: This is a weird one; apparently, `codecov/codecov-action` will actually run `gcov` _by-default_. So, not only does it take our `lcov.info` file, it also scans it again, and fuses those two. Basically, we were getting extra (bad) coverage from the `parts/` directory, plus the `examples` and `scripts/playground` files. 

`plugins: noop` turns off the `gcov` stuff, and apparently `disable_search: true` may be necessary also. This brings us back up to 91%, where we were before. I've updated the project limits to be >= 90%

CMakeLists.txt
```diff
@@ -64,7 +64,7 @@ endif()
 
 if(${DOCTEST_WITH_MAIN_IN_STATIC_LIB})
     add_library(
-        ${PROJECT_NAME}_with_main STATIC EXCLUDE_FROM_ALL
+        ${PROJECT_NAME}_with_main STATIC
```
mitchgrout, line 67: Probably for the best that we just start building this by-default

doctest/doctest.h
```diff
@@ -4130,7 +4131,7 @@ DOCTEST_SUPPRESS_PRIVATE_WARNINGS_PUSH
 #ifndef DOCTEST_CONFIG_DISABLE
 
 namespace doctest {
-namespace {
+inline namespace _ {
```
mitchgrout, line 4134: This is a screwy one. For gcc-5 _specifically_, a function defined in an anonymous-namespace that is also _unused_ cannot be inhibited via `-Wunused-function`. Seemingly, it's being done as part of a global wrap-up pass in the compiler, so none of this really works.

For now, I'm just cramming it into an `inline namespace` since that has different semantics. Long-term, I'll be pulling the implementation out into a `.cpp` file which should _generally_ resolve it.

Might also pay to just replace our anonymous namespaces with `namespace doctest::impl` or something, just to avoid having to do too much weird shit in `assemble.py`

.github/workflows/coverage.yml
Repiteo: Don't use codecov actions much in my projects so I wouldn't have caught this. Worth re-examining how other repos handle this integration down the road, in case lcov logic can be passed to this job directly

doctest/doctest.h
Repiteo: An anonymous namespace in a header file is already a bad idea, so addressing it in *some* form is for the best. If per-compilation unit concerns become a thing with the separate files, there's always the option of declaring the relevant functions as static; not sure if we've actually run across that issue in practice

Repiteo (approved):
The fact that everything just *worked* before despite this absurd namespace jumbling is a miracle in its own right. This further justifies the decision to separate the individual files, as this would've been a **much** bigger burden to track down otherwise. Any issues I have with these changes would be better served to individual file overhauls; nothing blocking here!

Issue #774, opened 2023-05-05
`-Werror` considered harmful


jktjkt:
**TL;DR**: please consider just enabling `-Werror` in your CI, but do *not* do that in CMake so that people who are building your stuff *see* the warnings and can decide whether to *act* on them. Chances are that they're using a different toolchain from you.

Every now and then, a new compiler adds a new type of warning. The most recent example which hit us was clang's very pedantic idea about *any* C-style buffers being considered unsafe (#766). These warnings are of course *not* enabled by default, not even by `-Wall` or `-Wextra` because they flag some perfectly valid code. Indeed, upstream itself [considers these warnings to be useful for projects which are willing to switch to an improved, hardened memory model](https://discourse.llvm.org/t/rfc-c-buffer-hardening/65734#programming-model-4). As such, these warnings are useful tools, but they are *not* recommended for every C++ project.

Unfortunately, doctest currently hard-enables *all* possible compiler warnings, *and* it activates them for anyone building the project by default. I think that this is a very bad idea, and I'm not alone; there's, for example, an [explicit policy in Gentoo](https://devmanual.gentoo.org/ebuild-writing/common-mistakes/index.html#-werror-compiler-flag-not-removed), and [this blog post makes an excellent point about its harmfulness](https://embeddedartistry.com/blog/2017/05/22/werror-is-not-your-friend/), while also pointing out how one can keep all the benefits.

PR #1014, opened 2026-01-04
chore: Disable werror outside of CI/CD

mitchgrout:
## Description

Quality-of-life change. `-Werror` is only enabled by-default when building on pipeline (but may still be optionally enabled in local dev environments).

Tested by quickly inducing a warning (specifically, an unused parameter warning) and checking that:

* With a regular `cmake ...` setup, warnings do not block the build, and
* When `-D DOCTEST_INTERNAL_WERROR=ON` is set, warnings become errors

## GitHub Issues

#774


mitchgrout:
cc @Repiteo - just the minimal change required. I'll look at de-cluttering the CMake config prior to 2.5 since it's a bit messy in places.

PR #1015, opened 2026-01-04
feat: Sync `clang-format` with LLVM

Repiteo:
- Related: #1005 

## Description
Applies the discussed direction from the above PR to `.clang-tidy`, granting us an LLVM baseline. Adjustments were made to represent the de-facto state of the repository, so not all rules were transfered over if their implementation was rare or incorrect. Keeping as a draft so that certain revealed elements can be further discussed and potentially adjusted/removed as necessary:
- Only formatter used atm is `clang-tidy`, meaning a lot of vestigal code coverage checkers remain (lgtm, OCLINT, etc). While not outright inhibiting, they *are* revealing their clutter more than ever
- Do we want to continue enforcing a 100-width limit? The majority of clang-format ignores that existed (now removed) were because this limit was constantly breached by macros. We can continue to selectively ignore if desired, or set the limit to `0` to drop the hardcap entirely

Almost every file is being changed anyway, so we aren't beholden to existing formatting styles in the first place. Though it'd obviously be preferable to have whatever noise *is* caused by this to only happen once, so we should wholly agree on the direction taken early

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,82 +1,246 @@
-# https://clang.llvm.org/docs/ClangFormatStyleOptions.html
-
-DisableFormat: false
-Language: Cpp
-Standard: Cpp11
-IndentWidth: 4
-TabWidth: 4
-UseTab: Never
-ColumnLimit: 100
-ReflowComments: false
-SortIncludes: false
-
-AlignConsecutiveAssignments: true
-AlignConsecutiveDeclarations: true
-AlignEscapedNewlinesLeft: false
-AlignOperands: true
-AlignTrailingComments: true
-AlignAfterOpenBracket: true
-DerivePointerAlignment: false
-PointerAlignment: Left
-IndentCaseLabels: true
-ContinuationIndentWidth: 8
-NamespaceIndentation: Inner
-CompactNamespaces: true
-FixNamespaceComments: true
+# Commented out parameters are those with the same value as base LLVM style.
+# We can uncomment them if we want to change their value, or enforce the
+# chosen value in case the base style changes (last sync: Clang 18.1.8).
+BasedOnStyle: LLVM
 AccessModifierOffset: -4
-
-SpaceAfterControlStatementKeyword: false
-SpaceAfterCStyleCast: false
-SpaceBeforeAssignmentOperators: true
-SpaceBeforeParens: Never
-SpaceInEmptyParentheses: false
-SpacesBeforeTrailingComments: 1
-SpacesInAngles: false
-SpacesInCStyleCastParentheses: false
-SpacesInContainerLiterals: false
-SpacesInParentheses: false
-SpacesInSquareBrackets: false
-Cpp11BracedListStyle: true
-
-KeepEmptyLinesAtTheStartOfBlocks: false
-MaxEmptyLinesToKeep: 1
-BinPackArguments: true
-BinPackParameters: true
-AlwaysBreakAfterReturnType: None
-AlwaysBreakAfterDefinitionReturnType: None
-AlwaysBreakTemplateDeclarations: true
-BreakConstructorInitializersBeforeComma: true
-ConstructorInitializerAllOnOneLineOrOnePerLine: false
-ConstructorInitializerIndentWidth: 8
-
-AllowShortIfStatementsOnASingleLine: false
-AllowShortLoopsOnASingleLine: false
-AllowShortBlocksOnASingleLine: true
+AlignAfterOpenBracket: BlockIndent
+# AlignArrayOfStructures: None
+# AlignConsecutiveAssignments:
+#   Enabled: false
+#   AcrossEmptyLines: false
+#   AcrossComments: false
+#   AlignCompound: false
+#   AlignFunctionPointers: false
+#   PadOperators: true
+# AlignConsecutiveBitFields:
+#   Enabled: false
+#   AcrossEmptyLines: false
+#   AcrossComments: false
+#   AlignCompound: false
+#   AlignFunctionPointers: false
+#   PadOperators: false
+# AlignConsecutiveDeclarations:
+#   Enabled: false
+#   AcrossEmptyLines: false
+#   AcrossComments: false
+#   AlignCompound: false
+#   AlignFunctionPointers: false
+#   PadOperators: false
+# AlignConsecutiveMacros:
+#   Enabled: false
+#   AcrossEmptyLines: false
+#   AcrossComments: false
+#   AlignCompound: false
+#   AlignFunctionPointers: false
+#   PadOperators: false
+AlignConsecutiveShortCaseStatements:
+  Enabled: true
+  # AcrossEmptyLines: false
+  # AcrossComments: false
+  # AlignCaseColons: false
+AlignEscapedNewlines: Right
+# AlignOperands: Align
+# AlignTrailingComments:
+#   Kind: Always
+#   OverEmptyLines: 0
+# AllowAllArgumentsOnNextLine: true
+# AllowAllParametersOfDeclarationOnNextLine: true
+# AllowBreakBeforeNoexceptSpecifier: Never
+AllowShortBlocksOnASingleLine: Always
 AllowShortCaseLabelsOnASingleLine: true
-AllowAllParametersOfDeclarationOnNextLine: true
-AllowShortFunctionsOnASingleLine: All
-
-BreakBeforeBinaryOperators: false
-BreakBeforeTernaryOperators: false
-BreakStringLiterals: true
-BreakBeforeBraces: Custom
-BraceWrapping:
-    AfterClass: true
-    AfterEnum: true
-    AfterNamespace: false
-    AfterStruct: true
-    AfterUnion: true
-    
-    BeforeCatch: false
-    BeforeElse: false
-    IndentBraces: false
-    AfterFunction: false
-    AfterControlStatement: false
-
-# penalties not thought of yet
-PenaltyBreakBeforeFirstCallParameter: 19
-PenaltyBreakComment: 60
-PenaltyBreakString: 1000
-PenaltyBreakFirstLessLess: 120
-PenaltyExcessCharacter: 1000000
-PenaltyReturnTypeOnItsOwnLine: 1000
+# AllowShortCompoundRequirementOnASingleLine: true
+# AllowShortEnumsOnASingleLine: true
+AllowShortFunctionsOnASingleLine: Empty
+# AllowShortIfStatementsOnASingleLine: Never
+# AllowShortLambdasOnASingleLine: All
+# AllowShortLoopsOnASingleLine: false
+# AlwaysBreakAfterDefinitionReturnType: None
+# AlwaysBreakAfterReturnType: None
+# AlwaysBreakBeforeMultilineStrings: false
+AlwaysBreakTemplateDeclarations: Yes
+# AttributeMacros:
+#   - __capability
+BinPackArguments: false
+BinPackParameters: false
+# BitFieldColonSpacing: Both
+# BraceWrapping:
+#   AfterCaseLabel: false
+#   AfterClass: false
+#   AfterControlStatement: Never
+#   AfterEnum: false
+#   AfterFunction: false
+#   AfterNamespace: false
+#   AfterObjCDeclaration: false
+#   AfterStruct: false
+#   AfterUnion: false
+#   AfterExternBlock: false
+#   BeforeCatch: false
+#   BeforeElse: false
+#   BeforeLambdaBody: false
+#   BeforeWhile: false
+#   IndentBraces: false
+#   SplitEmptyFunction: true
+#   SplitEmptyRecord: true
+#   SplitEmptyNamespace: true
+# BreakAdjacentStringLiterals: true
+# BreakAfterAttributes: Leave
+# BreakAfterJavaFieldAnnotations: false
+# BreakArrays: true
+# BreakBeforeBinaryOperators: None
+# BreakBeforeBraces: Attach
+# BreakBeforeConceptDeclarations: Always
+# BreakBeforeInlineASMColon: OnlyMultiline
+# BreakBeforeTernaryOperators: true
+# BreakConstructorInitializers: BeforeColon
+# BreakInheritanceList: BeforeColon
+# BreakStringLiterals: true
+ColumnLimit: 120
+# CommentPragmas: "^ IWYU pragma:"
+# CompactNamespaces: false
+# ConstructorInitializerIndentWidth: 4
+# ContinuationIndentWidth: 4
+# Cpp11BracedListStyle: true
+# DerivePointerAlignment: false
+# DisableFormat: false
+# EmptyLineAfterAccessModifier: Never
+# EmptyLineBeforeAccessModifier: LogicalBlock
+# ExperimentalAutoDetectBinPacking: false
+# FixNamespaceComments: true
+# ForEachMacros:
+#   - foreach
+#   - Q_FOREACH
+#   - BOOST_FOREACH
+# IfMacros:
+#   - KJ_IF_MAYBE
+# IncludeBlocks: Preserve
+# IncludeCategories:
+#   - Regex: ^"(llvm|llvm-c|clang|clang-c)/
+#     Priority: 2
+#     SortPriority: 0
+#     CaseSensitive: false
+#   - Regex: ^(<|"(gtest|gmock|isl|json)/)
+#     Priority: 3
+#     SortPriority: 0
+#     CaseSensitive: false
+#   - Regex: .*
+#     Priority: 1
+#     SortPriority: 0
+#     CaseSensitive: false
+# IncludeIsMainRegex: (Test)?$
+# IncludeIsMainSourceRegex: ""
+# IndentAccessModifiers: false
+# IndentCaseBlocks: false
+IndentCaseLabels: true
+# IndentExternBlock: AfterExternBlock
+# IndentGotoLabels: true
+# IndentPPDirectives: None
+# IndentRequiresClause: true
+IndentWidth: 4
+# IndentWrappedFunctionNames: false
+# InsertBraces: false
+# InsertNewlineAtEOF: false
+# InsertTrailingCommas: None
+# IntegerLiteralSeparator:
+#   Binary: 0
+#   BinaryMinDigits: 0
+#   Decimal: 0
+#   DecimalMinDigits: 0
+#   Hex: 0
+#   HexMinDigits: 0
+# JavaScriptQuotes: Leave
+# JavaScriptWrapImports: true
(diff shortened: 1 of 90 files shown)


.clang-format
```diff
@@ -1,82 +1,293 @@
...
+# # AlignOperands: DontAlign
+# # ConstructorInitializerIndentWidth: 8
+# # ContinuationIndentWidth: 8
+# # IndentCaseLabels: true
+# # IndentWidth: 4
+# # KeepEmptyLinesAtTheStartOfBlocks: false
+# # PointerAlignment: Left
+# # SortIncludes: false
+# # Standard: c++11
+# # TabWidth: 4
```
mitchgrout: I think you left these in by accident

doctest/parts/private/assert/handler.h
```diff
@@ -12,38 +12,38 @@ DOCTEST_SUPPRESS_PRIVATE_WARNINGS_PUSH
...
 
-    using detail::g_cs;
+using detail::g_cs;
 
-    void addAssert(assertType::Enum at) {
-        if((at & assertType::is_warn) == 0) //!OCLINT bitwise operator in conditional
-            g_cs->numAssertsCurrentTest_atomic++;
-    }
+void addAssert(assertType::Enum at) {
+    if ((at & assertType::is_warn) == 0) //! OCLINT bitwise operator in conditional
```
mitchgrout: `//!` will be special comments, we may just have to inhibit the one-space-after-comment rule if there's no way to special case it

https://oclint-docs.readthedocs.io/en/stable/howto/suppress.html

As an aside, kinda weird we have a suppression comment in non-macro code, I'll have to see why oclint is complaining later on.

doctest/parts/private/assert/result.cpp
```diff
@@ -10,62 +10,62 @@ DOCTEST_SUPPRESS_PRIVATE_WARNINGS_PUSH
...
+void ResultBuilder::translateException() {
+    m_threw     = true;
+    m_exception = translateActiveException();
+}
+
+bool ResultBuilder::log() {
+    if (m_at & assertType::is_throws) { //! OCLINT bitwise operator in conditional
+        m_failed = !m_threw;
+    } else if ((m_at & assertType::is_throws_as) &&
+               (m_at & assertType::is_throws_with)) { //! OCLINT bitwise operator in conditional
```
mitchgrout: Similarly here, the `//!` probably originally applied to both expressions

doctest/parts/private/assert/type.cpp
```diff
@@ -5,19 +5,20 @@ DOCTEST_SUPPRESS_PRIVATE_WARNINGS_PUSH
...
-// clang-format off
 const char* assertString(assertType::Enum at) {
-    DOCTEST_MSVC_SUPPRESS_WARNING_WITH_PUSH(4061) // enum 'x' in switch of enum 'y' is not explicitly handled
-    #define DOCTEST_GENERATE_ASSERT_TYPE_CASE(assert_type) case assertType::DT_ ## assert_type: return #assert_type
-    #define DOCTEST_GENERATE_ASSERT_TYPE_CASES(assert_type) \
-        DOCTEST_GENERATE_ASSERT_TYPE_CASE(WARN_ ## assert_type); \
-        DOCTEST_GENERATE_ASSERT_TYPE_CASE(CHECK_ ## assert_type); \
-        DOCTEST_GENERATE_ASSERT_TYPE_CASE(REQUIRE_ ## assert_type)
+    DOCTEST_MSVC_SUPPRESS_WARNING_WITH_PUSH(
+            4061) // enum 'x' in switch of enum 'y' is not explicitly handled
```
mitchgrout: This is a little ugly; presumably it's because we don't have reflowed-comments. Maybe in cases like this, we just move the comment above

.clang-format
```diff
@@ -1,82 +1,293 @@
...
+# PenaltyBreakComment: 300
+# PenaltyBreakFirstLessLess: 120
+# PenaltyBreakOpenParenthesis: 0
+# PenaltyBreakScopeResolution: 500
+# PenaltyBreakString: 1000
+# PenaltyBreakTemplateDeclaration: 10
+# PenaltyExcessCharacter: 1000000
+# PenaltyIndentedWhitespace: 0
+# PenaltyReturnTypeOnItsOwnLine: 60
+PointerAlignment: Left
```
mitchgrout: I personally prefer right-alignment, only because left-alignment is misleading for consecutive declarations: `int* x, y` - granted, we should never _have_ those, but it sticks with me.

mitchgrout (commented):
I haven't checked every file, but I don't think there'll be a particularly huge number of weird violations I haven't seen. We can always tweak the config later on, or just rework the code to avoid the weird formatting.

.clang-format
Repiteo: Had that fixed locally but forgot to push, oops

.clang-format
Repiteo: Right-alignment is what LLVM already uses, so that'd keep things simple

doctest/parts/private/assert/type.cpp
Repiteo: I'll just trim the comment, as that's what's already used elsewhere

doctest/parts/private/assert/handler.h
Repiteo: OCLINT isn't actually checked by the repo atm afaik, so there's the option of removing these comments altogether. Same goes for a few other style-checking comments, but I'm not sure if any have the same multiline issue

.clang-format
Repiteo: The main thing this ended up changing are sections where alignment is enforced:
```cpp
struct DOCTEST_INTERFACE AssertData {
    // common - for all asserts
    const TestCaseData *m_test_case;
    assertType::Enum    m_at;
    const char         *m_file;
    int                 m_line;
    const char         *m_expr;
    bool                m_failed;
```
I'd say that's entirely acceptable, but it does beg the question if we want alignment enforced to begin with

doctest/parts/private/assert/handler.h
Repiteo: There indeed was a rule to allow `//!OCLINT` cpmments to be untouched: `CommentPragmas`

doctest/parts/private/assert/result.cpp
Repiteo: While I was able to prevent comment pragmas from being split, I don't know if I can prevent the rest of the line from breaking. In this case I shortened it to the broader `//!OCLINT`, but that isn't guaranteed to always work

Almost every exclusion for OCLINT is "bitwise operator in conditional"; in case we're still using OCLINT or want to use it in the future, we could just add this to a global ignore equivalent

mitchgrout:
I'll aim to review all `parts/` files tonight, but at-worst there'll probably only be a few small tweaks to make.

doctest/parts/private/assert/result.cpp
```diff
@@ -10,62 +10,61 @@ DOCTEST_SUPPRESS_PRIVATE_WARNINGS_PUSH
...
 namespace detail {
 
-    Result::Result(bool passed, const String& decomposition)
-            : m_passed(passed)
-            , m_decomp(decomposition) {}
+Result::Result(bool passed, const String &decomposition)
+        : m_passed(passed), m_decomp(decomposition) {}
+
+ResultBuilder::ResultBuilder(assertType::Enum at, const char *file, int line, const char *expr,
+                             const char *exception_type, const String &exception_string)
```
mitchgrout: Been tossing this one up a bit, but I personally prefer the one-per-line bin-packed style if the line overflows like this, e.g.

```cpp
ResultBuilder::ResultBuilder(assertType::Enum at,
                             const char      *file,
                             int              line,
                             ...)
```

doctest/parts/private/matchers/approx.cpp
```diff
@@ -5,9 +5,8 @@ DOCTEST_SUPPRESS_PRIVATE_WARNINGS_PUSH
 namespace doctest {
 
 Approx::Approx(double value)
-        : m_epsilon(static_cast<double>(std::numeric_limits<float>::epsilon()) * 100)
-        , m_scale(1.0)
-        , m_value(value) {}
+        : m_epsilon(static_cast<double>(std::numeric_limits<float>::epsilon()) * 100), m_scale(1.0),
+          m_value(value) {}
```
mitchgrout: Same as with constructor args, I think constructor-lists should probably just be one-per-line if they can't all fit on a single line

doctest/parts/private/reporters/console.h
```diff
@@ -16,477 +16,524 @@ DOCTEST_SUPPRESS_PRIVATE_WARNINGS_PUSH
...
+                                1)));
+        auto passwidth = int(
+                std::ceil(log10(static_cast<double>(std::max(
+                                        p.numTestCasesPassingFilters - p.numTestCasesFailed,
+                                        static_cast<unsigned>(p.numAsserts - p.numAssertsFailed))) +
+                                1)));
+        auto       failwidth      = int(std::ceil(
+                log10(static_cast<double>(std::max(p.numTestCasesFailed,
+                                                              static_cast<unsigned>(p.numAssertsFailed))) +
+                                 1)));
```
mitchgrout: These calls are a little screwy; it's half because they're all sequential so alignment rules are triggering, but also the way it's treating spilt args is not so nice in general

doctest/parts/public/assert/comparator.h
```diff
@@ -8,35 +8,44 @@ DOCTEST_SUPPRESS_PUBLIC_WARNINGS_PUSH
...
-    template<class T>               struct decay_array<T[]>  { using type = T*; };
-
-    template<class T>   struct not_char_pointer              { static DOCTEST_CONSTEXPR int value = 1; };
-    template<>          struct not_char_pointer<char*>       { static DOCTEST_CONSTEXPR int value = 0; };
-    template<>          struct not_char_pointer<const char*> { static DOCTEST_CONSTEXPR int value = 0; };
-
-    template<class T> struct can_use_op : public not_char_pointer<typename decay_array<T>::type> {};
+template <class T> struct decay_array {
+    using type = T;
+};
```
mitchgrout: Would be nice if we could somehow retain this format. No biggie if we can't though

doctest/parts/public/matchers/approx.h
```diff
@@ -7,84 +7,105 @@ DOCTEST_SUPPRESS_PUBLIC_WARNINGS_PUSH
...
     Approx operator()(double value) const;
 
 #ifdef DOCTEST_CONFIG_INCLUDE_TYPE_TRAITS
     template <typename T>
-    explicit Approx(const T& value,
-                    typename detail::types::enable_if<std::is_constructible<double, T>::value>::type* =
-                            static_cast<T*>(nullptr)) {
+    explicit Approx(const T &value,
+                    typename detail::types::enable_if<std::is_constructible<double, T>::value>::type
+                            * = static_cast<T *>(nullptr)) {
```
mitchgrout: This one's a bit rough, but I dunno if we can do much about it besides just rework how it's implemented

doctest/parts/public/debugger.h
```diff
@@ -4,24 +4,25 @@
...
 #endif
 #elif defined(DOCTEST_PLATFORM_MAC)
 #if defined(__x86_64) || defined(__x86_64__) || defined(__amd64__) || defined(__i386)
-#define DOCTEST_BREAK_INTO_DEBUGGER() __asm__("int $3\n" : :) // NOLINT(hicpp-no-assembler)
+#define DOCTEST_BREAK_INTO_DEBUGGER() __asm__("int $3\n" : :)
 #elif defined(__ppc__) || defined(__ppc64__)
 // https://www.cocoawithlove.com/2008/03/break-into-debugger.html
-#define DOCTEST_BREAK_INTO_DEBUGGER() __asm__("li r0, 20\nsc\nnop\nli r0, 37\nli r4, 2\nsc\nnop\n": : : "memory","r0","r3","r4") // NOLINT(hicpp-no-assembler)
+#define DOCTEST_BREAK_INTO_DEBUGGER()                                                              \
+    __asm__("li r0, 20\nsc\nnop\nli r0, 37\nli r4, 2\nsc\nnop\n" : : : "memory", "r0", "r3", "r4")
```
mitchgrout: Super specific so I don't mind if it gets skipped, but no whitespace between `:` in ASM

doctest/parts/public/macros.h
```diff
@@ -116,92 +116,88 @@ namespace detail {
...
                                            DOCTEST_ANONYMOUS(DOCTEST_ANON_TMP_))
 
 #define DOCTEST_TEST_CASE_TEMPLATE_INSTANTIATE_IMPL(id, anon, ...)                                 \
-    DOCTEST_GLOBAL_NO_WARNINGS(DOCTEST_CAT(anon, DUMMY), /* NOLINT(cert-err58-cpp, fuchsia-statically-constructed-objects) */ \
-        doctest::detail::instantiationHelper(                                                      \
-            DOCTEST_CAT(id, ITERATOR)<__VA_ARGS__>(__FILE__, __LINE__, 0)))
+    DOCTEST_GLOBAL_NO_WARNINGS(                                                                    \
+            DOCTEST_CAT(                                                                           \
+                    anon,                                                                          \
+                    DUMMY), /* NOLINT(cert-err58-cpp, fuchsia-statically-constructed-objects) */   \
```
mitchgrout: `NOLINT` moved and no longer suppresses the warning afaik

doctest/parts/public/macros.h
```diff
@@ -789,50 +1031,49 @@ namespace detail {
...
+#define DOCTEST_FAST_REQUIRE_GE DOCTEST_REQUIRE_GE
+#define DOCTEST_FAST_WARN_LE DOCTEST_WARN_LE
+#define DOCTEST_FAST_CHECK_LE DOCTEST_CHECK_LE
+#define DOCTEST_FAST_REQUIRE_LE DOCTEST_REQUIRE_LE
+
+#define DOCTEST_FAST_WARN_UNARY DOCTEST_WARN_UNARY
+#define DOCTEST_FAST_CHECK_UNARY DOCTEST_CHECK_UNARY
+#define DOCTEST_FAST_REQUIRE_UNARY DOCTEST_REQUIRE_UNARY
+#define DOCTEST_FAST_WARN_UNARY_FALSE DOCTEST_WARN_UNARY_FALSE
+#define DOCTEST_FAST_CHECK_UNARY_FALSE DOCTEST_CHECK_UNARY_FALSE
```
mitchgrout, line 846: Aligned `#define` values would be nice

doctest/parts/public/matchers/approx.h
Repiteo: This definitely looks more verbose than necessary, as we *should* be able to just use several of these `type_traits` outright from forward-declared/`using` logic setup beforehand. I'll give it a test all the same

Repiteo:
@mitchgrout I'm back to my job now, so I don't have the same time to adjust this as I thought. Feel free to further adjust this PR yourself, or use the updated `.clang-format` in whatever CI PR you were cooking up

Repiteo:
Note: I've since removed several of the pragma-like comments (eg: OCLINT) during this refactor, as their tools are no longer in use in this repo (some might be dead?). This simplified a lot of the more glaring issues, as the formatting no longer has to account for an immutable comment

mitchgrout:
> Note: I've since removed several of the pragma-like comments (eg: OCLINT) during this refactor, as their tools are no longer in use in this repo (some might be dead?). This simplified a lot of the more glaring issues, as the formatting no longer has to account for an immutable comment

Yeah, I think the OCLint stuff might've been from Viktor and not from a consuming project, so we're probably fine to drop it. I might pull that out to a separate commit just for history purposes.

mitchgrout:
I've been doing a little hacking locally, and thought it might be easier to note my suggestions via a patch on your current commit. A few bits were:

- Bumping line-length to 120
- Disabling bin-packing for basically everything
- Disabling one-line functions
- Disabling assignment/declaration alignments

I've also manually reflowed some functions (esp. regarding displaying output) and disabled clang-format to preserve the structure a bit more.

I did prep this patch prior to your latest change-set, so the `.clang-format` might be a little out-of-date.

One thing that might be worth considering (just noted after looking at #1039) is dropping `// clang-format ...` comments as part of `scripts/assemble.py`. We don't have to `clang-format` the assembled header, since it's an artifact rather than a source.

Another thing is, reckon it's worth running `clang-format` over `examples/` and `tests/`? IMO it's about 50/50, a lot of `CHECK` statements have been arranged to make reading everything a bit nicer, so `clang-format` just comes in and ruins it unfortunately. Mostly just prep for #1038 

---

[8bd303ae.patch](https://github.com/user-attachments/files/25466698/8bd303ae.patch)

Repiteo:
Applied the patch! Didn't have to do much manual adjusting to resync with my latest update, mostly doing a sanity pass of the `clang-format` wrapped sections. I still integrated the removal of OCLINT/etc, but I'll isolate that to a standalone PR shortly, under the assumption that this'll just be rebased onto that

doctest/parts/public/macros.h
```diff
@@ -32,249 +34,261 @@ namespace detail {
...
 
-// clang-format off
-// for logging - disabling formatting because it's important to have these on 2 separate lines - see PR #557
-#define DOCTEST_INFO(...)                                                                          \
-    DOCTEST_INFO_IMPL(DOCTEST_ANONYMOUS(DOCTEST_CAPTURE_),                                         \
-                      DOCTEST_ANONYMOUS(DOCTEST_CAPTURE_OTHER_),                                   \
-                      __VA_ARGS__)
-// clang-format on
+#define DOCTEST_INFO(...)                                                                                              \
+    DOCTEST_INFO_IMPL(DOCTEST_ANONYMOUS(DOCTEST_CAPTURE_), DOCTEST_ANONYMOUS(DOCTEST_CAPTURE_OTHER_), __VA_ARGS__)
```
Repiteo, line 263: #557 was solved by changing the second anonymous input to `DOCTEST_CAPTURE_OTHER_`, not by having the two values in separate lines. The comment was removed and this is now formatted as normal

Repiteo:
Ran `clang-format` against the remaining cpp files to get a feel for how much of an impact it'd have. Most changes were pretty minor all things considered, but there's probably a few sections that you'd want to keep in `clang-format off/on` wrappers. I have zero clue why GitHub things there's conflicts that need to be resolved…

mitchgrout:
> Ran `clang-format` against the remaining cpp files to get a feel for how much of an impact it'd have. Most changes were pretty minor all things considered, but there's probably a few sections that you'd want to keep in `clang-format off/on` wrappers. I have zero clue why GitHub things there's conflicts that need to be resolved…

I think we skip formatting of `examples/` and `tests/` for now - besides that, I'm overall happy with the structure that we have now. I'll merge after we revert those dirs.

Thanks for the work on this, I know it's been a quite a while!

mitchgrout (approved):

PR #1023, opened 2026-01-16
Bump minimum cmake version to 3.14

G4m4:
Update of [!953](https://github.com/doctest/doctest/pull/953) initially proposed by @nim65s.

CMake used to emit a warning for setting the minimum to 3.5 or earlier, but starting with Cmake 4.0 it is [an error](https://cmake.org/cmake/help/latest/release/4.0.html#deprecated-and-removed-features).

This is especially painful on MacOS where brew relies on recent versions of CMake ([4.2.1 at the time of writing this](https://formulae.brew.sh/formula/cmake)).

I have bumped cmake_minimum_required(VERSION ***) to 3.14 everywhere I could find.
It also allows to simplify various workarounds in the build files, all based on CMake versions checks, allowing for a more standard and easier to maintain build chain.

I have then applied clang-format to all touched files but this makes for a surprising amount of changes, so maybe we can drop this specific formatting commit? What do you think?

Nothing else, not even cosmetic changes (but the aforementioned clang-format pass), were made.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,26 +1,31 @@
 cmake_minimum_required(VERSION 3.14)
 
 if(POLICY CMP0077)
-    cmake_policy(SET CMP0077 NEW)
+  cmake_policy(SET CMP0077 NEW)
 endif()
 
-################################################################################
-## DOCTEST
-################################################################################
+# ##############################################################################
+# DOCTEST
+# ##############################################################################
 
 file(READ ${CMAKE_CURRENT_SOURCE_DIR}/scripts/version.txt ver)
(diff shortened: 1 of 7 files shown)


mitchgrout:
Will hold off for now as the author for #953 is active again. When/if that change gets merged, we can just keep this one as a general modernization of the project. The `cmake-format` stuff should probably be a separate PR as it makes it a bit hard to see what your intended changes actually are, plus we should arguably check it on CI/CD as well if we're planning on adhering to it.

G4m4:
No worries, I'm happy that #953 wen through and I agree with the format pass that should be separate, so will close this PR!

PR #1038, opened 2026-02-15
ci: Add job to check clang-format

mitchgrout:
## Description

Adds `.github/workflows/clang-format.yml`, which provides a `clang-format` job to check that our source-files adhere to the project `.clang-format` file. If failures occur, a `clang-format.patch` file will be uploaded, which can be applied via `git apply clang-format.patch` to apply formatting fixes and fix the job.

~~Currently this will break as our code does not adhere to the `.clang-format` file. This will be fixed in #1015~~

## GitHub Issues

Relates to #1015

### DIFF
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -1,0 +1,79 @@
+name: Clang-Format
+
+on:
+  push:
+    branches: [dev, master]
+    paths-ignore:
+      - "doc/**"
+      - "scripts/**"
+      - "LICENSE.txt"
+      - "README.md"
+
+  pull_request:
+    paths-ignore:
+      - "doc/**"
+      - "scripts/**"
+      - "LICENSE.txt"
+      - "README.md"
+
+concurrency:
+  group: ${{ github.workflow }}|${{ github.ref_name }}|clang-format
+  cancel-in-progress: true
+
+jobs:
+  clang-format:
+    runs-on: ubuntu-24.04
+    env:
+      LLVM_VERSION: 22
+      DIRECTORIES: >-
+        doctest/parts
+
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Configure LLVM Upstream
+        run: |
+          # See apt.llvm.org
+          wget -qO- http://apt.llvm.org/llvm-snapshot.gpg.key \
+            | sudo tee /etc/apt/trusted.gpg.d/apt.llvm.org.asc
+
+          sudo add-apt-repository \
+            "deb http://apt.llvm.org/$(lsb_release -cs)/ llvm-toolchain-$(lsb_release -cs)-${{ env.LLVM_VERSION}} main"
+
+      - name: Install
+        run: |
+          sudo apt-get update
+          sudo apt-get install clang-format-${{ env.LLVM_VERSION }}
+
+      - name: Apply
+        run: |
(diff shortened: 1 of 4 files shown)


mitchgrout:
@Repiteo FYI, just setting this up as a companion to #1015, so we can ensure conformance to the formatting going forward

.github/workflows/clang-format.yml
```diff
@@ -0,0 +1,67 @@
...
+        run: |
+          find ${{ env.DIRECTORIES }} \
+            -regextype posix-extended \
+            -regex '^.*\.(cpp|hpp|cxx|hxx|cc|c|h)$' \
+            -exec clang-format-${{ env.LLVM_VERSION }} --style=file -i {} + \
+
+      - name: Diff
+        id: diff
+        run: |
+          git diff | tee clang-format.patch
```
mitchgrout, line 58: Not sure if there's a better way to achieve this, afaik `clang-format` doesn't supply a `--diff` flag or anything.

.github/workflows/clang-format.yml
```diff
@@ -0,0 +1,67 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}|${{ github.ref_name }}|clang-tidy
+  cancel-in-progress: true
+
+jobs:
+  clang-format:
+    runs-on: ubuntu-24.04
+    env:
+      LLVM_VERSION: 19
```
mitchgrout: Most recent clang-format available on Ubuntu-24.04. I was tossing up setting up the LLVM apt source, but I don't think the difference between clang-format 19 and 21 is _that_ significant. Lemme know if you think it's worth it though

.github/workflows/clang-format.yml
```diff
@@ -0,0 +1,67 @@
...
+            exit 1
+          fi
+        continue-on-error: true
+
+      - name: Upload Patch
+        if: steps.diff.outcome == 'failure'
+        uses: actions/upload-artifact@v4
+        with:
+          name: clang-format.patch
+          path: clang-format.patch
```
mitchgrout, line 70: Will need to rework this a smidge. In short, what happens at the moment is:

1. `Apply` passes, because `clang-format` did its job (re-formatted the code)
2. `Diff` passes, because while it internally exited with a bad code, it was marked as `continue-on-error` so it didn't mark the overall job as failed
3. `Upload Patch` passes, because it uploaded a file
4. The overall job passes because the last job passed(?)

.github/workflows/clang-format.yml
Repiteo: I see no harm in giving it a shot! If it finds a discrepancy, we can act upon it then and there

mitchgrout: Hrm, I do suspect that we'll need to use `clang-format-21` or above, since the job is reporting a real minor diff - [only 3 lines affected](https://github.com/doctest/doctest/actions/runs/22466021118/job/65072063607?pr=1038).

This kinda applies to our `ci.yml` file as well, but currently we're relying on grabbing LLVM toolchain bits from the stock Ubuntu apt sources. We could swap to using <https://apt.llvm.org/llvm.sh> to grab the versions we need instead (or butcher the relevant parts to just add the upstreams we need). Also just noting this since `clang-22` was released only 3 days ago, so it's not in the Ubuntu repos quite yet.

Thoughts?

PR #1039, opened 2026-02-22
chore: Correct public includes, clarify intent

Repiteo:
## Description
A handful of headers were out-of-order, or not properly accounting for the fact that the config might've not been passed. This aims to make that a bit clearer with `IWYU` pragmas[^1]; though not currently enforced, it helps demonstrate intent on what should be included as a baseline. This will (hopefully) make `public/config.h` the "baseline" include for most files, preventing footguns when we invariably start making the private files use explicit includes. Finally, adds `-Wunused-macros` to the common clang suppressions, as it's absurdly intrusive when trying to parse headers to have this false positive blaring the whole time

[^1]: https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md


doctest/parts/public/config.h
```diff
@@ -5,7 +5,12 @@
 #ifndef DOCTEST_PARTS_PUBLIC_CONFIG
 #define DOCTEST_PARTS_PUBLIC_CONFIG
 
+// IWYU pragma: begin_exports
+#include "doctest/parts/public/version.h"
+#include "doctest/parts/public/platform.h"
 #include "doctest/parts/public/compiler.h"
+#include "doctest/parts/public/warnings.h"
+// IWYU pragma: end_exports
```
mitchgrout, line 13: So just for my understanding, the gist is that we want to encourage units to `#include ".../config.h"` in lieu of the more util-focused headers, correct? And to do that, we're adding:

1. `IWYU pragma: private, include "public/config.h"` to say "these symbols are re-exported by config.h"
2. `IWYU pragma: export` to similarly do ^this

Correct?

scripts/assemble.py
```diff
@@ -35,12 +35,8 @@
   #define DOCTEST_LIBRARY_IMPLEMENTATION
   DOCTEST_CLANG_SUPPRESS_WARNING_POP
 
-  DOCTEST_SUPPRESS_PRIVATE_WARNINGS_PUSH
-
   $sources
 
-  DOCTEST_SUPPRESS_PRIVATE_WARNINGS_POP
-
```
mitchgrout, line 43: I'm surprised that we can just drop this without the compiler exploding rather violently. Our local (non-subproject) build runs with quite a few warnings inhibited (see `scripts/cmake/common.cmake`), so I don't know if this is a valid removal without further testing.

doctest/parts/public/config.h
Repiteo: That is correct!

scripts/assemble.py
Repiteo: It *should* be fine, as the individual private files have all been wrapped with their own `DOCTEST_SUPPRESS_PRIVATE_WARNINGS_*`. If there's a discrepancy to speak of, I wouldn't know how it could occur

mitchgrout: Oh duh, I forgot we pushed the warning-suppression inward. Yeah, that should be fine in that case.

PR #1040, opened 2026-02-22
chore: Strip unused annotations for legacy tools

Repiteo:
- Extracted from #1015

## Description
There exist several vestigial annotations for legacy tools within the codebase. This serves to remove them atomically, so it's not lost in the refactor process


mitchgrout:
LGTM

PR #1044, opened 2026-03-01
add GENERATE macro

aozgaa:
## Description

This adds first-class value-parameterized test execution through a new `GENERATE(...)` macro.

### User-facing behavior:
- Add `DOCTEST_GENERATE` / `GENERATE` in `public/macros.h`.
- `GENERATE(...)` produces one test run per value.
- Multiple `GENERATE` calls form a cartesian product.
- Generator traversal composes with `SUBCASE`, with generator choices outer to subcase traversal.
- `REQUIRE`/failures still advance to the next generator combination instead of stopping the whole parameter set.

### Implementation approach:
- Introduce a shared traversal model in `doctest/parts/private/traversal.{h,cpp}`.
- Replace the older subcase-specific backtracking state in `ContextState` with a single `TraversalState` that tracks:
  - the current decision path
  - the branches discovered on a given rerun
  - active subcase depth for filtering/reporting
- Keep `SUBCASE` and `GENERATE` on the same rerun/backtracking mechanism, instead of layering a second traversal system on top.
- Make generator evaluation select a branch index from the current traversal depth, then advance the decision depth just like subcase traversal does.
- Reset traversal state per test case and per rerun, while preserving the selected decision prefix across reruns.
- Reuse the same traversal state for fatal unwinding so reporters still receive balanced `subcase_end` events when a test aborts inside active subcases.

### Documentation:
- Replace the old "future support / workaround" section for value-parameterized tests with actual `GENERATE` documentation.
- Update the feature list and FAQ to reflect that value-parameterized tests are now supported.

### Test coverage:
- Add unit tests for generate scenarios:
  - basic usage
  - cartesian products
  - dependent generators
  - capture nontrivial types
  - interaction with `SUBCASE`
  - interaction with `REQUIRE`
  - control flow interaction
- tests/control-flow/subcases ran checked by golden outputs
- Add dedicated filter tests for generator-aware test-case and subcase filtering.
- importantly, no changes in `SUBCASE` unit tests/golden baselines

Notes:
- The core design goal here was to make generators a natural extension of doctest's existing rerun model, rather than a separate data-driven execution layer.
- This keeps ordering, filtering, reporting, and exception behavior aligned with existing `SUBCASE` semantics.

## GitHub Issues

Related:
- #321 My own implementation of value parameterization (with multiple parameters)
- #563 Another (pure macro) approach to value parameterization
- #887 Idea for parameterized test macro

Possibly partially addresses:
- #930 Add support for templated test cases - parameterized by value

Issue #930, opened 2025-08-01
Add support for templated test cases - parameterized by value


definability:
## Introduction

There is also support for [Templated test cases - parameterized by type](https://github.com/doctest/doctest/blob/1da23a3e8119ec5cce4f9388e91b065e20bf06f5/doc/markdown/parameterized-tests.md#templated-test-cases---parameterized-by-type) and several recommendations regarding [Value-parametrized test cases](https://github.com/doctest/doctest/blob/1da23a3e8119ec5cce4f9388e91b065e20bf06f5/doc/markdown/parameterized-tests.md#value-parameterized-test-cases).

I believe an intermediate feature can be implemented based on the existing code &mdash; tests parameterisation using compile-time values. Like type-parameterised tests, the parameters are known at compile time, and like value-parameterised tests, users have more ways to shorten the test cases.

## Proposed syntax

1. Directly pass the list of types to the templated test case:
    ```c++
    TEST_CASE_CVALUE(name, V, 1, 1U, 1L, 1LL, 1Z) {
        CHECK_EQ(V, 1);
    }
    ```
2. Define the templated test case with a unique name (identifier) for later instantiation:
    ```c++
    TEST_CASE_CVALUE_DEFINE(name, V, test_id) {
        CHECK_EQ(V, 1);
    }
    TEST_CASE_CVALUE_INVOKE(test_id, true, 1UL, 1.0);
    TEST_CASE_CVALUE_APPLY(test_id, std::tuple<true, 1UL, 1.0>);
    ```

#### Boolean

Functions may depend on multiple boolean flags.
There are scenarios when checking the program's correctness with any flags is required.

The current workaround: use [std::bool_constant](https://en.cppreference.com/w/cpp/types/integral_constant.html#Helper_alias_templates) instantiations `std::true_type` and `std::false_type`.

#### Enumerations

Instead of boolean flags, functions may depend on enumeration types.

The current workaround: [std::integral_constant](https://en.cppreference.com/w/cpp/types/integral_constant.html).

#### Types allowing absence of value

Pointers and [std::optional](https://en.cppreference.com/w/cpp/utility/optional.html) allow using special values meaning the absenece of value ([std::nullopt](https://en.cppreference.com/w/cpp/utility/optional/nullopt.html) and [nullptr](https://en.cppreference.com/w/cpp/language/nullptr.html) correspondingly).

[...]

Issue #887, opened 2025-01-08
Idea for parameterized test macro


jrosiek:
Hi. Just a quick idea for a macro with nicer syntax and better output (IMO) than what is in the docs. Requires more modern C++ compiler due to syntax used.

```
#define PARAMETERIZED_SUBCASE(name, collection, variable_name)                                                         \
    DOCTEST_SUBCASE(name)                                                                                              \
      for (auto [i, it] = std::make_tuple(0, (collection).begin()); it != (collection).end(); ++it, ++i)                 \
        if (auto variable_name = *it; true)                                                                            \
          DOCTEST_SUBCASE(std::string("[" + std::to_string(i) + "] " + ::doctest::toString(variable_name).c_str()).c_str())
```

Usage:
```
auto all_values = std::vector<int>{1,2,3,4,5,6};

PARAMETERIZED_SUBCASE("my test", all_values, value)
{
  CHECK(value > 3);
}
```

cdeln:
Hi, this ideas have been circulating for quite some while. I'll try to organize them under the new label category/value-parameterization .

Issue #563, opened 2021-12-09
Another (pure macro) approach to value parameterization


Saalvage:
## Description
I have written a small macro-only tool to help me parameterize test cases.
Here it is in its entirity:
```cpp
#define CONCAT(a, b) CONCAT_INNER(a, b)
#define CONCAT_INNER(a, b) a ## b

#define PP_NARG(...) PP_NARG_(__VA_ARGS__,PP_RSEQ_N())
#define PP_NARG_(...) PP_ARG_N(__VA_ARGS__)
#define PP_ARG_N(_1, _2, _3, _4, _5, _6, _7, _8, _9, N, ...) N
#define PP_RSEQ_N() 9, 8, 7, 6, 5, 4, 3, 2, 1

#define PARENS ()

#define EXPAND(...) EXPAND4(EXPAND4(EXPAND4(EXPAND4(__VA_ARGS__))))
#define EXPAND4(...) EXPAND3(EXPAND3(EXPAND3(EXPAND3(__VA_ARGS__))))
#define EXPAND3(...) EXPAND2(EXPAND2(EXPAND2(EXPAND2(__VA_ARGS__))))
#define EXPAND2(...) EXPAND1(EXPAND1(EXPAND1(EXPAND1(__VA_ARGS__))))
#define EXPAND1(...) __VA_ARGS__

#define DEBRACE(...) __VA_ARGS__

#define FWD(func, ...) func(__VA_ARGS__)

#define FOR_EACH(func, constVar, ...) __VA_OPT__(EXPAND(FOR_EACH_HELPER(func, constVar, __VA_ARGS__)))

#define FOR_EACH_HELPER(func, constVar, sub, ...) FWD(func, DEBRACE constVar, sub) \
                                                  __VA_OPT__(FOR_EACH_AGAIN PARENS (func, constVar, __VA_ARGS__))
#define FOR_EACH_AGAIN() FOR_EACH_HELPER

#define SUBCASE_VAR_GENERATE(nested, var, ...) FOR_EACH(SUBCASE_GENERATE, (nested, var), __VA_ARGS__)
#define SUBCASE_UNPACK(nested, var) SUBCASE_VAR_GENERATE(nested, var)

#define SUBCASE_SPECIAL1(a) SUBCASE_UNPACK(, DEBRACE a)
#define SUBCASE_SPECIAL2(a, ...) SUBCASE_UNPACK(SUBCASE_SPECIAL1(__VA_ARGS__), DEBRACE a)
[...]

PR #1047, opened 2026-03-02
feat: add GENERATE macro

aozgaa:
(Note, re-created https://github.com/doctest/doctest/pull/1044 after removing some scope-creep/churn about signal handling+signal testing)

## Description

This adds first-class value-parameterized test execution through a new `GENERATE(...)` macro.

### User-facing behavior:
- Add `DOCTEST_GENERATE` / `GENERATE` in `public/macros.h`.
- `GENERATE(...)` produces one test run per value.
- Multiple `GENERATE` calls form a cartesian product.
- Generator traversal composes with `SUBCASE`, with generator choices outer to subcase traversal.
- `REQUIRE`/failures still advance to the next generator combination instead of stopping the whole parameter set.

### Implementation approach:
- Introduce a shared traversal model in `doctest/parts/private/traversal.{h,cpp}`.
- Replace the older subcase-specific backtracking state in `ContextState` with a single `TraversalState` that tracks:
  - the current decision path
  - the branches discovered on a given rerun
  - active subcase depth for filtering/reporting
- Keep `SUBCASE` and `GENERATE` on the same rerun/backtracking mechanism, instead of layering a second traversal system on top.
- Make generator evaluation select a branch index from the current traversal depth, then advance the decision depth just like subcase traversal does.
- Reset traversal state per test case and per rerun, while preserving the selected decision prefix across reruns.
- Reuse the same traversal state for fatal unwinding so reporters still receive balanced `subcase_end` events when a test aborts inside active subcases.

### Documentation:
- Replace the old "future support / workaround" section for value-parameterized tests with actual `GENERATE` documentation.
- Update the feature list and FAQ to reflect that value-parameterized tests are now supported.

### Test coverage:
- Add unit tests for generate scenarios:
  - basic usage
  - cartesian products
  - dependent generators
  - capture nontrivial types
  - interaction with `SUBCASE`
  - interaction with `REQUIRE`
  - control flow interaction
- tests/control-flow/subcases ran checked by golden outputs
- Add dedicated filter tests for generator-aware test-case and subcase filtering.
- importantly, no changes in `SUBCASE` unit tests/golden baselines

Notes:
- The core design goal here was to make generators a natural extension of doctest's existing rerun model, rather than a separate data-driven execution layer.
- This keeps ordering, filtering, reporting, and exception behavior aligned with existing `SUBCASE` semantics.

## GitHub Issues

Related:
- #321 My own implementation of value parameterization (with multiple parameters)
- #563 Another (pure macro) approach to value parameterization
- #887 Idea for parameterized test macro

Possibly partially addresses:
- #930 Add support for templated test cases - parameterized by value

### DIFF
--- a/examples/all_features/generators.cpp
+++ b/examples/all_features/generators.cpp
@@ -302,7 +302,13 @@ TEST_CASE("GENERATE with fixed for loop") {
         for (int i = 0; i < 3; ++i) {
             SUBCASE("fixed name") {
                 js.push_back(3 + 2 * i + GENERATE(0, 1));
-                MESSAGE("i=", i, " js=", doctest::toString(js));
+                std::ostringstream oss;
+                oss << "i=" << i << " js=[ ";
+                for (auto j: js) {
+                    oss << j << " ";
+                }
+                oss << "]";
+                MESSAGE(oss.str());
             }
         }
     }
--- a/examples/all_features/test_output/generators.cpp.txt
+++ b/examples/all_features/test_output/generators.cpp.txt
@@ -1102,7 +1102,7 @@ TEST CASE:  GENERATE with fixed for loop
   loop with fixed name
   fixed name
 
-generators.cpp(0): MESSAGE: i=0 js=[3]
+generators.cpp(0): MESSAGE: i=0 js=[ 3 ]
 
 ===============================================================================
 generators.cpp(0):
@@ -1115,42 +1115,7 @@ DEEPEST SUBCASE STACK REACHED (DIFFERENT FROM THE CURRENT ONE):
   fixed name
   fixed name
 
-generators.cpp(0): MESSAGE: i=1 js=[3, 5]
-
-===============================================================================
-generators.cpp(0):
-TEST CASE:  GENERATE with fixed for loop
-  loop with fixed name
-  fixed name
-
-DEEPEST SUBCASE STACK REACHED (DIFFERENT FROM THE CURRENT ONE):
-  loop with fixed name
-  fixed name
-  fixed name
-  fixed name
-
-generators.cpp(0): MESSAGE: i=2 js=[3, 5, 7]
-
-===============================================================================
-generators.cpp(0):
-TEST CASE:  GENERATE with fixed for loop
-  loop with fixed name
-  fixed name
-
-generators.cpp(0): MESSAGE: i=0 js=[4]
-
-===============================================================================
-generators.cpp(0):
-TEST CASE:  GENERATE with fixed for loop
-  loop with fixed name
-  fixed name
-
-DEEPEST SUBCASE STACK REACHED (DIFFERENT FROM THE CURRENT ONE):
-  loop with fixed name
-  fixed name
-  fixed name
-
-generators.cpp(0): MESSAGE: i=1 js=[4, 6]
+generators.cpp(0): MESSAGE: i=1 js=[ 3 5 ]
 
 ===============================================================================
 generators.cpp(0):
@@ -1164,7 +1129,42 @@ DEEPEST SUBCASE STACK REACHED (DIFFERENT FROM THE CURRENT ONE):
   fixed name
   fixed name
 
-generators.cpp(0): MESSAGE: i=2 js=[4, 6, 8]
+generators.cpp(0): MESSAGE: i=2 js=[ 3 5 7 ]
+
+===============================================================================
+generators.cpp(0):
+TEST CASE:  GENERATE with fixed for loop
+  loop with fixed name
+  fixed name
+
+generators.cpp(0): MESSAGE: i=0 js=[ 4 ]
+
+===============================================================================
+generators.cpp(0):
+TEST CASE:  GENERATE with fixed for loop
+  loop with fixed name
+  fixed name
+
+DEEPEST SUBCASE STACK REACHED (DIFFERENT FROM THE CURRENT ONE):
+  loop with fixed name
+  fixed name
+  fixed name
+
+generators.cpp(0): MESSAGE: i=1 js=[ 4 6 ]
+
+===============================================================================
+generators.cpp(0):
+TEST CASE:  GENERATE with fixed for loop
+  loop with fixed name
+  fixed name
+
+DEEPEST SUBCASE STACK REACHED (DIFFERENT FROM THE CURRENT ONE):
+  loop with fixed name
+  fixed name
+  fixed name
+  fixed name
+
+generators.cpp(0): MESSAGE: i=2 js=[ 4 6 8 ]
 
 ===============================================================================
 generators.cpp(0):
--- a/examples/all_features/test_output/generators.cpp_xml.txt
+++ b/examples/all_features/test_output/generators.cpp_xml.txt
@@ -1183,21 +1183,21 @@
         <SubCase name="fixed name" filename="generators.cpp" line="0">
           <Message type="WARNING" filename="generators.cpp" line="0">
             <Text>
-              i=0 js=[3]
+              i=0 js=[ 3 ]
             </Text>
           </Message>
         </SubCase>
         <SubCase name="fixed name" filename="generators.cpp" line="0">
           <Message type="WARNING" filename="generators.cpp" line="0">
             <Text>
-              i=1 js=[3, 5]
+              i=1 js=[ 3 5 ]
             </Text>
           </Message>
         </SubCase>
         <SubCase name="fixed name" filename="generators.cpp" line="0">
           <Message type="WARNING" filename="generators.cpp" line="0">
             <Text>
-              i=2 js=[3, 5, 7]
+              i=2 js=[ 3 5 7 ]
             </Text>
           </Message>
         </SubCase>
@@ -1206,21 +1206,21 @@
         <SubCase name="fixed name" filename="generators.cpp" line="0">
           <Message type="WARNING" filename="generators.cpp" line="0">
             <Text>
-              i=0 js=[4]
+              i=0 js=[ 4 ]
             </Text>
           </Message>
         </SubCase>
         <SubCase name="fixed name" filename="generators.cpp" line="0">
           <Message type="WARNING" filename="generators.cpp" line="0">
             <Text>
-              i=1 js=[4, 6]
+              i=1 js=[ 4 6 ]
             </Text>
           </Message>
         </SubCase>
         <SubCase name="fixed name" filename="generators.cpp" line="0">
           <Message type="WARNING" filename="generators.cpp" line="0">
             <Text>
-              i=2 js=[4, 6, 8]
+              i=2 js=[ 4 6 8 ]
             </Text>
           </Message>
         </SubCase>
--- a/examples/all_features/test_output/no_multi_lane_atomics.txt
+++ b/examples/all_features/test_output/no_multi_lane_atomics.txt
@@ -1114,7 +1114,7 @@ TEST CASE:  GENERATE with fixed for loop
   loop with fixed name
   fixed name
 
-generators.cpp(0): MESSAGE: i=0 js=[3]
+generators.cpp(0): MESSAGE: i=0 js=[ 3 ]
 
 ===============================================================================
 generators.cpp(0):
@@ -1127,42 +1127,7 @@ DEEPEST SUBCASE STACK REACHED (DIFFERENT FROM THE CURRENT ONE):
   fixed name
   fixed name
 
-generators.cpp(0): MESSAGE: i=1 js=[3, 5]
-
-===============================================================================
-generators.cpp(0):
-TEST CASE:  GENERATE with fixed for loop
-  loop with fixed name
-  fixed name
-
-DEEPEST SUBCASE STACK REACHED (DIFFERENT FROM THE CURRENT ONE):
-  loop with fixed name
-  fixed name
-  fixed name
-  fixed name
-
-generators.cpp(0): MESSAGE: i=2 js=[3, 5, 7]
-
-===============================================================================
-generators.cpp(0):
-TEST CASE:  GENERATE with fixed for loop
-  loop with fixed name
-  fixed name
-
-generators.cpp(0): MESSAGE: i=0 js=[4]
-
-===============================================================================
-generators.cpp(0):
-TEST CASE:  GENERATE with fixed for loop
-  loop with fixed name
-  fixed name
-
-DEEPEST SUBCASE STACK REACHED (DIFFERENT FROM THE CURRENT ONE):
-  loop with fixed name
-  fixed name
-  fixed name
-
-generators.cpp(0): MESSAGE: i=1 js=[4, 6]
+generators.cpp(0): MESSAGE: i=1 js=[ 3 5 ]
 
 ===============================================================================
 generators.cpp(0):
@@ -1176,7 +1141,42 @@ DEEPEST SUBCASE STACK REACHED (DIFFERENT FROM THE CURRENT ONE):
   fixed name
   fixed name
 
-generators.cpp(0): MESSAGE: i=2 js=[4, 6, 8]
+generators.cpp(0): MESSAGE: i=2 js=[ 3 5 7 ]
+
+===============================================================================
+generators.cpp(0):
+TEST CASE:  GENERATE with fixed for loop
+  loop with fixed name
+  fixed name
+
+generators.cpp(0): MESSAGE: i=0 js=[ 4 ]
+
+===============================================================================
+generators.cpp(0):
+TEST CASE:  GENERATE with fixed for loop
+  loop with fixed name
+  fixed name
+
+DEEPEST SUBCASE STACK REACHED (DIFFERENT FROM THE CURRENT ONE):
+  loop with fixed name
+  fixed name
+  fixed name
+
+generators.cpp(0): MESSAGE: i=1 js=[ 4 6 ]
+
+===============================================================================
+generators.cpp(0):
+TEST CASE:  GENERATE with fixed for loop
+  loop with fixed name
+  fixed name
+
+DEEPEST SUBCASE STACK REACHED (DIFFERENT FROM THE CURRENT ONE):
+  loop with fixed name
+  fixed name
+  fixed name
+  fixed name
+
+generators.cpp(0): MESSAGE: i=2 js=[ 4 6 8 ]
 
 ===============================================================================
 generators.cpp(0):
--- a/examples/all_features/test_output/no_multithreading.txt
+++ b/examples/all_features/test_output/no_multithreading.txt
@@ -1114,7 +1114,7 @@ TEST CASE:  GENERATE with fixed for loop
   loop with fixed name
   fixed name
 
-generators.cpp(0): MESSAGE: i=0 js=[3]
+generators.cpp(0): MESSAGE: i=0 js=[ 3 ]
 
 ===============================================================================
 generators.cpp(0):
@@ -1127,42 +1127,7 @@ DEEPEST SUBCASE STACK REACHED (DIFFERENT FROM THE CURRENT ONE):
   fixed name
   fixed name
 
-generators.cpp(0): MESSAGE: i=1 js=[3, 5]
-
-===============================================================================
-generators.cpp(0):
-TEST CASE:  GENERATE with fixed for loop
-  loop with fixed name
-  fixed name
-
-DEEPEST SUBCASE STACK REACHED (DIFFERENT FROM THE CURRENT ONE):
-  loop with fixed name
-  fixed name
-  fixed name
-  fixed name
-
-generators.cpp(0): MESSAGE: i=2 js=[3, 5, 7]
-
-===============================================================================
-generators.cpp(0):
-TEST CASE:  GENERATE with fixed for loop
-  loop with fixed name
-  fixed name
-
-generators.cpp(0): MESSAGE: i=0 js=[4]
-
-===============================================================================
-generators.cpp(0):
-TEST CASE:  GENERATE with fixed for loop
-  loop with fixed name
-  fixed name
-
-DEEPEST SUBCASE STACK REACHED (DIFFERENT FROM THE CURRENT ONE):
-  loop with fixed name
-  fixed name
-  fixed name
-
-generators.cpp(0): MESSAGE: i=1 js=[4, 6]
+generators.cpp(0): MESSAGE: i=1 js=[ 3 5 ]
 
 ===============================================================================
 generators.cpp(0):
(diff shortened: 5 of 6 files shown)


aozgaa:
Code coverage gaps are due to limitations of testing signal handlers. Doing this in a reasonable way is a significant change (or at least I struggled to keep to contained), so would prefer to address in a separate PR.

doctest/parts/public/generator.h
```diff
@@ -0,0 +1,28 @@
...
+
+#ifndef DOCTEST_CONFIG_DISABLE
+namespace doctest {
+namespace detail {
+
+DOCTEST_INTERFACE size_t acquireGeneratorDecisionIndex(size_t count);
+
+template <typename T, typename... Rest>
+T acquireGeneratorValue(T first, Rest... rest) {
+    const T values[] = {first, static_cast<T>(rest)...};
```
mitchgrout, line 17: AFAIK [fold-expressions](https://en.cppreference.com/w/cpp/language/fold.html) are C++17 onwards, but we're aiming for C++11 conformance.

I think it's still valuable to support C++11 where possible, so we could either gate `GENERATE` behind >=C++17, or rework the fold into an explicit recursive template function.

Another option would be to gate it behind a `DOCTEST_CONFIG_ENABLE_GENERATORS` or something, so people can opt-in. It could be a good way to introduce more experimental content without breaking user code too much.

doctest/parts/public/macros.h
```diff
@@ -190,6 +190,9 @@ int instantiationHelper(const T &) {
     if (const doctest::detail::Subcase &DOCTEST_ANONYMOUS(DOCTEST_ANON_SUBCASE_) DOCTEST_UNUSED =                      \
             doctest::detail::Subcase(name, __FILE__, __LINE__))
 
+// for generating value-parameterized test inputs
+#define DOCTEST_GENERATE(...) doctest::detail::acquireGeneratorValue(__VA_ARGS__)
```
mitchgrout, line 194: Nifty implementation! It is a little bit copy-heavy; in this example:

```cpp
TEST_CASE("") {
  auto x = GENERATE(1, ..., N);
}
```

...we will incur at-least `N+1` copies per-element (`N(N+1)` total), since we have a copy per-entry of the test-case, and a copy on-return from `acquireGeneratorValue`. I don't think there's any easy way around this though, but it would be worthwhile doing a quick benchmark to see how it compares to the previously-recommended approach / a bare-loop.

Not a blocker though IMO

doctest/parts/public/macros.h
```diff
@@ -532,6 +535,10 @@ int instantiationHelper(const T &) {
 // for subcases
 #define DOCTEST_SUBCASE(name)
 
+// for generating value-parameterized test inputs
+#define DOCTEST_GENERATE_IMPL(first, ...) (first)
```
mitchgrout, line 539: Unused?

examples/all_features/generators.cpp
```diff
@@ -0,0 +1,422 @@
...
+using namespace std;
+DOCTEST_MAKE_STD_HEADERS_CLEAN_FROM_WARNINGS_ON_WALL_END
+
+// ---------------------------------------------------------------------------
+// Basic: single generator
+// ---------------------------------------------------------------------------
+
+TEST_CASE("GENERATE - single, two values") {
+    auto x = GENERATE(10, 20);
+    cout << "x=" << x << endl;
```
mitchgrout: Would be nice to use `MESSAGE("x = ", x)` or similar to ensure each reporter plays nicely with the output

examples/all_features/generators.cpp
```diff
@@ -0,0 +1,422 @@
...
+// ---------------------------------------------------------------------------
+// Cartesian product: multiple generators, last cycles fastest
+// ---------------------------------------------------------------------------
+
+TEST_CASE("GENERATE - two generators, cartesian product") {
+    auto x = GENERATE(1, 2);   // outer: cycles slowest
+    auto y = GENERATE(10, 20); // inner: cycles fastest
+    cout << "x=" << x << " y=" << y << endl;
+    CHECK(x >= 1);
+    CHECK(y >= 10);
```
mitchgrout: A slightly more verbose, but more explicit test could be...

```cpp

static std::size_t iteration = 0;
static std::deque<std::tuple<int, int>> expectations = {
  { 1, 10 },
  { 1, 20 },
  { 2, 10 },
  { 2, 20 },
};

auto expected = expectations.at(iteration++);
CHECK( std::make_tuple(x, y) == expected );
```

It avoids depending too heavily on the `cout` calls to ensure order

examples/all_features/generators.cpp
```diff
@@ -0,0 +1,422 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// GENERATE with control flow
+// ---------------------------------------------------------------------------
+
+DOCTEST_MSVC_SUPPRESS_WARNING_PUSH  // context for spectre warning suppression
+DOCTEST_MSVC_SUPPRESS_WARNING(5045) // Spectre mitigation warning can surface at GENERATE call sites
+
+    TEST_CASE("GENERATE with fixed for loop") {
```
mitchgrout: Iffy indent here

mitchgrout (commented):
Thanks for the contribution! Big fan of how ergonomic the `GENERATE` macro is. A few overall comments:

* It may be good to split out the `traversal.cpp` changes into a separate PR; I suspect this will fix a few existing issues, so having a separate commit for it would be nice for traceability / bisecting errors when/if they occur. We can also add some [regression tests](https://github.com/doctest/doctest/tree/dev/tests/regression) to help keep them from occurring again
* The current `GENERATOR` impl. is good for fixed values, but I anticipate situations where we might want to use a container-type, e.g.
  ```cpp
  auto data = std::vector(64);
  std::iota(data.begin(), std::end(), 0); // data = { 0, 1, ..., 63 };
  auto item = GENERATE(data); // Could we possibly make this work?
  ```
  However, we could start with the current impl, and see if we can't fit in extra behaviour in the future
* I'm toying with the idea of adding [integration tests](https://github.com/doctest/doctest/tree/1007-integration-tests/tests/integration) to replace some of our examples; this could be a good candidate as well for those. If preferred, I can focus a bit more on that task to make the groundwork done for you?

doctest/parts/public/macros.h
aozgaa: I'm not sure why I ended up with this but the intention was to make `GENERATE(42)` work (ie: generate with one arg), and this indirection was added to help in that case. But maybe my initial macro was just buggy.

Updating.

doctest/parts/public/generator.h
aozgaa: I think this a [pack expansion](https://en.cppreference.com/w/cpp/language/parameter_pack.html) rather than a fold (since there is no binary operator like `+`,`-`, or `,`) on the line, and should be ifne since c++11.

I'm not sure if the CI has a C++11 build to exercise this (funnily enough, gcc at least never had c++11 as a default version from what i can tell).

... but at least when I add `-DCMAKE_CXX_STANDARD=11` on my dev box (ubuntu 24.04), all tests pass.

doctest/parts/public/macros.h
aozgaa: actually I see it now (again) -- the simpler approach:
```
#define DOCTEST_GENERATE(first, ...) (first)
```
doesn't work with some clang (many versions) which errors with
```
/workspace/examples/all_features/generators.cpp:144:14: error: must specify at least one argument for '...' parameter of variadic macro [-Werror,-Wgnu-zero-variadic-macro-arguments]
    auto x = GENERATE(42);
             ^
/workspace/doctest/doctest.h:3629:51: note: expanded from macro 'GENERATE'
#define GENERATE(...) DOCTEST_GENERATE(__VA_ARGS__)
                                                  ^
/workspace/doctest/doctest.h:3284:9: note: macro 'DOCTEST_GENERATE' defined here
#define DOCTEST_GENERATE(first, ...) (first)
        ^
1 error generated.
```
in this case.

Reverted.

(but there may be some more idiomatic way to do this -- please share if so!)

examples/all_features/generators.cpp
aozgaa: switched to using `MESSAGE` everywhere.

examples/all_features/generators.cpp
aozgaa: nice! Wasn't sure if using a static `iteration` inside a test case is a great idea (but then I have a somewhat naive mental model about whether we'd ever re-run a test multiple times within a single execution -- can this ever happen?)

aozgaa: changed!

examples/all_features/generators.cpp
aozgaa: seems using `DOCTEST_MSVC_SUPPRESS_WARNING_WITH_PUSH` instead fixes it.

doctest/parts/public/macros.h
aozgaa: I ran a half-assed comparison (thanks LLM) and it looks like in terms of:
1) compilation-time: `GENERATE` is about 4x faster than `SUBCASE(<varying string>)`
2) run-time: `GENERATE` is about 5x faster than `SUBCASE(<varying string>)`.

I also tried to write a generate-like macro which uses pack-indexing to avoid materializing the array and it is pretty close in terms of performance (within 5% on my one-off runs).

This is all a bit handwavy and machine-specific but hopefully gives some directional indication?

Could tidy up into something publishable. Maybe a build/runtime performance regression would make sense for CI? (but only a vague thought since I know maintaining such stuff is a time sink)

aozgaa:
> It may be good to split out the traversal.cpp changes into a separate PR; I suspect this will fix a few existing issues, so having a separate commit for it would be nice for traceability / bisecting errors when/if they occur. We can also add some [regression tests](https://github.com/doctest/doctest/tree/dev/tests/regression) to help keep them from occurring again

Yes, totally reasonable. I can take a crack at this (may take a few days with other stuff going on)

aozgaa:
> * The current `GENERATOR` impl. is good for fixed values, but I anticipate situations where we might want to use a container-type, e.g.
>   ```c++
>   auto data = std::vector(64);
>   std::iota(data.begin(), std::end(), 0); // data = { 0, 1, ..., 63 };
>   auto item = GENERATE(data); // Could we possibly make this work?
>   ```

Yeah I think a `GENERATE_RANGE` or similar could be doable, but prefer to leave that as follow-on for if/when needed.

aozgaa:
> I'm toying with the idea of adding [integration tests](https://github.com/doctest/doctest/tree/1007-integration-tests/tests/integration) to replace some of our examples; this could be a good candidate as well for those. If preferred, I can focus a bit more on that task to make the groundwork done for you?

Lacking context on that work and use-cases, it's hard for me to say. Is there a particular test that you'd prefer to run via integration compared to the golden-output diffing? Something like "parse the xml output and check it has 8 subcases/generates"?

That could be nice, but it's mainly an ergonomic benefit compared to manually parsing the `test_outputs/*.xml` when the baseline changes down the road.

doctest/parts/public/generator.h
mitchgrout: Ah my bad, this would indeed fall under pack expansions.

doctest/parts/public/macros.h
mitchgrout: Yeah I was stung by `-Wgnu-zero-variadic-macro-arguments` only just today; it's a real pain. 

C++11 mandates that macro-expansion will specify at-least every argument, including var-args. However, you can get around it by just having a trailing comma, which essentially is setting the var-args to a blank expansion.

In general, most compilers are fine with a missing varargs value, and only warn at higher levels (e.g. for `gcc`, you need `-pedantic` afaik).

examples/all_features/generators.cpp
mitchgrout: From memory no, a test-case _should_ be run only as many times as there are unique traversals through it. I try to use `static` state inside test-cases sparingly as it leaks across traversals, but in this instance it's exactly what we're after

mitchgrout:
> > I'm toying with the idea of adding [integration tests](https://github.com/doctest/doctest/tree/1007-integration-tests/tests/integration) to replace some of our examples; this could be a good candidate as well for those. If preferred, I can focus a bit more on that task to make the groundwork done for you?
> 
> Lacking context on that work and use-cases, it's hard for me to say. Is there a particular test that you'd prefer to run via integration compared to the golden-output diffing? Something like "parse the xml output and check it has 8 subcases/generates"?
> 
> That could be nice, but it's mainly an ergonomic benefit compared to manually parsing the `test_outputs/*.xml` when the baseline changes down the road.

Have a look at #1007 if you're interested. The short of it is, I'm trying to convert most of our `examples/` into smaller test-cases which do the golden-output diffing internally, rather than depending on `cmake` to do it for us. IMO it makes it a bit easier to see the impact of a change, and will hopefully help to prevent problems like #1029, since seemingly we just blindly added the golden-output file without really checking _what_ it was saying.

Ultimately I'll be converting most of our `examples/` into these tests, so we can have more user-friendly examples instead. It's not a hard requirement to do one or the other ultimately

mitchgrout:
Sorry for the wait. Happy with the general implementation - when you have the time, can you please rebase onto `dev` and I'll merge in when you're happy.

This and #1083 are the only major bits left before I'll release a v2.5 candidate - would be good to give people a bit of time to trial it in their own projects before committing to an immutable v2.5.0 release.

aozgaa:
> Sorry for the wait.

Not at all, in a happy coincidence I've been vacationing.

I've just pushed by way of merge, but will fixup the CI/linting in the next few commits. Will @ you when ready.

Also, it sounds like a rebase-workflow is preferred in this repo over merging dev? (I see very few merge commits in git history, like 1 in 2018).

Do you do squash on merge or actually strongly prefer me to rebase? (I'm indifferent, have done these things both ways in the past)

aozgaa:
@mitchgrout this looks good now. Let me know if you prefer a rebased/clean history prior to merge. Okay to squash down to one commit with a message like `add GENERATE macro`.

examples/all_features/generators.cpp
```diff
@@ -0,0 +1,458 @@
...
+}
+
+TEST_CASE("multiple GENERATEs on a single line") {
+    // clang-format off
+    // NOLINTNEXTLINE(readability-isolate-declaration)
+    const int i = GENERATE(1, 2);
+    const int j = GENERATE(10, 20);
+    // clang-format on
+    MESSAGE("i=", i, " j=", j);
+}
```
mitchgrout, line 67: iirc these used to be on the same line, right? Probably just a clang-format thing.

examples/all_features/generators.cpp
```diff
@@ -0,0 +1,458 @@
...
+        for (int i = 0; i < 3; ++i) {
+            SUBCASE("fixed name") {
+                js.push_back(3 + 2 * i + GENERATE(0, 1));
+                std::ostringstream oss;
+                oss << "i=" << i << " js=[ ";
+                for (auto j: js) {
+                    oss << j << " ";
+                }
+                oss << "]";
+                MESSAGE(oss.str());
```
mitchgrout, line 311: We now support [stringifying STL constructs](https://github.com/doctest/doctest/pull/1074)

examples/all_features/generators.cpp
```diff
@@ -0,0 +1,458 @@
...
+            }
+        }
```
mitchgrout, line 313: Looking at this one a bit closer, I'm a little confused actually. The relevant output from `generators.cpp.text` is:

```text
generators.cpp(0):
TEST CASE:  GENERATE with fixed for loop
  loop with fixed name
  fixed name

generators.cpp(0): MESSAGE: i=0 js=[ 3 ]

generators.cpp(0):
TEST CASE:  GENERATE with fixed for loop
  loop with fixed name
  fixed name

DEEPEST SUBCASE STACK REACHED (DIFFERENT FROM THE CURRENT ONE):
  loop with fixed name
  fixed name
  fixed name

generators.cpp(0): MESSAGE: i=1 js=[ 3 5 ]

generators.cpp(0):
TEST CASE:  GENERATE with fixed for loop
  loop with fixed name
  fixed name

DEEPEST SUBCASE STACK REACHED (DIFFERENT FROM THE CURRENT ONE):
  loop with fixed name
  fixed name
  fixed name
  fixed name

generators.cpp(0): MESSAGE: i=2 js=[ 3 5 7 ]
```

i.e. we re-enter `SUBCASE("fixed name")` multiple times based off the loop. When I see `for (...) { SUBCASE("...") { ... } }`, I usually interpret doctest as needing to _unroll_ the loop into `N` distinct sibling subcases. In this instance though, we're interpreting it as-if the `for` loop was inside the `SUBCASE`?

Just wanting to confirm whether this is intentional. I believe this would be a departure from the current v2.4 behaviour, but I'll have to check.

examples/all_features/generators.cpp
```diff
@@ -0,0 +1,458 @@
...
+        std::vector<int> js;
+        // N.B. this produces 3 * 2 = 6 traversal leaves: each loop iteration has a
+        // distinct subcase identity, and each of those subcases has its own
+        // GENERATE(0, 1). State resets for each rerun, so the vector contains
+        // only the value for the current dynamic subcase.
+        // This is analogous to:
+        // int i = GENERATE(0, 1, 2);
+        // int j = GENERATE(0, 1);
+        for (int i = 0; i < 3; ++i) {
+            SUBCASE(("dynamic " + std::to_string(i)).c_str()) {
```
mitchgrout: We also now support [conversion from `std::string` to `doctest::String`](https://github.com/doctest/doctest/pull/1081)

mitchgrout:
> @mitchgrout this looks good now. Let me know if you prefer a rebased/clean history prior to merge. Okay to squash down to one commit with a message like `add GENERATE macro`.

Merge strategy is squash by-default. As required I usually just update the squash-commit message and drop the body message unless retaining it is useful. No need to manually squash.

mitchgrout (commented):
One small point of concern, my bad for not noticing earlier. I'll do some poking around to see what doctest v2.4.0 - v2.4.12 has done up until this point.

examples/all_features/generators.cpp
mitchgrout: I set up the following test harness and ran it across `v2.4.{1 .. 12}`, and on this branch:

```cpp
#include <doctest/doctest.h>
#include <string>

TEST_CASE("Iterated fixed sections") {
  auto state = 0u;
  for (auto idx = 0u; idx < 3; ++idx) {
    SUBCASE("Section") {
      ++state;
      MESSAGE("Section = ", idx, " state = ", state);
    }
  }
}

TEST_CASE("Iterated dynamic sections") {
  auto state = 0u;
  for (auto idx = 0u; idx < 3; ++idx) {
    SUBCASE(("Section" + std::to_string(idx)).c_str()) {
      ++state;
      MESSAGE("Section = ", idx, " state = ", state);
    }
  }
}
```

...the intent being to use `state` to check if a subcase was being genuinely entered. The results are interesting:

- `v2.4.1 .. v2.4.8`:
    - Iterated fixed sections: 
        - Section = 0, state = 1
    - Iterated dynamic sections:
        - Section = 0, state = 1
        - Section = 1, state = 1
        - Section = 2, state = 1  
- `v2.4.9 .. v2.4.12 + aozgaa:ao/generator-13`:
    - Iterated fixed sections: 
        - Section = 0, state = 1
        - Section = 1, state = 2
        - Section = 2, state = 3
    - Iterated dynamic sections:
        - Section = 0, state = 1
        - Section = 1, state = 1
        - Section = 2, state = 1  

On a hunch, I checked the before/after commits of #598, which was attempting to fix subcase re-entry (e.g. what happens in v2.4.1 - v2.4.8 for the fixed section case) -- it is indeed the case that #598 changed this.

Given that your change aligns with the _current_ behaviour, I'm happy to accept the change as-is, but I think that the behaviour as it exists _probably_ needs to be changed to match the output observed for dynamic sections.

examples/all_features/generators.cpp
aozgaa: changed, removed `.c_str()` references in this file.

examples/all_features/generators.cpp
aozgaa: changed to use (minor output formatting change but results are semantically the same)

examples/all_features/generators.cpp
aozgaa: yep, moved them back to the same line, thanks for double checking.

examples/all_features/generators.cpp
aozgaa: I didn't start with a strict mental model of this behavior when I started but did observe this with the test cases as well. I didn't do the backwards investigation and just accepted the starting behavior (ie: what was on `dev`, which was something like 2.4.12).

My internal monologue was something like:
1) loop around constant name:
```
for(...) {
  SECTION("fixed name") {
     // block
  }
}
```
doesn't really have a clean interpretation, and is somewhat UB since we don't runtime check the number of times we've entered a SECTION and count those distinctly (the regression test behavior is somewhat silly imo).
2) loop with dynamic name:
```
for(auto i: ...) {
  SECTION(std::to_string(i)) {
     // block
  }
}
```
behaves like the `SECTION`s are siblings, that is:
```
SECTION("1") { ...}
SECTION("2") { ...}
SECTION("3") { ...}
...
```
This was previously documented to work the the implementation retains that in the presence of `GENERATE` (which act like a bunch of sibling subclasses).

aozgaa: Maybe it would make sense to explicitly document the fixed section name as unsupported?

examples/all_features/generators.cpp
aozgaa: I think the implementation is buggy, or at least not deterministic across platforms, rendering it unusable for golden tests. I see CI failures like this depending on the compiler/platform:
```
  -generators.cpp(0): MESSAGE: i=1 js=[3, 5]
  +generators.cpp(0): MESSAGE: i=1 js={3, 5}
```

aozgaa: reverted to manual `ostringstream` while this gets sorted out. (preferably doesn't block this PR)

examples/all_features/generators.cpp
aozgaa: documentation added.

examples/all_features/generators.cpp
aozgaa: for posterity -- here is a failing CI run when using the doctest-based stringification: https://github.com/doctest/doctest/actions/runs/23572214307

mitchgrout: Ah that's fun. That's coming from [`examples/all_features/stringification.cpp`](https://github.com/doctest/doctest/blob/dev/examples/all_features/stringification.cpp#L101-L114) which declares an overload of `operator<<` for `std::vector<T>`. I assumed it wouldn't bleed through to this, but given it's not marked `static` it probably is.

All good for now to keep your inlined stringification. I'll probably just change `stringification.cpp` to have its own types to do this for, rather than screwing with `std`.

examples/all_features/generators.cpp
mitchgrout: Yeah it's kinda bordering on undefined behaviour IMO. Thanks for the additional docs. I think I may just leave it be at the moment unless I can find a real simple fix for it, but I guess _as it stands_ it's backwards-compatible in a real strange way.

doctest/parts/public/macros.h
mitchgrout: No need to worry about noting this down; just wanting to check that `GENERATE` is at least comparable to repeated `SUBCASE`

PR #1053, opened 2026-03-05
Fixes misleading template test name formatting

### LINKED ISSUE #1052: Misleading template test name formatting in some rare cases
Let's say I have a test defined as:
```c++
TEST_CASE_TEMPLATE("Test Set/TSet", TestPair, Typelist<Set, Element>) {...}
```

When failed, it will generate a report like:
```
TEST CASE:  Test empty Set/TSet<Typelist<Set, Element>>
```

The way the template argument is directly concatenated and wrapped in < and > makes it seem like TSet is specialized with the wrong parameter. It is a niche case, but could potentially save a couple of minutes of pondering and minimize misunderstanding in the future.

I propose the formatting to be changed to:
```
TEST CASE:  Test empty Set/TSet with template argument Typelist<Set, Element>
```

Epixu:
## Description
Changes the way templated test names are logged.

## GitHub Issues
Closes #1052

### DIFF
--- a/doctest/parts/private/test_case.cpp
+++ b/doctest/parts/private/test_case.cpp
@@ -57,7 +57,7 @@ TestCase &TestCase::operator*(const char *in) {
     m_name = in;
     // make a new name with an appended type for templated test case
     if (m_template_id != -1) {
-        m_full_name = String(m_name) + "<" + m_type + ">";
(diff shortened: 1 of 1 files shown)


mitchgrout:
Moving this to draft for the time being (see https://github.com/doctest/doctest/issues/1052#issuecomment-4002403826).

PR #1056, opened 2026-03-06
fix: Rework SUBCASE traversal

### LINKED ISSUE #782: endless loop with multiple `DOCTEST_SUBCASE`s when an exception is thrown prior to any subcase
The following file causes an endless loop:
```c++
#include <doctest/doctest.h>

TEST_CASE("wtf")
{
    static int cnt = 0;
    if (cnt++ >= 1) {
        throw std::runtime_error{"wat"};
    }
    DOCTEST_SUBCASE("1") {}
    DOCTEST_SUBCASE("2") {}
}
```
The expected behavior is that this simply reports an unexpected exception, and terminates cleanly.

* Operating System: **NixOS**
* Compiler+version: **clang 16.0.1**, **GCC 12.2.0** (both produce the same results)

aozgaa:
## Description

This factors out the `SUBCASE` traversal/path rework from [#1047](https://github.com/doctest/doctest/pull/1047) into a separate PR.

The main motivation is the review feedback from [mitchgrout](https://github.com/doctest/doctest/pull/1047#pullrequestreview-3879091086) suggesting that the `traversal.cpp` changes should land separately for traceability / bisectability, especially since they may also fix existing `SUBCASE` issues on their own. This PR is that split.

### User-facing behavior:
- No new public API.
- No `GENERATE(...)` support in this PR.
- `SUBCASE` traversal is reworked to use the new path/decision traversal model introduced in `traversal.h/.cpp`.
- Expected behavior should remain the same for existing tests/examples, aside from fixes to traversal edge cases.

### Implementation approach:
- Introduce `doctest/parts/private/traversal.{h,cpp}` as the traversal model for `SUBCASE` reruns.
- Replace the older subcase-specific backtracking state in `ContextState` with `TraversalState`.
- Track traversal as:
  - the selected decision path
  - the set of sibling subcases discovered at each depth on a rerun
  - active subcase depth for filtering/reporting
- Keep fatal unwinding integrated with the traversal state so reporter callbacks still receive balanced `subcase_end` events.
- Remove all `GENERATE`-related public API, docs, examples, and traversal bookkeeping from this branch so the change is limited to the `SUBCASE` traversal refactor.

### Test coverage:
- Add dedicated unit tests for `TraversalState`:
  - traversal over discovered subcase paths
  - unwinding active subcases after abrupt exit
- Re-run existing `SUBCASE`-related coverage and golden-output tests to confirm behavior remains stable.

### Notes:
- This is intentionally a factoring PR, not a feature PR.
- The goal is to isolate the traversal refactor so it can be reviewed and bisected independently from the value-parameterization work in [#1047](https://github.com/doctest/doctest/pull/1047).
- If/when `GENERATE(...)` comes back, it can be rebased onto this smaller traversal-only change.

## GitHub Issues

Fixes #782


aozgaa:
Coverage gap in `handler.cpp` is similar to other PR -- it is really finicky to test and also specify exactly what should happen here (as in general signal handling esp in presence of tests with threading is very subtle). And in any event this testing gap is not new.

mitchgrout:
Seems that `TraversalState::advance()` can trigger an assertion in `std::vector::operator[]` . Was testing against the MVP for #782 on gcc-15 and hit:

`/usr/local/include/c++/15.1.0/bits/stl_vector.h:1263: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](size_type) [with _Tp = doctest::detail::DecisionPoint; _Alloc = std::allocator<doctest::detail::DecisionPoint>; reference = doctest::detail::DecisionPoint&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.`

On `dev` the behaviour is that we hit an infinite-loop

aozgaa:
Ah yeah there is a baked in assumption that we don't throw "sometimes" before reaching a subcase.

Maybe we can discuss in the original issue in terms of desired behavior / traversals?

We can address here or in a follow-up?

aozgaa:
repro'd the assertion failure by building with `_GLIBCXX_ASSERTIONS` against the added tests, as in
```
CXXFLAGS=-D_GLIBCXX_ASSERTIONS cmake -S . -B build  -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -GNinja --fresh -DDOCTEST_TEST_MODE=COLLECT && cmake --build build && ctest --test-dir build --output-on-failure)
````
then fixed with the above changes which prune the traversal.

File: tests/regression/test.782.cpp [hunk 1]
```diff
@@ -0,0 +1,16 @@
+#define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
+#include <doctest/doctest.h>
+
+TEST_CASE("inconsistent init before subcases") {
+#ifndef DOCTEST_CONFIG_NO_EXCEPTIONS
+    static int cnt = 0;
+    if (++cnt == 2) {
+        throw std::runtime_error{"#782 repro exception"};
```
aozgaa, line 8: I know could have used `throw_if` but then the test regex matching in the cmake properties is awkward to setup reasonably when actually running with exceptions.

(though it seems like the golden output tests are wildly inconsistent if compiling with `-fno-exceptions  -DDOCTEST_CONFIG_NO_EXCEPTIONS_BUT_WITH_ALL_ASSERTS`)

aozgaa:
... the CI failure seems unrelated to this change -- it seems like:
1) a network failure led to failures to fetch an up-to-date package manifest:
```
2026-03-08T04:07:31.1886074Z Reading package lists...
2026-03-08T04:07:31.2006853Z W: Failed to fetch http://archive.ubuntu.com/ubuntu/dists/bionic/InRelease  Could not resolve 'archive.ubuntu.com'
2026-03-08T04:07:31.2008237Z W: Failed to fetch http://archive.ubuntu.com/ubuntu/dists/bionic-updates/InRelease  Could not resolve 'archive.ubuntu.com'
2026-03-08T04:07:31.2009730Z W: Failed to fetch http://archive.ubuntu.com/ubuntu/dists/bionic-backports/InRelease  Could not resolve 'archive.ubuntu.com'
2026-03-08T04:07:31.2010898Z W: Some index files failed to download. They have been ignored, or old ones used instead.
```
2) so an older version of cmake (`3.10.2`) was used:
```
2026-03-08T04:07:50.8437529Z Preparing to unpack .../51-cmake_3.10.2-1ubuntu2.18.04.2_amd64.deb ...
2026-03-08T04:07:50.8443867Z Unpacking cmake (3.10.2-1ubuntu2.18.04.2) ...
```
3) and this one can't handle the "modern" cli invocations like `cmake -S . -B build`
```
2026-03-08T04:08:12.7772424Z cmake -S . -B build -D CMAKE_BUILD_TYPE=Debug -D DOCTEST_TEST_MODE=COMPARE -D DOCTEST_INTERNAL_WERROR=ON -D CMAKE_CXX_COMPILER=clang++-5.0
2026-03-08T04:08:12.7827060Z ##[error]Process completed with exit code 2.
```

Basically, it looks like the CI itself is flaky in some way.
Adding some `set -euo pipefail` might help identify such issues earlier in the pipeline

mitchgrout:
Sorry for the wait. Just rebased onto `dev` to fix some minor conflicts with changes since the PR was first put up.

Regarding the CI, yeah it's a bit flaky. The docker-based jobs now correctly `set -eu` to ensures earlier steps terminate when they fail, but it's still a bit annoying to have to manually re-run jobs. If you've got any fixes I'm happy to look over, but otherwise I'll try to get around to them in sometime in the next few months.

PR #1067, opened 2026-03-12
chore: Add dedicated wrapper file for `windows.h`

Repiteo:
## Description
Was originally part of a PR attempting to trim/remove `prelude.h`, but the Windows header specifically sent me down a rabbit-hole that's better to tackle in isolation. It turns out that `DOCTEST_MAKE_STD_HEADERS_CLEAN_FROM_WARNINGS_ON_WALL_BEGIN` really only exists for `windows.h`, and as such could have it and the cascading logic consolidated into a single wrapper file. This had the following effects:
- Removes the need for `DOCTEST_MAKE_STD_HEADERS_CLEAN_FROM_WARNINGS_ON_WALL_BEGIN` entirely; all instances of it have been removed
- `prelude.h` is now quite a bit lighter, with all windows-specific logic relocated to either the wrapper header or the specific header/implementation that requires it
- `ticks_t` converted to `uint64_t`, now platform-agnostic
- Disabled `C4738` for 32-bit release builds in `common.cmake`, as an optimization bug was causing the warning to appear in `corecrt_math.h`
- `reportFatal` relocated from `private/assert/handler.h` to `private/signals.h`, as that's the only header/impl to use it
- Removed `-Wnonportable-system-include-path` from common clang warnings, as it's only necessary for the windows-specific headers, which got appropriately suppressed in the wrapper specifically
- Handful of examples/tests that included `<windows.h>` now include `<doctest/parts/private/windows_wrapper.h>` instead, with a note that just `windows.h` is preferable when not enabling all warnings


mitchgrout:
As a heads-up, I just merged #1066 which should make the CI speed _significantly_ better (about 3x faster). I'll have a look over this change proper later in the day, but skimming it, nothing seems overtly wrong.

Does feel a bit odd to create a header thats sole purpose is to include another header, but it solves a very particular problem I suppose.

doctest/parts/private/windows_wrapper.h
mitchgrout, line 1: Naming-wise, I'd probably go with `doctest/parts/private/ext/windows.h`

doctest/parts/private/signals.h
```diff
@@ -3,6 +3,22 @@
...
+#if defined(WINAPI_FAMILY) && (WINAPI_FAMILY == WINAPI_FAMILY_APP) && !defined(DOCTEST_CONFIG_NO_WINDOWS_SEH)
+#define DOCTEST_CONFIG_NO_WINDOWS_SEH
+#endif
+#if DOCTEST_MSVC && !defined(DOCTEST_CONFIG_WINDOWS_SEH)
+#define DOCTEST_CONFIG_WINDOWS_SEH
+#endif
+#if defined(DOCTEST_CONFIG_NO_WINDOWS_SEH) && defined(DOCTEST_CONFIG_WINDOWS_SEH)
+#undef DOCTEST_CONFIG_WINDOWS_SEH
+#endif
+
```
mitchgrout: Long-term I'd like to keep all `DOCTEST_CONFIG_*` shenanigans in `config.h` where possible. Generally just makes observability a bit nicer, and can help with testing

scripts/cmake/common.cmake
```diff
@@ -202,6 +202,10 @@ if(MSVC)
     add_compiler_flags(/permissive-)   # force standard conformance - this is the better flag than /Za
     add_compiler_flags(/Wall) # turns on warnings from levels 1 through 4 which are off by default - https://msdn.microsoft.com/en-us/library/23k5d385.aspx
 
+    if (CMAKE_SIZEOF_VOID_P EQUAL 4 AND CMAKE_BUILD_TYPE STREQUAL "Release")
+        add_compiler_flags(/wd4738)
+    endif()
+
```
mitchgrout: What's the reasoning here?

tests/doctest/test_debugger.cpp
```diff
@@ -1,18 +1,15 @@
 #include <doctest/doctest.h>
 
-DOCTEST_MAKE_STD_HEADERS_CLEAN_FROM_WARNINGS_ON_WALL_BEGIN
 #if defined(DOCTEST_PLATFORM_WINDOWS)
-  DOCTEST_CLANG_SUPPRESS_WARNING_WITH_PUSH("-Wnonportable-system-include-path")
-  #include <windows.h>
-  DOCTEST_CLANG_SUPPRESS_WARNING_POP
+  // Prefer `#include <windows.h>`, but this uses our wrapper to sidestep `/Wall` errors.
```
mitchgrout: Can leave this comment in `windows_wrapper.h`; don't need to copy it to `#include` locations

doctest/parts/private/signals.h
Repiteo: I originally moved them because it seemed like a requirement for `windows.h` to define them, but I think they're actually something predefined when targeting something like UWP? I'll go ahead and move this back to `config.h` next push

scripts/cmake/common.cmake
Repiteo: > Disabled `C4738` for 32-bit release builds in `common.cmake`, as an optimization bug was causing the warning to appear in `corecrt_math.h`

Specifically: 32-bit with optimizations enabled was causing it. Not sure why, but this suppresses it cleanly

doctest/parts/private/windows_wrapper.h
Repiteo: I don't think it can be called `windows.h`, as that makes including the system header throw a bit of a fit, but I could maybe use something like `win32.h` instead

mitchgrout:
As a heads-up, #1070 has a small tweak to `reportFatal` that will need to be propagated to `signals.cpp`

doctest/parts/private/windows_wrapper.h
mitchgrout: `win32.h` could work, but at that stage I'd also assume it pulls in `io.h` as a whole 'os-support' header and the like. Name is ultimately not a big problem

Repiteo:
Went with `doctest/parts/private/ext/windows.h` for the name & migrated the relevant `DOCTEST_CONFIG_*` options back to `config.h`. Also dropped the `common.cmake` conditional to showcase the issue more clearly (assuming it still pops up after the CI changes)

Repiteo:
Huh, looks like the issue resolved itself. In that case, I'll keep the cmake file unchanged

scripts/cmake/common.cmake
Repiteo: Turned out to be unnecessary after the CI changes; removed

doctest/parts/private/windows_wrapper.h
Repiteo: My concerns were unfounded; the `windows.h` name worked just fine

mitchgrout:
> Huh, looks like the issue resolved itself. In that case, I'll keep the cmake file unchanged

Probably was an issue on older compilers under a particular (probably x86) build. Since #1066 has disabled those runs, the problem is no longer visible.

doctest/parts/private/ext/windows.h
```diff
@@ -0,0 +1,61 @@
+#ifndef DOCTEST_PARTS_PRIVATE_EXT_WINDOWS
+#define DOCTEST_PARTS_PRIVATE_EXT_WINDOWS
+
+#include "doctest/parts/public/config.h"
+
+#ifdef DOCTEST_PLATFORM_WINDOWS
```
mitchgrout, line 8: Was thinking about this for a bit, we may want to exclude this. My thinking is, `ext/windows.h` is our wrapper around `windows.h` to make it safe to use. But, it's still essentially `windows.h`. So it should probably be the callers obligation to check if they should even be using it.

doctest/parts/private/color.cpp
```diff
@@ -1,8 +1,13 @@
 #include "doctest/parts/private/prelude.h"
 #include "doctest/parts/private/context_state.h"
+#include "doctest/parts/private/ext/windows.h" // IWYU pragma: keep
 
 DOCTEST_SUPPRESS_PRIVATE_WARNINGS_PUSH
 
+#ifdef DOCTEST_PLATFORM_WINDOWS
+#include <io.h>
+#endif
+
```
mitchgrout, line 8: Just noticed but is this meant to be under `PRIVATE_WARNINGS`?

mitchgrout (dismissed):
Nothing major, up to you if we want to adjust any further. Let me know when you're ready to merge

doctest/parts/private/ext/windows.h
Repiteo: That's a good point; I'll adjust it so the caller provides the wrapper

doctest/parts/private/color.cpp
Repiteo: Oh shoot, good catch! Part of this pass was seeing if other windows include suppressions were bogus, and it turned out that this file needed no suppressions whatsoever. But I forgot to swap this back while testing, so lemme fix that

Repiteo:
This *should* be ready to merge, but I don't know enough about the coverage setup to know why that action is now failing

mitchgrout:
> This _should_ be ready to merge, but I don't know enough about the coverage setup to know why that action is now failing

See #1028 -- I added in a [`--remap` option](https://github.com/doctest/doctest/blob/dev/scripts/assemble.py#L108-L120) to `scripts/assemble.py` which adds in a `#line <file> <line>` directive when building `doctest/doctest.h` so that coverage report could link back to individual source/header files; otherwise, it just points everything to `doctest.h` and the report is borderline unreadable.

I suspect it's interacting poorly with the `IWYU` pragma. In `color.cpp` you've got it as `IWYU pragma: keep`, but `scripts/assemble.py` is configured to exclude _any_ line with `IWYU pragma:` in it. Maybe it needs to be tweaked to retain the line but strip the comment if it's `IWYU pragma: keep`?

Repiteo:
Adjusted the script to simply strip the pragma comments, while keeping the lines in-tact. Unfortunately, that didn't seem to fix the issue.

Repiteo:
Is there any benefit to running coverage on `doctest/doctest.h` over the individual `doctest/parts/private/*.cpp` files?

mitchgrout:
> Adjusted the script to simply strip the pragma comments, while keeping the lines in-tact. Unfortunately, that didn't seem to fix the issue.

I'll see if I can try debugging it locally today.

mitchgrout:
> Is there any benefit to running coverage on `doctest/doctest.h` over the individual `doctest/parts/private/*.cpp` files?

It's mostly a convenience. Trying to load the full doctest.h file on codecov is super laggy, which makes it hard to see exactly what it's complaining about, esp. if it's regarding patch coverage. 

It's not a hard requirement, just a real nice-to-have. If it causes long term problems I'm happy to revert it.

Repiteo:
What I mean is: do we need to run code coverage on that file at all? If we're pretending it's the individual source files instead, I would assume we could just run code coverage on the implementation files directly

Take this with a grain of salt; I don't use code coverage myself, so I could be entirely misunderstanding the use-case

mitchgrout:
> What I mean is: do we need to run code coverage on that file at all? If we're pretending it's the individual source files instead, I would assume we could just run code coverage on the implementation files directly

The problem mostly boils down to the fact that:

- Tests in `examples/` use the single-header form, so coverage will produce info relative to that file
- Tests in `tests/` use the library form, so coverage will produce info relative to `doctest/parts/*`
- The coverage instrumentation does not intrinsically know that these are the same files (unless the #line directive is present)

For the moment, the `--remap` stuff helps to link these two together, so we get accurate coverage information. In the long-run, it would be nice if every test just utilized the library form instead of the single-header form. But, since we expect users of the project to use the single-header form, we would want to at least have _some_ tests (or entire CI runs) use the single-header form, to avoid any issues that crop up during amalgamation (esp. things like duplicate symbols).

mitchgrout:
> I'll see if I can try debugging it locally today.

Strange, doesn't reprod for me locally. I'll keep poking around to see if I can figure out what's going wrong.

mitchgrout:
Got a hunch as to what the problem is. When I first wrote `scripts/assemble.py` I realised it would eventually be a problem, but since it wasn't affecting anything yet, didn't do anything to resolve it

During assembly, whenever we see one of 'our' files, we recurse and start spitting out the contents of the included header (if not already done). This generally works fine, but doesn't respect semantics of something like...

```cpp
// s1.cpp
#if defined(USE_SOMETHING)
  #include <something.h>
#endif

// s2.cpp
#include <something.h>
```

In this case, if `s1` is processed first, `something.h` gets included if-and-only-if `USE_SOMETHING` is defined. So in `s2`, the include gets skipped since it's already been processed. But, this now changes the semantics to `s2` only including `<something.h>` if `USE_SOMETHING` is defined.

Now in our case, every source file is _mostly_ including `doctest/parts/private/ext/windows.h` in the same way (except `signals.h` but that's probably not an issue); that is to say, we have a pattern of:

```cpp
#ifdef DOCETST_PLATFORM_WINDOWS
  #include "doctest/parts/private/ext/windows.h"
#endif
```

...which was something I requested so that `<.../ext/windows.h>` is just a drop-in replacement for `<windows.h>`. 

I suspect that the `--remap` stuff isn't playing nicely with it. We could verify this by moving the guard back in, and dropping the guard at all the include sites.

As to why that only breaks on the pipeline, I'm kinda stumped. But, it's the best answer I have at the moment.

mitchgrout:
If preferred, we can merge this as-is and I can deal with rectifying the coverage job in another PR

Repiteo:
Ahhh, I see. In that case, we'd probably want to keep the platform wrapper within the file after all

As a heads-up: I'm not by my usual workstation atm, so it may take a bit for me to adjust that

mitchgrout:
Since this is more a QoL chore I'm going to re-target this for v2.6 since we seem to have all the major v2.5 changes in. Hopefully it'll be tagged this weekend, assuming I can't find anything weird with it

Repiteo:
Returned the wrapper back to the file itself, with a `TODO` noting the issue and that it's meant to be a temporary measure

tests/doctest/test_color.win32.cpp
```diff
@@ -2,7 +2,7 @@
 #include <doctest/doctest.h>
 #include <utils/callback.h>
 
-#if defined(DOCTEST_CONFIG_COLORS_WINDOWS)
+#if defined(DOCTEST_CONFIG_COLORS_WINDOWS) && defined(_MSC_VER)
```
mitchgrout: Why do we need `_MSC_VER` here? I'd assume that filters out `mingw64` when compiling with `gcc`, so this test would end up being a no-op.

doctest/parts/public/std/fwd.h
```diff
@@ -6,15 +6,14 @@
...
-DOCTEST_MAKE_STD_HEADERS_CLEAN_FROM_WARNINGS_ON_WALL_BEGIN
 #include <cstddef>
 #include <ostream>
 #include <istream>
-DOCTEST_MAKE_STD_HEADERS_CLEAN_FROM_WARNINGS_ON_WALL_END
 #else // DOCTEST_CONFIG_USE_STD_HEADERS
 
-// Forward declaring 'X' in namespace std is not permitted by the C++ Standard.
-DOCTEST_MSVC_SUPPRESS_WARNING_WITH_PUSH(4643)
+DOCTEST_MSVC_SUPPRESS_WARNING_PUSH
```
mitchgrout, line 14: Missing `_POP`?

mitchgrout (commented):
Just two small notes.

mitchgrout:
In case you were wondering about the `gcc-15` job, it started failing while implementing #1110 (a pretty routine addition of a new compiler). I'll look into it in a bit, but for now we'll just have to live with the failure (at least gcc-16 works, so I'm not expecting any gcc-15 specific errors to crop up anytime soon).

tests/doctest/test_color.win32.cpp
Repiteo: It failed when I was testing locally, but if it works on this CI then it might be an unrelated bug. Will remove

doctest/parts/public/std/fwd.h
Repiteo: This was just turning one warning suppression into two, so the previous `_POP` was reused

Issue #1102, opened 2026-04-04
Large compile-time regression with `DOCTEST_WITH_MAIN_IN_STATIC_LIB`


vittorioromeo:
### Description

Compiling Doctest with `DOCTEST_WITH_MAIN_IN_STATIC_LIB` enabled now results in the compilation of around ~30 separate TUs, each of which includes `prelude.h` -- see https://github.com/doctest/doctest/blob/master/CMakeLists.txt#L65C6-L102.

`prelude.h` includes a bunch of heavy headers, and includes this comment which I believe is now outdated:

> // required includes - will go only in one translation unit!

As a consequence, a full rebuild of my project skyrocketed from ~40s to ~130s:

```
33324 ms: /home/vittorioromeo/.cache/CPM/doctest/5fe2/doctest/parts/private/prelude.h (included 34 times, avg 980 ms), included via:
  33x: <direct include>
  1x: traversal.h 

11635 ms: /usr/include/c++/15.2.1/fstream (included 36 times, avg 323 ms), included via:
  33x: prelude.h 
  2x: <direct include>
  1x: traversal.h prelude.h 

11432 ms: /usr/include/c++/15.2.1/string (included 56 times, avg 204 ms), included via:
  33x: prelude.h fstream istream ios ios_base.h locale_classes.h 
  11x: <direct include>
  2x: ostream ostream.h ios ios_base.h locale_classes.h 
  2x: StringifyStdStringUtil.hpp 
  1x: iomanip ios_base.h locale_classes.h 
  1x: locale locale_classes.h 
  ...

11255 ms: /usr/include/c++/15.2.1/ios (included 42 times, avg 267 ms), included via:
  33x: prelude.h fstream istream 
  2x: ostream ostream.h 
  1x: iostream ostream ostream.h 
  1x: traversal.h prelude.h fstream istream 
  1x: blockingconcurrentqueue.h concurrentqueue.h thread stop_token std_thread.h unique_ptr.h ostream.h 
  1x: elf.hpp utils.hpp error.hpp microfmt.hpp iostream ostream ostream.h 
  ...

11178 ms: /home/vittorioromeo/OHW/VRSFML/examples/bubble_idle/BubbleIdleMain.hpp (included 30 times, avg 372 ms), included via:
  30x: <direct include>

3038 ms: /home/vittorioromeo/OHW/VRSFML/test/TestUtilities/Doctest.hpp (included 101 times, avg 30 ms), included via:
  101x: <direct include>

[...]

PR #1123, opened 2026-05-17
chore: Remove `prelude.h`

Repiteo:
- Requires #1067

## Description
Removes the fatty `prelude.h` from the repository entirely. This file was never meant to be a long-term solution, so getting it out of the way is ideal. In its place, all files now reference only the headers they require. Any logical changes were kept to an absolute minimum, though a few cstdlib functions were given a `std::` prefix to keep includes down.

## GitHub Issues
Indirectly addresses the root problem in #1102


doctest/parts/private/atomic.h
```diff
@@ -1,12 +1,27 @@
...
+#ifndef DOCTEST_CONFIG_NO_MULTITHREADING
+#define DOCTEST_DECLARE_MUTEX(name) std::mutex name;
+#define DOCTEST_DECLARE_STATIC_MUTEX(name) static DOCTEST_DECLARE_MUTEX(name)
+#define DOCTEST_LOCK_MUTEX(name) std::lock_guard<std::mutex> DOCTEST_ANONYMOUS(DOCTEST_ANON_LOCK_)(name);
+#else
+#define DOCTEST_DECLARE_MUTEX(name)
+#define DOCTEST_DECLARE_STATIC_MUTEX(name)
+#define DOCTEST_LOCK_MUTEX(name)
+#endif
+
```
mitchgrout, line 24: We do end up using `DOCTEST_{DECLARE,LOCK}_MUTEX` in a few other headers / TUs, but I'm happy for this to be here for the time being.

mitchgrout (dismissed):
Happy with the change, just need to double-check content after merging #1067

doctest/parts/private/atomic.h
Repiteo: Given those files would need to include `mutex.h` anyway, I think this is the best location for them

mitchgrout:
Just merged the prereq PR, so ready for rebase now.

Regarding `clang-tidy` we can either inhibit the duplicate-include rule, or just drop `doctest.h` from running on the assembled header. I'd lean more towards the first, just so we're linting the same file most consumers will be using.

mitchgrout (approved):
LGTM

PR #1132, opened 2026-05-18
ci: Enable GCC 15 on CI, bump checkout actions

Repiteo:
## Description
Tweaks the GitHub Actions in two small ways to ensure all runners are successful once more:
- Bumps all checkout actions from v5 to v6 (future-proofing)
- Change GCC 15 Ubuntu runner from 24.04 to 26.04 (`g++-15-multilib` is unavailable on 24.04)

## GitHub Issues
See: https://github.com/doctest/doctest/pull/1067#issuecomment-4473480092

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -30,6 +30,6 @@ jobs:
           - macos-latest
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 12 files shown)


mitchgrout (approved):
LGTM

PR #1103, opened 2026-04-09
build(cmake): Use single-source `doctest.cpp` for `doctest_with_main` target

mitchgrout:
## Description

Drops all `parts/*` sources from `doctest_with_main` in favour of a singular `doctest.cpp` source which internally just contains:

```cpp
#define DOCTEST_CONFIG_IMPLEMENT
#include "doctest.h"
```

This should approximate the original `doctest.cpp` file that we use to have in `parts/*` _before_ it was broken down into individual sources. By having a singular source, it should pay down some of the extreme include costs from `prelude.h`. This was benchmarked further to confirm:

```bash
$ hyperfine                                       \
  --warmup 3 --runs 16                          \
  --parameter-list commit "47ce36f,1af15d2"     \
  --setup "git checkout {commit}"               \
  --prepare "ninja -C ${BUILD_DIR} clean"       \
  "ninja -C ${BUILD_DIR} -j1 doctest_with_main" \

Benchmark 1: ninja -C build/gcc -j1 doctest_with_main [47ce36f]
  Time (mean ± σ):     22.884 s ±  0.265 s    [User: 20.518 s, System: 2.342 s]
  Range (min … max):   22.594 s … 23.596 s    16 runs
 
Benchmark 2: ninja -C build/gcc -j1 doctest_with_main [1af15d2]
  Time (mean ± σ):      2.781 s ±  0.247 s    [User: 2.539 s, System: 0.241 s]
  Range (min … max):    2.479 s …  3.261 s    16 runs
 
Summary
  'ninja -C build/gcc -j1 doctest_with_main [1af15d2]' ran
    8.23 ± 0.74 times faster than 'ninja -C build/gcc -j1 doctest_with_main [47ce36f]'
```

In our CMake configuration, we define `DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN` -- the in-source define is just to more adequately indicate that this source will actually provide some implementation.

Long-term, it would be good to provide at-least the following library forms for consumers to pick-and-choose from:

1. **The single-header form**: A dependency that provides _just_ `doctest.h` and nothing else
2. **The single-TU form**: A dependency that provides `doctest.h`, and the accompanying `doctest.cpp` for convenience
3. **The unassembled form**: A dependency that provides all of `parts/*`, primarily for internal testing

[...]

PR #1135, opened 2026-05-26
feat: Add CMake option to compile individual parts

Repiteo:
- Follow-up to #1123

## Description
At some point, the CMake setup was changed to target a single `doctest.cpp` instead of the individual parts, to help with speed. Now that speed is no longer a concern, we should at least make sure that those individual parts will indeed compile as expected.

## GitHub Issues
N/A


mitchgrout:
So my bad there; when we were chunking-up `doctest.h` into parts, we were adding them to the `doctest_with_main` target. This is a public-facing target that consumers may (and currently are) using. Since we had ~30 sources in there all looking at `prelude.h`, we ended up massively increasing the compile-time for such users (see #1102). To mitigate that at least temporarily, I swapped us back to a single-source in an albeit hacky way (see #1103), with the intent to eventually get back a from-parts build. That being said, even with `prelude.h` gone, we'll still likely see a pretty significant increase in compile-time compared to a unity build if we do by-parts.

I'll look over the change in detail soon, just wanted to drop this in for context.

Issue #1136, opened 2026-05-27
Support for fmtlib & std::format


jollie-fin:
### Description

Allowing use of std::formatter (or fmt::formatter) with default format, for printing.

This opens using doctest without having to duplicate logging function and do UB

This is related to !843, so it is a form of issue reopening, but this times it comes with a PR :)

( https://github.com/jollie-fin/doctest/tree/formatter )

### Context

We have a significant codebase relying on doctest for testing.

We are slowly switching from operator<< to std::formatter, and it would be lovely to be able to automatically log them on assertion with default formatting string.

It would avoid having to duplicate stringification just for doctest.

I do know about !843 , but since I have worked on a solution, I thought I would open a new issue.

PR #1137, opened 2026-05-27
feat(format): adds support for fmtlib & std::format

jollie-fin:
## Description
This allows for stringification through std::format or fmt::format. It is fully optional and doesn't impact regular stringification

Full support from C++11 -> C++23 is handled

## GitHub Issues
Resolves https://github.com/doctest/doctest/issues/1136


jollie-fin:
There are some issues with partially implemented C++20 & C++23 in old versions of gcc/clang that don't interact well with libfmt.

Will fix later

Issue #1159, opened 2026-05-30
MSVC -  Specializing standard library template std::tuple is forbidden by N5014 (use /Wv:18)


phil-zxx:
### Description

Since the last update of MSVC (May 2026, 18.6.2, 14.51.36231) doctest is throwing the below warning/error when compiling on Windows.

```
doctest.h(686): error C2220: the following warning is treated as an error
doctest.h(686): warning C5285: cannot declare a specialization for 'std::tuple': Specializing this
                standard library template is forbidden by N5014 [tuple.tuple.general]/1

doctest.h(686): note: see declaration of 'std::tuple'
doctest.h(686): note: to simplify migration, consider the temporary use of /Wv:18 flag with the version of
                the compiler with which you used to build without warnings
```

### Steps to reproduce

Compile a `cpp` file with the below content in VS 2026 on the latest May 2026 update:
```cpp
#define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
#include <doctest.h>

```

### doctest version

2.5.2

### Operating system

Windows

### Architecture

x64

### Compiler

MSVC  (18.6.2, 14.51.36231)

mitchgrout:
Sounds like we just need to add in an extra inline suppression. If you're happy to, have a crack at it and pop it up for review. Alternatively, you can consider using [`DOCTEST_CONFIG_USE_STD_HEADERS`](https://github.com/doctest/doctest/blob/master/doc/markdown/configuration.md#doctest_config_use_std_headers) to avoid the error entirely.

phil-zxx:
Sure thing. Let me know how this looks https://github.com/doctest/doctest/pull/1160

mitchgrout:
Fixed in v2.6

PR #1160, opened 2026-05-31
Suppress the "Specializing standard library template std::tuple is forbidden" warning

phil-zxx:
## Description

TBD

## GitHub Issues

#1159

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -683,8 +683,7 @@ template <class charT, class traits>
 class basic_istream;
 typedef basic_istream<char, char_traits<char>> istream; // NOLINT(modernize-use-using)
 template <class... Types>
-// Specializing template std::tuple is forbidden
-DOCTEST_MSVC_SUPPRESS_WARNING_WITH_PUSH(5285)
+DOCTEST_MSVC_SUPPRESS_WARNING_WITH_PUSH(5285) // Specializing template std::tuple is forbidden
 class tuple;
 DOCTEST_MSVC_SUPPRESS_WARNING_POP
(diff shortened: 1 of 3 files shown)


mitchgrout:
As this is your first PR please check out the CONTRIBUTING.md. Specifically:

- Pull requests should target `dev` -- I'll handle cherry picking back into v2.5.x
- Please use the pull request template that was specified on creation

Regarding the change itself, please check out `warnings.h` as this should provide the required macros to do warning suppression without having to manually `#ifdef` or `#pragma`.

I also assume you've tested this with your reprod and confirmed it works as expected?

phil-zxx:
Let me know if the change looks better now.

And yes, i tested it locally and it works. Both using my first change, and now also using your suggestion via `DOCTEST_MSVC_SUPPRESS_WARNING`.

mitchgrout:
> Let me know if the change looks better now.
> 
> And yes, i tested it locally and it works. Both using my first change, and now also using your suggestion via `DOCTEST_MSVC_SUPPRESS_WARNING`.

Not _exactly_ what I had in mind sorry. My comment was more that, `warnings.h` provides macros to do the suppression without having to explicitly run a `#pragma`. Your change has added it to the global pool of suppressed warnings which, while it works, is maybe a bit heavy-handed.

See https://github.com/doctest/doctest/blob/master/doctest/parts/public/warnings.h#L38-L48 for reference

phil-zxx:
@mitchgrout Hoping this looks better now

mitchgrout:
Thanks for the change!

PR #1161, opened 2026-06-01
cmake: drop deprecated `-Wstrict-overflow=1` `gcc` flag

trofi:
## Description

`gcc-17` formally deprecated `-Wstrict-overflow` options in https://gcc.gnu.org/git/?p=gcc.git;a=commitdiff;h=b31a43560e88600d412318f75a43ac8e562ba8c5. Informally the option was depracated since `gcc-8`.

I only noticed it because `gcc-17` deprecated it slightly incorrectly and started to fail `doctest` build: https://gcc.gnu.org/PR125558

While it's a `gcc` bug `gcc` devs suggested removing use of the flag. This change does the removal.

### DIFF
--- a/scripts/cmake/common.cmake
+++ b/scripts/cmake/common.cmake
@@ -109,7 +109,6 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
     add_compiler_flags(-Wundef)
     add_compiler_flags(-Wredundant-decls)
     add_compiler_flags(-Wshadow)
-    add_compiler_flags(-Wstrict-overflow=5)
     add_compiler_flags(-Wwrite-strings)
     add_compiler_flags(-Wpointer-arith)
     add_compiler_flags(-Wcast-qual)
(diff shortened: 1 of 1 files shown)


mitchgrout:
Oh, gcc-17 is available? I swear it was only just recently that I added gcc-16...

Change is chill, although I might recommend that we swap to optionally enabling if the version is `<= 16`, as it preserves the warning for older compilers which IMO is still valuable. I'll spin up a PR later today to add gcc-17 into our build matrix which should demo the breakage pretty clearly

trofi:
> Oh, gcc-17 is available? I swear it was only just recently that I added gcc-16...

It's a `master` branch of `gcc`. No releases yet. It's planned around Apr/May of 2027: https://gcc.gnu.org/develop.html

> Change is chill, although I might recommend that we swap to optionally enabling if the version is `<= 16`, as it preserves the warning for older compilers which IMO is still valuable. I'll spin up a PR later today to add gcc-17 into our build matrix which should demo the breakage pretty clearly

Added a `if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 17.0)` guard.

mitchgrout:
Any preference for backporting to v2.5.x?

trofi:
If it gets released one way or another within a few month without backporting then it would be perfect for me. If you expect it to take about a year without it, then backporting sounds reasonable.

mitchgrout:
> If it gets released one way or another within a few month without backporting then it would be perfect for me. If you expect it to take about a year without it, then backporting sounds reasonable.

Haven't had as much time as I hoped, so I've backported to v2.5.3 and cut that release just now

PR #1170, opened 2026-06-21
fix: Allow doctest::String to be constructed from null values

### LINKED ISSUE #1045: SIGSEGV when handling `CHECK_NOTHROW(throw nullptr)`
The following code triggers a segmentation fault:

```cpp
TEST_CASE("...") {
  CHECK_NOTHROW(throw nullptr);
}
```

mitchgrout:
## Description

Update the `String(const char *)` constructor to accept `nullptr` values. The `String(const char *, size_type)` constructor already passively handles the case of `(nullptr, 0)`, but the `String(const char *)` variant unconditionally passes the input to `std::strlen`; this is not safe, since `strlen` is undefined on null inputs. 

This manifested in the problem of:

```cpp
CHECK_NOTHROW(throw nullptr);
```

...resulting in a SIGSEGV, since...

1. The `translateActiveException` handler specialized for `const char *` inputs, which `nullptr` would convert to, which
2. Was implicitly converted into a `doctest::String` via the `String(const char *)` converting-constructor, which
3. Provided the value to `std::strlen`

This fix essentially just adds a conditional check to the constructor to yield 0 in such cases. 

## GitHub Issues

Fixes #1045

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -8209,7 +8209,7 @@ String::~String() {
 } // NOLINT(clang-analyzer-cplusplus.NewDeleteLeaks)
 
 String::String(const char *in)
-    : String(in, std::strlen(in)) {}
+    : String(in, in? std::strlen(in) : 0) {}
 
 String::String(const char *in, size_type in_size) {
     memcpy(allocate(in_size), in, in_size);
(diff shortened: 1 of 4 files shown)


mitchgrout:
Maybe better to avoid `String(nullptr)` altogether. Better parity with `std::string` which forbids constructing from `null`. In which case, the best fix is probably to update `translateCurrentException` itself. Could also consider just saying, "don't throw `nullptr`" but best to fail gracefully IMO

mitchgrout:
Closed in lieu of #1173

Issue #76, opened 2017-06-24
testing private members - ability to write test cases in class bodies


r-barnes:
## Description
Some classes have private methods which cannot be conveniently tested through a public interface.

What is the recommended way to test a private method using doctest?

### Extra information
* doctest version: **v1.2.1**
* Operating System: **Lubuntu 17.04 (Zesty)**
* Compiler+version: **G++ 6.3.0**

onqtam:
```
#define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
#include "doctest/doctest/doctest.h"

class A {
    void private_method() { printf("hello!\n"); }
    
#ifdef DOCTEST_LIBRARY_INCLUDED
public:
    static void test1() {
        A a;
        a.private_method();
        CHECK(1 == 2);
    }
#endif
};

DOCTEST_REGISTER_FUNCTION(A::test1, "test name")

// or like this:
// TEST_CASE("test name") { A::test1(); }
```

onqtam:
I'm closing this - not sure what else I could suggest

JankoDedic:
Could this be solved using C++17 `inline` variables? A global dummy variable could be defined inside of the class as `inline static`.

onqtam:
@JankoDedic  turns out it is possible! there was one other problem but I managed to solve it, so now I'm thinking of what to call the new macro - perhaps ```TEST_CASE_CLASS``` ...?

JankoDedic:
That's amazing to hear! I just made it work a few minutes ago as well :). Not sure about the name, but `TEST_CASE_CLASS` is good with me :)

onqtam:
I just pushed the change in the dev branch - here is how to use it:

```c++
class my_type {
    int data = 5;

    TEST_CASE_CLASS("private test! no problem") {
        my_type local;
        CHECK(local.data == 2);
    }

public:
    my_type() = default;

    TEST_CASE_CLASS("public test! doesn't matter") {
        my_type local;
        CHECK(local.data == 2);
    }
};
```

u3shit:
Unfortunately this TEST_CASE_CLASS macro requires writing the test case inside the class body, which might be ok when you have a header only library, but it's not that good when you have cpp/hpp files, and you would have to include a bunch of heavy includes in the hpp just for the test case. But I think it should be possible to split it into two, something like this (POC code follows, this won't work with templates):
```c++
#define DOCTEST_TEST_CASE_CLASS_DECLARE(name) \
[...]

PR #1191, opened 2026-07-07
Fix for https://github.com/doctest/doctest/issues/608

datadiode:
Proposal from https://github.com/doctest/doctest/issues/76#issuecomment-4885959921

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -573,17 +573,15 @@ DOCTEST_SUPPRESS_PUBLIC_WARNINGS_PUSH
 
 namespace doctest {
 namespace detail {
-DOCTEST_CLANG_SUPPRESS_WARNING_WITH_PUSH("-Wunused-function")
-static DOCTEST_CONSTEXPR int consume(const int *, int) noexcept {
-    return 0;
-}
-DOCTEST_CLANG_SUPPRESS_WARNING_POP
+struct consumer {
+    consumer(int) DOCTEST_NOEXCEPT {}
+};
 } // namespace detail
 } // namespace doctest
 
 #define DOCTEST_GLOBAL_NO_WARNINGS(var, ...)                                                                           \
     DOCTEST_CLANG_SUPPRESS_WARNING_WITH_PUSH("-Wglobal-constructors")                                                  \
-    static const int var = doctest::detail::consume(&var, __VA_ARGS__);                                                \
+    static const doctest::detail::consumer var = __VA_ARGS__;                                                          \
     DOCTEST_CLANG_SUPPRESS_WARNING_POP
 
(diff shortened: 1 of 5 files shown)


mitchgrout:
A few preliminary comments:

- Please use the merge-request template; just helps to capture what the intent of the change is
- Since the source issue/discussion has no info regarding this, what is the actual reproduction of this problem?
- Can you produce a test which shows the problem (plus this fixing it?) -- see `tests/regression`
- `consumer::consumer` should probably be `inline`

I'm not 100% convinced that this construct is going to be free of unused variable warnings. However, I can't find any combination within gcc / clang that complains about it, so perhaps we're just good there?

datadiode:
That was an instructive journey.  
Apologies for also fixing the nullptr derefs on `TestSuite::m_test_suite` which prevented some lights from turning green.  
Declaring `consumer::consumer()` as `inline` is allowed but doesn't make the code more valid.

doctest/parts/public/test_suite.h
```diff
@@ -11,7 +11,7 @@ namespace doctest {
 namespace detail {
 
 struct DOCTEST_INTERFACE TestSuite {
-    const char *m_test_suite = nullptr;
+    const char *m_test_suite = "";
```
mitchgrout: You mentioned this in your prior comment; what's the actual issue being resolved here?

mitchgrout (commented):
Change itself is fine, but I would appreciate a bit more context regarding the change. At least:

- What the error you're resolving is (presumably `-Wunused-variable` or similar)
- What compiler/version/platform(s)

Alt. you can create an issue separate from #76 to capture this

doctest/parts/public/test_suite.h
datadiode: This fixes the nullptr derefs on `TestSuite::m_test_suite` which prevented some lights from turning green.  
When the nullptr derefs happened, the call stack was at `Context::run()` &rarr; `matchesAny()` &rarr; `wildcmp()`.

datadiode:
> Change itself is fine, but I would appreciate a bit more context regarding the change. At least:
> 
>     * What the error you're resolving is (presumably `-Wunused-variable` or similar)
> 
>     * What compiler/version/platform(s)
> 
> 
> Alt. you can create an issue separate from #76 to capture this

That separate issue exists as #608. Honestly, I've nothing to add.

datadiode:
So `inline static` initialization does not take part in the compilation unit's static initialization order, which means it can execute before the code snippet that, as the comment says, _is here to clear the 'current test suite' for the current translation unit - at the top_.

datadiode:
Closing in favor of https://github.com/doctest/doctest/pull/1193.

PR #1193, opened 2026-07-28
Make test functions less anonymous

datadiode:
#### What
Enables `TEST_CASE()` and `TEST_CASE_TEMPLATE_DEFINE()` to access class internals by allowing classes to friend the underlying test functions, which are therefore created as template functions with a known name but a varying template parameter to make them unique throughout the compilation unit.

With these enhancements, `TEST_CASE_CLASS()` basically loses its point and should therefore be removed rather than fixed.

[This example](https://github.com/datadiode/ModernCppStarter/commit/6803ba7aa7257458861e9d0fe4973f700179b7ef) shows how the feature would integrate into existing code.

The `TEST_SUITE_USING()` thing is off topic hence should arguably go into a separate PR.

#### Why
For fun :smile:

## GitHub Issues
Related to #608, #76.


mitchgrout:
Your PR description and implementation are a bit misaligned, so I'm going to quickly note what my understanding of your change is. You're requiring that fixtures introduce a `template <typename _> void test_case()` which acts as a template-specialization point, by which we can have:

```cpp
class derived : public base { };

template <>
void base::test_case<derived>() { ... }
```

And, by virtue of the `test_case` method being defined on the base class, allows the specialization to peer into the private members of the fixture.

---

While the idea is novel, the assumption you've made is that the fixture is always user-controlled; this is never stated or implied in the docs. I could legally have something like...

```cpp
TEST_CASE_FIXTURE(std::ostringstream, "Writing data") {
  *this << "value";
  CHECK(str() == "value");
}
```

...which given that one of `doctest`'s selling points is its integrability into your source files, is more likely than you might think. This change would break setups like this.

datadiode:
Yes my proposal changes `TEST_CASE_FIXTURE` in a breaking way.

As for the inheritance, `class derived : public base`, it no longer happens but is rather replaced by the `template <int>` method, because the inheritance would again block access to internals of classes which `friend` the text fixture to give the test access to their internals.

Test fixtures which are beyond of the control of the one who authors the test cases seem quite an anomaly.
Indeed I don't think it happens in the wild.

Anyways, I am happy with reverting the changes around `TEST_CASE_FIXTURE`. Yet better, remove it altogether.  
https://github.com/snick-a-doo/IBM650 uses doctest and creates its text fixtures from within plain `TEST_CASE` bodies, cleanly, without magic.

mitchgrout:
> Test fixtures which are beyond of the control of the one who authors the test cases seem quite an anomaly. Indeed I don't think it happens in the wild.

Was more meaning that the type may not be owned by the test itself, i.e. might be a class from your public interface. Let's say I'm writing something like...

```cpp
class lib::int_parser {
public:
  int parse(std::string);

private:
  int parse_binary(std::string);
  int parse_octal(std::string);
  int parse_decimal(std::string);
  int parse_hex(std::string);
};
```

It's not unreasonable to assume that I might end up with a test like...

```cpp
TEST_CASE_FIXTURE(lib::int_parser, "Parsing decimal values") {
  CHECK(parse("10") == 10);
}
```

...since after all, my class has a default constructor, and it avoids the need to litter the test with `parser.<...>` where its otherwise clear from the context. Having to add in a `template <...> void test_case()` to my class, in an instance where it works fine without it, feels odd.

I also suspect that your approach would not work in the case of...

```cpp
template <typename T>
class object {
public:
  template <int>
  void test_case();
};
```

...since attempting to have:

```cpp
TEST_CASE_FIXTURE(object<float>, "...")
```

...would require a specialization of the form:

```cpp
template <>
void object<float>::test_case<LINE>()
```

...which is not well-formed afaik. Currently, this pattern would be perfectly fine under the current implementation.

---

More generally, the public consensus is that testing private code is bad idea:

- https://google.github.io/googletest/advanced.html#testing-private-code
- https://stackoverflow.com/questions/105007/should-i-test-private-methods-or-only-public-ones
- https://www.reddit.com/r/cpp_questions/comments/r8qkfu/best_way_to_test_private_memberfunctions/

...and further, is not supported by Catch2, Boost.Test, CppUTest, ...; the only mainstream one appears to be gtest, which again advises against it where possible.

datadiode:
So I just learned that the well-formed way would be like so:
```C++
template <>
template <>
void object<float>::test_case<__LINE__>();
```

I have no strong incentive to try fix the situation with `TEST_CASE_FIXTURE`, hence will revert that change.

There exists truth beyond the public consensus.  
https://www.reddit.com/r/cpp_questions/comments/t19hw8/comment/hyh99z1/ makes a point in this regard.

mitchgrout:
Thought I'd responded to this earlier, sorry.

My personal view is that testing is almost-always sufficiently performed via the public interface. If a private aspect of a structure is complex enough to mandate testing directly, then we should probably be asking ourselves firstly why such a complex object needs to exist in the first place, and whether it can be appropriately decomposed. The decomposition then naturally provides a means for testing the once-private behaviour.

Anyway, I'm not adverse to having _opt-in_ support for private testing like you have here, but: 

- It must not come at the detriment to compile-time and run-time performance for existing tests, and
- It must not require any changes to existing tests

I'll try to benchmark this change later on against a particularly large test-suite that I have to see if there's any degradation in compile/run-time performance.

PR #1196, opened 2026-08-13
Pass --no-colors=true during doctest_discover_tests discovery

jameswatsoncraft:
## Description
Pass `--no-colors=true` during `doctest_discover_tests` discovery so discovery output stays stable when color detection would otherwise flip.

## Type
- [x] Bugfix

## Notes
Some boilerplate was tool-assisted; I checked the CMake discovery path before opening.

## Checklist
- [x] Targets `dev`

PR #1198, opened 2026-08-14
feat(assert): add CHECK_THROWS_WITH_PREDICATE for exception inspection (#846)

### LINKED ISSUE #846: Feature request: predicate on CHECK_THROWS
## Description

The current assertion macros can only assert on either exception type or exception message. However, in my use case, I need to assert that it has a specific exception code.

Can we have

* Either an assertion macro that returns the captured exception
* Or an assertion macro accepting a lambda as a predicate to further distinguish an expected exception from an unexpected one

jameswatsoncraft:
## Description
Add `CHECK_THROWS_WITH_PREDICATE` (and WARN/REQUIRE variants) so thrown exceptions can be inspected with a predicate.

## GitHub Issue
Closes #846

## Type
- [x] Feature addition

## Notes
Tooling helped with scaffolding; macro behavior and coverage are intentional.

## Checklist
- [x] Targets `dev`

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -1516,6 +1516,11 @@ struct ContextOptions {
 
 DOCTEST_INTERFACE const ContextOptions *getContextOptions();
 
+/// Returns the active test case name, or an empty string outside of a test.
+DOCTEST_INTERFACE const char *current_test_name();
+/// Returns the innermost active subcase name, or an empty string outside of a subcase.
+DOCTEST_INTERFACE const char *current_subcase_name();
+
 } // namespace doctest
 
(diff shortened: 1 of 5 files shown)


mitchgrout (changes requested):

- `DT_WARN_THROWS_WITH_PREDICATE` uses `is_throws_as` which is incorrect
- Macro directly sets `.m_failed` and `.m_decomp` which is not required
- Example is insufficient to demonstrate behaviour fully

Change also clearly contains content from #1195 / #1196.

I've got an implementation of this change cooked up already; just building the project integration test suite before introducing it so that adequate testing can be performed.

PR #1199, opened 2026-08-14
feat(approx): add Catch2-compatible margin() absolute tolerance (#389)

### LINKED ISSUE #389: Approx Epsilon Usage & Constructors
Hi,

As I often deal with floating point values, I tend to use the `Approx` class quite often. Now, let's consider the below example. The `CHECK` will pass even though the true absolute difference is `0.00001 = 1.e-05`. If we use `Approx(testValue).epsilon(1.e-12)` then the test will give `CHECK(false)`.
```cpp
const double testValue = 0.98765;
const double trueValue = 0.98766;
CHECK(Approx(testValue) == trueValue);  // this gives CHECK(true)
```
Usually, I deem values to be correct if they pass a certain `absolute` threshold:
```cpp
abs(lhs - rhs) < epsilon;                                                   // (1)
```
However, `doctest` implements this formula:
```cpp
abs(lhs - rhs) < epsilon * ( rhs.m_scale + max(abs(lhs), abs(rhs)) );       // (2)
```
which is slightly different.

---

Questions:

1. Why does `epsilon` default to `std::numeric_limits<float>::epsilon()*100` = `1.2e-05` as opposed to `std::numeric_limits<double>::epsilon()*100` = `2.2e-14`. I am asking since all comparisons are based on `double` values, and not `float` values. Personally, I find `1.2e-05` to be quite a low threshold.<br>

2. In `catch2` one can write `Approx(testValue).margin(1.e-12).epsilon(0)` to exactly achieve method `(1)` above.
(a) Is there benefit in adding extra settings to doctest's `Approx` class? So that one can achieve both method `(1)` and `(2)`?
(b) In order to avoid these long factory-like expressions (as `catch2` has them), one could allow to add them via an additional constructor. E.g. `Approx(double value, double epsilon, double margin)` would help shorten the expression to `Approx(testValue, 1.e-12, 0)`.

Happy to help write the code for this, but wanted to get some feedback on my questions first.

jameswatsoncraft:
## Description
Add Catch2-style `Approx::margin()` for absolute floating-point tolerance.

## GitHub Issue
Closes #389

## Type
- [x] Feature addition

## Notes
Tooling helped with scaffolding; comparison behavior matches the issue ask.

## Checklist
- [x] Targets `dev`
- [x] Parts + reassembled header

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -1359,6 +1359,16 @@ struct DOCTEST_INTERFACE Approx {
     }
 #endif //  DOCTEST_CONFIG_INCLUDE_TYPE_TRAITS
 
+    Approx &margin(double newMargin);
+
+#ifdef DOCTEST_CONFIG_INCLUDE_TYPE_TRAITS
+    template <typename T>
+    typename std::enable_if<std::is_constructible<double, T>::value, Approx &>::type margin(const T &newMargin) {
+        m_margin = static_cast<double>(newMargin);
+        return *this;
+    }
+#endif // DOCTEST_CONFIG_INCLUDE_TYPE_TRAITS
+
     Approx &scale(double newScale);
 
 #ifdef DOCTEST_CONFIG_INCLUDE_TYPE_TRAITS
@@ -1407,6 +1417,7 @@ struct DOCTEST_INTERFACE Approx {
 #endif // DOCTEST_CONFIG_INCLUDE_TYPE_TRAITS
 
     double m_epsilon;
+    double m_margin;
     double m_scale;
     double m_value;
 };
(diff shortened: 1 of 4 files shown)


mitchgrout (changes requested):

- `marginComparison` cannot be defined in an anonymous namespace as after assembly, will generate an anonymous namespace in a header file.
- `marginComparison(...) || <original...>` is suspicious as setting `.margin(...)` should override tolerance-based comparisons; `||` does not guarantee this.
- Lacks any unit-tests to demonstrate behaviour
- Lacks any examples to demonstrate behaviour

Change also clearly contains content from https://github.com/doctest/doctest/pull/1195

PR #1200, opened 2026-08-14
feat: add --print-test to print names as tests complete (#75)

### LINKED ISSUE #75: Print test names as they complete
## Description

Doctest currently has the following feature:

     -s,   --success=<bool>                include successful assertions in output

However, using this results in millions of prints, since I use stochastic testing in some cases.

It would be doctest if it had an additional feature:

     -pt,   --print_test=<bool>       Print test names to output as they are run

Currently, the closest you can get is using

     -d,   --duration=<bool>               prints the time duration of each test

Which prints the test name and the time it took to run. This is essentially what I want, but figuring out that the `-d` command is what you need to run to print test names is perhaps a little counter-intuitive. So maybe this can be resolved via a simple change to the help file.

### Extra information
* doctest version: **v1.2.1**
* Operating System: **Lubuntu 17.04**
* Compiler+version: **GCC 6.3.0**

jameswatsoncraft:
﻿## Summary
Adds `--print-test` / `-pt` to print each test name as it completes without dumping every successful assertion (#75).

`-s/--success` is too noisy for stochastic suites, and `-d/--duration` was the closest workaround but always includes timing. This flag prints only the test name on completion.

## Changes
- New `ContextOptions::print_test`
- CLI: `-pt` / `--print-test=<bool>`
- Console reporter prints `tc->m_name` when enabled (duration still takes precedence if both are set)
- Help text updated

Closes #75

### DIFF
--- a/doctest/doctest.h
+++ b/doctest/doctest.h
@@ -1359,6 +1359,16 @@ struct DOCTEST_INTERFACE Approx {
     }
 #endif //  DOCTEST_CONFIG_INCLUDE_TYPE_TRAITS
 
+    Approx &margin(double newMargin);
+
+#ifdef DOCTEST_CONFIG_INCLUDE_TYPE_TRAITS
+    template <typename T>
+    typename std::enable_if<std::is_constructible<double, T>::value, Approx &>::type margin(const T &newMargin) {
+        m_margin = static_cast<double>(newMargin);
+        return *this;
+    }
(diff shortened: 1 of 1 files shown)


mitchgrout (changes requested):

Misinterpretation of source issue. Initial request was for a flag, discussion leant towards implementing a reporter. 

Change also clearly contains content from #1195 / #1199 and has directly modified the `doctest.h` header.

PR #1204, opened 2026-08-19
Fix counter overflow

bigerl:
note: The changes to the `doctest.hpp` come from `assemble.py`. 

## Description

We are using doctest at Tentris for internal projects that recently exceeded the int (32-bit) used for counting assertions. I replaced it with 
```
using counter_type = unsigned long long;
```

I was a bit surprised that no one else encountered this so far. At least, I couldn't find any related issues or PRs.

It comes with (claude generated) regression tests.


bigerl:
@mitchgrout Checking in. Have you decided what you want to do with this PR?

Issue #205, opened 2019-03-21
Memory leaks just by including doctest.h


edmundvermeulen:
## Description

We upgraded from doctest 1.2.6 to 2.2.3. Now Visual Studio reports memory leaks.

### Steps to reproduce

The following simple example code gives 3 memory leaks when run.

```cpp
#include "stdafx.h"

#define DOCTEST_CONFIG_IMPLEMENT
#include "doctest.h"

int main()
{
    int flag = _CrtSetDbgFlag(_CRTDBG_REPORT_FLAG);
    _CrtSetDbgFlag(flag | _CRTDBG_ALLOC_MEM_DF | _CRTDBG_LEAK_CHECK_DF);

    return 0;
}
```

```
Detected memory leaks!
Dumping objects ->
{175} normal block at 0x004F5620, 8 bytes long.
 Data: <  O     > EC 98 4F 00 00 00 00 00 
{163} normal block at 0x004F5578, 8 bytes long.
 Data: <| O     > 7C 8E 4F 00 00 00 00 00 
{156} normal block at 0x004F57E0, 8 bytes long.
 Data: < 5O     > AC 35 4F 00 00 00 00 00 
Object dump complete.
```

### Extra information
* doctest version: 2.2.3
* Operating System: Windows 10
* Compiler+version: Microsoft Visual Studio 2015

onqtam:
strange... valgrind on Unix doesn't report anything.

Version 2.0+ uses a lot more things from the STL - this could be a bug in the toolchain (i've had such issues many times before where a specific version of the compiler is giving trouble).

I'll try to reproduce it and investigate but I don't know when I'll get to this.

edmundvermeulen:
The same leaks occur in VS2019 Preview 2.

Further investigation shows that this is caused by g_infoContexts being thread_local. When DOCTEST_THREAD_LOCAL is defined to nothing then the leaks go away. Is it necessary to have it be thread_local if all the tests run from the same thread?

onqtam:
All tests always run from the same thread but if one test case spawns multiple threads and uses asserts and logging with INFO() - then that needs to be thread-local. If you don't have such a scenario you may indeed use DOCTEST_THREAD_LOCAL.

Thanks for diagnosing this - I'm very curious as to why a thread-local global std::vector is a problem. If you have any ideas I would be grateful - otherwise I'll look into it when I have the chance... But for now version 2.3 with xml reporting is the priority.

edmundvermeulen:
[...]

PR #1206, opened 2026-09-01
Fix apparent leaks with mfc

otrempe:
## Description
Refer to https://github.com/doctest/doctest/issues/205#issuecomment-5479310995 for the why of this pull request.

This pull request removes some static allocations and disable memory tracking for the ones that were not removed.

A test has been added to validate no apparent leaks are detected.
The test fails as expected when run on the dev branch without this PR's fixes

## GitHub Issues
https://github.com/doctest/doctest/issues/205

### DIFF
--- a/doctest/parts/private/context_scope.cpp
+++ b/doctest/parts/private/context_scope.cpp
@@ -12,10 +12,16 @@ DOCTEST_DEFINE_INTERFACE(IContextScope)
 
 #ifndef DOCTEST_CONFIG_DISABLE
 namespace detail {
-DOCTEST_THREAD_LOCAL std::vector<IContextScope *> g_infoContexts; // for logging with INFO()
+DOCTEST_THREAD_LOCAL std::vector<IContextScope *> *g_infoContexts = nullptr;
+
+std::vector<IContextScope *> &getInfoContexts() {
+    if (g_infoContexts == nullptr)
+        g_infoContexts = new std::vector<IContextScope *>;
+    return *g_infoContexts;
+}
 
 ContextScopeBase::ContextScopeBase() {
-    g_infoContexts.push_back(this);
+    getInfoContexts().push_back(this);
 }
 
 ContextScopeBase::ContextScopeBase(ContextScopeBase &&other) noexcept {
@@ -23,7 +29,7 @@ ContextScopeBase::ContextScopeBase(ContextScopeBase &&other) noexcept {
         other.destroy();
     }
     other.need_to_destroy = false;
-    g_infoContexts.push_back(this);
+    getInfoContexts().push_back(this);
 }
 
 // destroy cannot be inlined into the destructor because that would mean calling stringify after
@@ -35,7 +41,7 @@ void ContextScopeBase::destroy() {
         this->stringify(&s);
         g_cs->stringifiedContexts.emplace_back(s.str().c_str());
     }
-    g_infoContexts.pop_back();
+    getInfoContexts().pop_back();
 }
 
 } // namespace detail
--- a/doctest/parts/private/context_scope.cpp
+++ b/doctest/parts/private/context_scope.cpp
@@ -2,6 +2,7 @@
 #include "doctest/parts/private/context_state.h"
 #include "doctest/parts/private/exceptions.h"
 
+#include <memory>
 #include <sstream>
 
 DOCTEST_SUPPRESS_PRIVATE_WARNINGS_PUSH
@@ -12,12 +13,16 @@ DOCTEST_DEFINE_INTERFACE(IContextScope)
 
 #ifndef DOCTEST_CONFIG_DISABLE
 namespace detail {
-DOCTEST_THREAD_LOCAL std::vector<IContextScope *> *g_infoContexts = nullptr;
+DOCTEST_THREAD_LOCAL std::unique_ptr<std::vector<IContextScope *>> g_infoContexts;
 
 std::vector<IContextScope *> &getInfoContexts() {
     if (g_infoContexts == nullptr)
-        g_infoContexts = new std::vector<IContextScope *>;
+        g_infoContexts.reset(new std::vector<IContextScope *>);
     return *g_infoContexts;
+}
+
+void cleanupInfoContexts() {
+    g_infoContexts.reset();
 }
 
 ContextScopeBase::ContextScopeBase() {
--- a/doctest/parts/private/static_registration.h
+++ b/doctest/parts/private/static_registration.h
@@ -1,0 +1,46 @@
+#ifndef DOCTEST_PARTS_PRIVATE_STATIC_REGISTRATION
+#define DOCTEST_PARTS_PRIVATE_STATIC_REGISTRATION
+
+#include "doctest/parts/public/config.h"
+
+// Registrations are intentionally performed during static initialization. In a
+// debug-CRT build these objects can still be alive when a host DLL asks the CRT
+// for a leak dump during its detach notification. Mark only allocations made
+// while registering doctest metadata as ignored, so that such a snapshot does
+// not mistake framework lifetime state for a user allocation leak.
+#if DOCTEST_MSVC && defined(_DEBUG)
+#include <crtdbg.h>
+#endif
+
+DOCTEST_SUPPRESS_PRIVATE_WARNINGS_PUSH
+
+namespace doctest {
+namespace detail {
+
+class StaticRegistrationMemoryGuard {
+public:
+    StaticRegistrationMemoryGuard() noexcept {
+#if DOCTEST_MSVC && defined(_DEBUG)
+        m_flags = _CrtSetDbgFlag(_CRTDBG_REPORT_FLAG);
+        _CrtSetDbgFlag(m_flags & ~_CRTDBG_ALLOC_MEM_DF);
+#endif
+    }
+
+    ~StaticRegistrationMemoryGuard() {
+#if DOCTEST_MSVC && defined(_DEBUG)
+        _CrtSetDbgFlag(m_flags);
+#endif
+    }
+
+private:
+#if DOCTEST_MSVC && defined(_DEBUG)
+    int m_flags;
+#endif
+};
+
+} // namespace detail
+} // namespace doctest
+
+DOCTEST_SUPPRESS_PRIVATE_WARNINGS_POP
+
+#endif // DOCTEST_PARTS_PRIVATE_STATIC_REGISTRATION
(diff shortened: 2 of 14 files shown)


doctest/parts/private/context_scope.cpp
```diff
@@ -12,18 +13,28 @@ DOCTEST_DEFINE_INTERFACE(IContextScope)
 
 #ifndef DOCTEST_CONFIG_DISABLE
 namespace detail {
-DOCTEST_THREAD_LOCAL std::vector<IContextScope *> g_infoContexts; // for logging with INFO()
+DOCTEST_THREAD_LOCAL std::unique_ptr<std::vector<IContextScope *>> g_infoContexts;
```
mitchgrout, line 16: Not sure I follow why this needs to be a `unique_ptr<vector<IContextScope *>>`. Why can't we just `.clear()` the vector on `cleanupInfoContexts()`? Surely Windows is just complaining about the fact that a thread-local vector may-or-may-not have a `.data()` field, even if it's cleared (due to the space being reserved). 

Saves the effort of having to maybe-initialize a `unique_ptr`.

doctest/parts/private/static_registration.h
```diff
@@ -0,0 +1,46 @@
+#ifndef DOCTEST_PARTS_PRIVATE_STATIC_REGISTRATION
+#define DOCTEST_PARTS_PRIVATE_STATIC_REGISTRATION
+
+#include "doctest/parts/public/config.h"
+
+// Registrations are intentionally performed during static initialization. In a
+// debug-CRT build these objects can still be alive when a host DLL asks the CRT
+// for a leak dump during its detach notification. Mark only allocations made
+// while registering doctest metadata as ignored, so that such a snapshot does
+// not mistake framework lifetime state for a user allocation leak.
```
mitchgrout, line 10: Would be more appropriately associated with the `StaticRegistrationMemoryGuard` class itself

doctest/parts/private/static_registration.h
```diff
@@ -0,0 +1,46 @@
...
+#if DOCTEST_MSVC && defined(_DEBUG)
+        _CrtSetDbgFlag(m_flags);
+#endif
+    }
+
+private:
+#if DOCTEST_MSVC && defined(_DEBUG)
+    int m_flags;
+#endif
+};
```
mitchgrout, line 39: Similarly more clearly expressed as...

```cpp
#if DOCTEST_MSVC && defined(_DEBUG)
class StaticRegistrationMemoryGuard {
  ...
};
#else
class StaticRegistrationMemoryGuard {
  // Intentionally empty
};
#endif
```

...since otherwise you're slinging around the same `#if` in 3 places.

doctest/parts/private/static_registration.h
```diff
@@ -0,0 +1,46 @@
...
+
+namespace doctest {
+namespace detail {
+
+class StaticRegistrationMemoryGuard {
+public:
+    StaticRegistrationMemoryGuard() noexcept {
+#if DOCTEST_MSVC && defined(_DEBUG)
+        m_flags = _CrtSetDbgFlag(_CRTDBG_REPORT_FLAG);
+        _CrtSetDbgFlag(m_flags & ~_CRTDBG_ALLOC_MEM_DF);
```
mitchgrout, line 25: So just for my understanding, the intent is:

- The constructor temporarily disables memory tracking
- The destructor restores it
- Whilst disabled, any newly constructed objects aren't tracked(?)

...which means that if we do:

```cpp
StaticRegistrationMemoryGuard guard;
static int *immortal = new int;
```

...won't be reported as a leak, correct?

tests/regression/test.205.cpp
mitchgrout, line 1: In the cmake config, just conditonally-define this test based off whether we're running MSVC and debug. Cleans up the test code immensely.

tests/regression/test.205.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include <doctest/doctest.h>
+
+#include <thread>
+
+#if defined(_MSC_VER) && defined(_DEBUG)
+#include <crtdbg.h>
+#include <doctest/parts/private/reporters/debug_output_window.h>
+#endif
+
+TEST_CASE("doctest test run owns its temporary memory") {
```
mitchgrout, line 11: More loosely just, "doctest does not leak any memory from CRTs perspective", since we are most definitely still maintaining a bunch of non-temporary memory (e.g. reporters)

tests/regression/test.205.cpp
```diff
@@ -0,0 +1,48 @@
...
+    }};
+    worker.join();
+
+    CHECK(true);
+}
+
+int main(int argc, char **argv) {
+#if defined(_MSC_VER) && defined(_DEBUG)
+    {
+        // This used to own a thread-local stream for the process lifetime.
```
mitchgrout, line 28: Meaningless comment?

tests/regression/test.205.cpp
```diff
@@ -0,0 +1,48 @@
...
+        doctest::ContextOptions options{};
+        doctest::detail::DebugOutputWindowReporter reporter{options};
```
mitchgrout, line 30: Not really sure what you're trying to achieve with this block

tests/regression/test.205.cpp
```diff
@@ -0,0 +1,48 @@
...
+        doctest::Context context{argc, argv};
+        context.setOption("quiet", true);
+        context.setOption("no-debug-output", true);
+        result = context.run();
+    }
+
+#if defined(_MSC_VER) && defined(_DEBUG)
+    if (_CrtDumpMemoryLeaks())
+        return EXIT_FAILURE;
+#endif
```
mitchgrout, line 45: From the reprod in !1206, don't we have to have an accompanying `_CrtSetDbgFlag(...)` to actually begin tracking stuff? Is `_CrtDumpMemoryLeaks()` here actually going to do anything in the absence of that prior code?

tests/CMakeLists.txt
```diff
@@ -133,6 +133,13 @@ doctest_add_unit_test(WITH_MAIN SOURCE regression/test.1045.cpp EXCEPT ${FNOEXCE
 doctest_add_unit_test(WITH_MAIN SOURCE regression/test.1165.cpp EXCEPT ${FNOEXCEPT})
 doctest_add_unit_test(WITH_MAIN SOURCE regression/test.1171.cpp)
 
+# The debug CRT reports allocations which are still alive before static
+# destructors run. This regression test verifies that doctest does not leave
+# apparent leaks in that situation.
```
mitchgrout, line 138: Unnecessary comment

tests/CMakeLists.txt
```diff
@@ -133,6 +133,13 @@ doctest_add_unit_test(WITH_MAIN SOURCE regression/test.1045.cpp EXCEPT ${FNOEXCE
...
+if(MSVC)
+    doctest_add_unit_test(SOURCE regression/test.205.cpp)
+endif()
```
mitchgrout, line 141: `doctest_add_unit_test(... IF ${MSVC})` would be the pattern in this file. SEe other comments about also rolling in checks for _DEBUG. Honestly, you could probably just leave that bit out and demo that we don't get any reports in any of the build profiles we try, not just _DEBUG.

mitchgrout (changes requested):

doctest/parts/private/context_scope.cpp
otrempe: First, thanks for the quick review!

You guessed right. This is because when an element in added to this vector, the vector allocates capacity. With a plain vector, when cleared, the contained elements are removed, but the capacity allocation is still there and is seen as an apparent leak by the CRT leak check that runs before the vector is actually destroyed at program shutdown.

With a plain vector, shrink_to_fit() could be added to cleanupInfoContexts(), but this is not guaranteed as stated in the cppreference documentation:
"It is a non-binding request to reduce capacity() to size(). It depends on the implementation whether the request is fulfilled."

The other solution to use a plain vector would be to swap g_infoContexts with an empty vector.

If you prefer the plain vector swap, I will make this change. If not, I could add a comment to support the design choice of a unique_ptr.

Just let me know.

doctest/parts/private/static_registration.h
otrempe: Indeed! I will move it right before the class.

doctest/parts/private/static_registration.h
otrempe: Absolutely!

doctest/parts/private/static_registration.h
otrempe: Exactly. If an instance scopes a true leak, it won't be reported by the CRT leak tracking.
This, however, does not prevent other leak detector tools to work.

tests/CMakeLists.txt
otrempe: Given https://github.com/doctest/doctest/pull/1206#discussion_r3967170499 and this comment, I am no sure to understand the desired end result.

From what I think I understand, I would not modify the current condition "if(MSVC)" in the CMakeLists.txt.
In the test.205.cpp file, I would remove all preprocessor conditions.

This test would then fully run only on MSVC CIs because of the CMakeLists condition and would not be restricted to debug builds.

Is my understanding correct?

tests/regression/test.205.cpp
otrempe: See https://github.com/doctest/doctest/pull/1206#discussion_r3970914898

tests/regression/test.205.cpp
otrempe: Before this pull request, the instanciation of a reporter initialized a thread local static std::ostreamstring that survived until shutdown. It was detected as an apparent leak.

This block is there to verify the fix where the stream is now owned by the reporter instance which is destroyed before the _CrtDumpMemoryLeaks() is executed.

Should I change the comment "This used to own a thread-local stream for the process lifetime" to be more precise about the purpose of this block?

tests/regression/test.205.cpp
otrempe: See https://github.com/doctest/doctest/pull/1206#discussion_r3971092394

tests/regression/test.205.cpp
otrempe: The flag _CRTDBG_ALLOC_MEM_DF is already enabled by default in the MSVC debug CRT.

Other flags in the reprod block are just for automatic reporting at process shutdown and the textual location of the report.

This test depends only on the boolean result of _CrtDumpMemoryLeaks and not on the report itself, so it is not necessary to have all these flags.

